PR #348, opened 2014-05-20
Feature/jst

rrahn:
Adds new module for journaled string trees.


esiragusa:
@rrahn I close it as it must be re-created.

Issue #535, opened 2014-10-04
Add soft/hard clipping support to FragmentStore


holtgrewe:
The FragmentStore does neither support soft nor hard clipping.

For soft-clipping, the full sequence is in the SEQ field of SAM files and the anchor gaps data structure supports clipping with anchors. This should not be a problem.

For hard clipping, one would have to assume that the full sequence can be loaded into RAM (FragmentStore is not meant for genome alignment). When loading, one could fill the hard clipped region with N characters and when another region is found for the same sequence one could replace N by the actual sequence characters.

PR #580, opened 2014-10-06
[FIX] Fixed one issues with "S" in cigar string.

singerj:
The "S" in the cigar string was treated wrongly. However, this commit does not fully fix the problem as can be seen in the output of the test. The second line of the alignment is shifted by one. Since I don't know how this problem should be fixed (remove the first character in the second row, insert an gap in the first row, ...) I did not finish the test.

Note that this problem is also related to https://lists.fu-berlin.de/htdig/seqan-dev/2014-September/msg00015.html


esiragusa:
See #535.

PR #598, opened 2014-10-09
[fix, internal] iterators and infix in respect to infixes on modstrings

h-2:
as discussed previously; probably not complete yet.


esiragusa:
@h-2 I move this to 2.1.0.

esiragusa:
This hotfix is wrong according to what said in #644.

Issue #746, opened 2014-12-10
seqan_flexbar app tests fail


rrahn:
Hey does some one know what's wrong with that?
Seems the tests are failing since quite a while now.
Could track them back to 23rd of November (Then CDash runs out of logs). 
Is this a known issue?
@singerj @holtgrewe Who is maintaining it?

PR #754, opened 2014-12-14
[FIX] Fixed an issues with an incorret use of readRecord.

singerj:
Fixed issue #746. App test runs through.


esiragusa:
I close this PR as #755 is now solved.

PR #769, opened 2015-01-03
[INTERNAL] Using requirements.txt in Travis builds

holtgrewe:


esiragusa:
@holtgrewe Is this PR fixing Sphinx manual builds under travis? Could you please re-submit it to develop so that it gets in 2.0.0?

PR #773, opened 2015-01-12
Feature/add adapt stdarray

h-2:
add sequence interface for std::array (c++11)


rrahn:
Why do we need this already for the 2.0.0 release?
Would it not be sufficient to use the String<TValue, Array<> > for the moment?
I mean we would have to discuss the things with the String class anyway for the future, so I wouldn't add more dependencies that are not necessary for the release?
Especially as there is no official C++-11 support for the library, is there?

h-2:
it's a requirement for the blast module, but since that isn't going i we can postpone this, as well.

esiragusa:
@h-2 why is the c++11 std::array required by the blast module? By the way, it seems to me that std::array should replace seqan::ArrayString.

h-2:
> @h-2 why is the c++11 std::array required by the blast module? 

@esiragusa It was used for code-readability and convenience since c++11 is required anyway.

> By the way, it seems to me that std::array should replace seqan::ArrayString.

Just like the AllocString and std::string :stuck_out_tongue:

esiragusa:
Okay then we'll have a look at it when you make a PR for the blast module.

h-2:
@esiragusa or @rrahn please re-review.

h-2:
superseded by #997

Issue #699, opened 2014-11-16
Fatal Error in Banded Chain Alignment Traceback


rrahn:
The traceback computation of the banded chain alignment crashes with an assertion. 
See error log by Brett Bowman.
```
bbowman@localhost:~/git/SRSLI$ ./src/C++/srsli test/data/query/ecoli_5p_sample.fastq test/data/reference/ecoliK12_5p_assembly.fasta 
Query 0
Finished finding seeds
Finishing chaining seeds
Starting alignment of sequences
/usr/include/seqan/seeds/banded_chain_alignment_traceback.h:195 Assertion failed : isGlued == true was: 0 != 1 (Fatal error while trying to connect trace backs: No glue point available!)
Aborted (core dumped)
```

rrahn:
Turned out that the seeds were to far from each other and the score limit was reached causing a bitoverflow. Anyway there could be a check to test whether the actual size of the alignment matrix could cause troubles with a to low score value type.
But the the bandedChainAlignment does not work with the vector adaption.
There is an issue with the end() function if using a vector to hold the seeds.

rrahn:
#775 resolves issue with using vector as seed chain.
The remaining tasks are enhancements and can be done for 2.1.0

Issue #556, opened 2014-10-04
Simplify begin(), end().


holtgrewe:
Currently, the usage of begin() and end() and iterators is made complicated by rooted iterators and the selection between standard and rooted iterators.

This forces to use begin(X, Standard()) to make absolutely sure that a standard iterator (and thus pointer) is used in some situation and the cost for the rooted iterator (pointer back to the container) is not paid.

Similar to the C++11 standard's cbegin() and cend() functions, it would be nice to introduce rbegin() and rend() for returning rooted iterators. The second argument to begin() and end() could go away, programs could be much simpler, Iterator<T>::Type always returns the standard iterator.

The default iterator idea and metafunction can also go away that is often confused with standard iterators.

PR #775, opened 2015-01-13
[FIXED-#699] Fixes error with std::vector adaption in clang.

rrahn:
Fixes an issue when using std::vector for the seed chain passed to the banded-chain-aignment algorithm.

### DIFF
--- a/include/seqan/seeds/banded_chain_alignment_impl.h
+++ b/include/seqan/seeds/banded_chain_alignment_impl.h
@@ -86,11 +86,11 @@ struct BandedChainTracking
 
 template <typename TSeedSet>
 inline bool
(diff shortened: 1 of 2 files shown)


rrahn:
Travis build errors not related.

h-2:
I think the issue is more general than this, I am seeing this in other places, too. By adding `, Standard())` you force the use of `seqan::begin`, instead of the now globally available `std::begin`. I thought that our namespace should have precedence anyway, but apparently it doesn't.
But, appending `, Standard())`  really only solves the issue exactly at that place (and isn't really nice, since we wanted to get rid of that anyway, right?). I did the same thing in other places by using `seqan::begin` explicitly, but we really need a global solution to this.

rrahn:
You are absolutely right and there is already an issue for this  #556.

Issue #739, opened 2014-12-08
Adjust copyright year


esiragusa:
Edit and run util/bin/adjust_copy_right_year.sh

PR #787, opened 2015-01-19
Fix copyright year

esiragusa:
See #739.


esiragusa:
@rrahn what should we do with the INFO files within each module? These files seem totally useless to me...

Issue #780, opened 2015-01-14
enum in dox not sorted numerically...


h-2:
... but alphabetically instead. This is really confusing as the order of the enum values is important to the user.

esiragusa:
:+1:

rrahn:
Fixed by #788

esiragusa:
Hey is there any other place where we should disable lex sorting? Function calls, template arguments, etc?

rrahn:
As far as I could see, the order of the function parameters is in the same order as listed in the signature.
But I just saw this here: http://docs.seqan.de/seqan/develop/?p=String#String%23resizeSpace
Here the order of beginPos and endPos is changed in the parameter description compared to the parameter list in the signature.
I'll will have a look if I figure out why.

PR #788, opened 2015-01-20
[FIXES-#780] Disable sorting of enum values.

rrahn:
Disabled lexicographical sorting of enum values.


rrahn:
Dox errors not related.

PR #793, opened 2015-01-21
[fix] test_pipe to catch up with amino acid change

h-2:
fixes test_pipe


h-2:
since I don't see how this pr could have effected anything else I have merged it.

PR #797, opened 2015-01-22
[fix] warnings in seqcons2 on 32bit

h-2:
see http://cdash.seqan.de/viewBuildError.php?type=1&buildid=118021

[sorry for pushing this into upstream repo, will remove it after merge]


h-2:
will merge tomorrow noon

PR #802, opened 2015-01-26
[DOC,TEST] Building documentation is now a test.

holtgrewe:


holtgrewe:
Replaces #769 but might still need some work.

PR #805, opened 2015-01-27
Fix/win build

rrahn:
Changes:
   * Fixes timeout of demo for msa consensus on win platforms.
      * Note the python docu suggest to use communicate() instead of wait() if stdout and stderr are piped in the subprocess.


rrahn:
Travis errors not related.

PR #812, opened 2015-01-30
[fix] use std::abs instead of abs

h-2:
remove warning on `clang>=3.5` [-Wabsolute-value, http://llvm.org/releases/3.5.0/tools/clang/docs/ReleaseNotes.html]


h-2:
will merge in 30min :)

PR #813, opened 2015-01-30
Fix Fiona timeout on windows.

rrahn:
Changes
   * Changed python script to not get stuck in a deadlock due to piping the stdout and stderr.


weese:
looks good!

PR #814, opened 2015-01-30
Fix index leaks

esiragusa:
*Please don't merge it!*


weese:
So partialSum isn't the reason of memleaks?

weese:
Could the uninitialized problem in save be due to values in Jochens flattened-tree String that are neither filled nor accessed but now written to disk in save?

PR #815, opened 2015-01-30
Fixes issues with bgzf stream

rrahn:
Changes:
   * Changes off_t to off_type the offset type used by the bgzf stream. (They are different types on Windows)
   * Fixes seekoff if running queue is empty and current job does not encode the right block.

### DIFF
--- a/tests/bam_io/test_bam_file.h
+++ b/tests/bam_io/test_bam_file.h
@@ -480,14 +480,14 @@ SEQAN_DEFINE_TEST(test_bam_io_bam_file_bam_file_seek)
     seqan::BamHeader header;
     readHeader(header, bamFile);
 
-    seqan::String<seqan::Pair<off_t, int> > recs;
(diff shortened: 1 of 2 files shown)


tests/bam_io/test_bam_file.h
```diff
@@ -480,14 +480,14 @@ SEQAN_DEFINE_TEST(test_bam_io_bam_file_bam_file_seek)
     seqan::BamHeader header;
     readHeader(header, bamFile);
 
-    seqan::String<seqan::Pair<off_t, int> > recs;
+    seqan::String<seqan::Pair<uint64_t, int> > recs;
```
esiragusa: Shouldn't be off_type as well?

rrahn: Ok, I changed it to the Position<BamFile>::Type type, which is the same type returned by position().

weese:
Looks good

Issue #831, opened 2015-02-05
The escaping of html file names in the dox system seems to be odd


rrahn:
For example see ```FragmentStore::PredefinedAnnotationTypes```, where the frame-based link does not point to the correct address (when searching with ```?p=FragmentStore::PredefinedAnnotationTypes```).

The problem occurs, when the ```::``` in the name is converted in the ```write_html.py``` to some own encoding theme. Within ```common.js```, the URI is encoded using ```encodingURIComponent``` leading to a different sequence name. Moreover the script cuts of the part before the colons and directs it to the ```FragmentStore.html```, which is not the linked site to be displayed.

This needs some more thoughts and it would be sufficient to fix it after 2.0.0.
@holtgrewe @bkahlert Your insights to this would be nice!

holtgrewe:
This is one of the things we inherited from Andrea's dddoc ;) Yes, let's talk about this stuff after 2.0

h-2:
@xenigmax can you have a look? Thanks!

rrahn:
Apparently the current solution does not work as it generates html files with invalid characters for windows platforms.
What we really need to do, is to replace the ```::``` part of the html file nam with the encoding scheme used by the write_html.py. 
But be cautious the ```::``` is also used to for displaying member functions which might cause some troubles to generically apply the encoding to all ```::``` occurrences.

rrahn:
@xenigmax I wrote this little snippet here:
```
/**
 * Encode html file name. 
 * This function encodes html file names compliant with the write_html.py.
 */
(function ($) {
	$.extend({
		encodeName: function(name) {
            var tmp = "";
            for (var i = 0, len = name.length; i < len; i++) {           
                if (name[i].match(/[0-9a-z]/i))
                    tmp += name[i]
                else
                    tmp += '_' + name[i].charCodeAt(0);
            }
            // console.log('The encoded name: ' + tmp);
            return tmp;
		}
	});
})(jQuery);
```
We could use this!

xenigmax:
@rrahn  
Hey.. I'm sorry but I'm really confused about those things..
I found another thing that does not work properly.

There are some cases which use "#" multiple times. Below is an example.
http://docs.seqan.de/seqan/master/?p=Fragment#Fragment%23fragmentBegin

Should we seprate this from the current one?

xenigmax:
@rrahn 
In the .js code you suggested, we added "_" for skipping after discussion.

But there is a file which contains "_" in the middle of the name.
[...]

PR #832, opened 2015-02-05
[FIX] Fixes html name encoding.

rrahn:
Hot fix to get correct escaping of html names containing numbers, e.g. ```Blosum62```
This does not solve the general problem as described here: #831, but will fix most issues with entities containing numbers in their name.

### DIFF
--- a/util/py_lib/seqan/dox/write_html.py
+++ b/util/py_lib/seqan/dox/write_html.py
@@ -22,10 +22,11 @@ def escapeForXml(s):
 
 def escapeName(name):
     """Escape a name such that it is safe to use for files and anchors."""
+    """TODO(rmaerker): Encode special chars using urllib.quote(c.encode('utf8'))"""
     escape = '_'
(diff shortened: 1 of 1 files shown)


util/py_lib/seqan/dox/write_html.py
```diff
@@ -22,10 +22,11 @@ def escapeForXml(s):
 
 def escapeName(name):
```
esiragusa: Why don't you just replace escapeName(name) with urllib.quote(name) ?

rrahn: I can not just use urllib yet, as more complicated things will happen, due to automatic encoding/decoding of URI addresses. The presented hot fix solves the problems for the majority of fraud links due to numbers in the URI name.
The remaining issues (very rare cases), lead to valid addresses due to a bug in the common.js (#831).

rrahn:
No related errors detected.

PR #834, opened 2015-02-06
[DOC] Added LICENSE and README file for seqcons2.

rrahn:


rrahn:
@holtgrewe Can you please have a look at it.
Trying to do the packaging right now.

rrahn:
Ok anyway, we need it right now. I think this can be updated later on as well.

PR #843, opened 2015-02-10
[INTERNAL] Fixs sorting of versions in RSS Writer.

rrahn:


rrahn:
Nothing that could be detected during ci builds.

PR #848, opened 2015-02-13
[FEATURE,FIX,TEST] PackedString: Adds efficient testEqual for packed str...

rrahn:
...ings. Fixes bug for bit testAllOnes and testAllZeros for bit vector.

Adds test for testEqual for bit vector spec.

### DIFF
--- a/include/seqan/sequence/string_packed.h
+++ b/include/seqan/sequence/string_packed.h
@@ -1318,13 +1318,13 @@ reserve(
 }
 
 // ----------------------------------------------------------------------------
-// Function testEqual()
+// Function isEqual()
 // ----------------------------------------------------------------------------
 
 template <typename TValue, typename THostSpec>
 inline bool
-testEqual(String<TValue, Packed<THostSpec> > const & lhs,
-          String<TValue, Packed<THostSpec> > const & rhs)
+isEqual(String<TValue, Packed<THostSpec> > const & lhs,
+        String<TValue, Packed<THostSpec> > const & rhs)
 {
     typedef String<TValue, Packed<THostSpec> > TPackedString;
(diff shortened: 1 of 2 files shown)


include/seqan/sequence/string_packed.h
```diff
@@ -1966,28 +2052,6 @@ testAllOnes(String<bool, Packed<THostSpec> > const & obj)
                                      (length(obj) % TTraits::VALUES_PER_HOST_VALUE))));
 }
 
-// ----------------------------------------------------------------------------
```
esiragusa: Why did you remove open/save?

tests/sequence/test_string_packed_extension.h
```diff
@@ -598,6 +598,97 @@ void testStringPackedExtensionTestAllOnes()
     }
 }
 
+void testStringPackedExtensionTestEqual()
```
esiragusa: Don't we have generic sequence_v2 tests covering also isEqual?

include/seqan/sequence/string_packed.h
esiragusa: Oh it was moved above - never mind :+1:

include/seqan/sequence/string_packed.h
```diff
@@ -1316,6 +1317,87 @@ reserve(
     return capacity(seq);
 }
 
+// ----------------------------------------------------------------------------
```
esiragusa: Overloading ordEqual(), e.g. to have N != N, would have no effect here.

esiragusa: I think there should be at least one of isGreater() or isLess()...

esiragusa:
@rrahn I would put it in develop (2.1.0).

PR #851, opened 2015-02-14
Fix/finitealph value size type (develop)

meiers:
finite alphabets may have bigger sizes than uint8 can hold.
Note that this is the type of the ValueSize and not the type of Value (which is specified manully for SimpleType).


include/seqan/basic/alphabet_residue.h
```diff
@@ -592,7 +592,7 @@ struct Finite;
 template <typename TValue, unsigned SIZE>
 struct ValueSize<SimpleType<TValue, Finite<SIZE> > >
 {
-    typedef __uint8 Type;
+    typedef __uint64 Type;
```
esiragusa: I think this type should be unsigned as in the above template definition...

meiers:
done

PR #861, opened 2015-02-22
[FIX] Fixing bug in jumpToOrphans (master).

holtgrewe:


include/seqan/stream/formatted_file.h
```diff
@@ -400,6 +400,11 @@ struct FormattedFileContext
     typedef Nothing Type;
 };
 
+template <typename TFormattedFile, typename TStorageSpec>
+struct FormattedFileContext<TFormattedFile const, TStorageSpec> :
```
esiragusa: If the file is const, its context should also be const, right?

holtgrewe: Hm, `TStorageSpec` is something like `Dependent<>` or `Owner<>`. I am not sure about the implications of making tags consts, we used to have problems with const vs. non-const tags. What do you think?

include/seqan/stream/formatted_file.h
```diff
@@ -400,6 +400,11 @@ struct FormattedFileContext
...
+    FormattedFileContext<TFormattedFile, TStorageSpec>
```
esiragusa: Sorry, I should have commented the line below... I'm not referring to TStorageSpec. The FormattedFileContext of a TFormattedFile const should be a FormattedFileContext::Type const.

esiragusa: Hey @holtgrewe isn't this what you want?

```cpp
template <typename TFormattedFile, typename TStorageSpec>
struct FormattedFileContext<TFormattedFile const, TStorageSpec>
{
    typedef FormattedFileContext<TFormattedFile, TStorageSpec> const Type;
};
```

esiragusa:
@holtgrewe could you please answer my question about the FormattedFileContext? So I can quickly merge the PR. Thanks!

include/seqan/stream/formatted_file.h
holtgrewe: I guess that would work.

esiragusa: Can you please change that in your PR? I cannot...

holtgrewe:
@esiragusa OK, hope that's fine now.

PR #868, opened 2015-02-25
[FEATURE] tools to compute variant density and basepair-accurate read depth

weese:
I added tools to compute read depth and number of variants window-wise given a sorted SAM/BAM or a VCF file. The output is a 4-column bed file with lines like ```contigName, begin, end, value``` where value is the read depth or number of variants.

If the window-size is 0, fx_read_depth outputs each maximal stretch of equal read depth as one interval.

(Sorry for the extra stuff from the tabix PR. Should be gone after merging the corresponding PR)

*REMARK:* 
fx_read_depth differs from the existing fx_bam_coverage tool in that it computes basepair-accurate read depths (similar to bedtools genomecov) while fx_bam_coverage computes only the number of reads starting in each window (independent of their length).


esiragusa:
Hey Dave, by chance did you fix fx_bam_coverage in this PR? Here the call to readSequence() doesn't work...

weese:
Done. Fixes:
- the tool still checked the old-style error codes instead of a boolean [corrected]
- building a fasta index was always called [changed to save and later open the index] (however, there is no need for a fasta index overall)
- building a fasta index took ages since I replaced the Mmap String by an ifstream [fixed], reason: tell on a fstream calls seek which always empties and later reloads the stream buffer |-(

Issue #866, opened 2015-02-24
FormattedFile::readRecord is slow with std::string


mr-c:
`stream/tokenization.h:311 Assertion failed : optr < ochunk.end should be true but was 0`

Here's a minimal test case:

```
(env)mcrusoe@athyra:~/khmer/gl-master$ g++ -I/home/mcrusoe/src/seqan/seqan-1.4.2/core/include/ seqan-std-string-test-1.4.cc -o seqan-std-string-test-1.4 -g
(env)mcrusoe@athyra:~/khmer/gl-master$ g++ -I/home/mcrusoe/src/seqan/seqan-seqan-v2.0.0/include/ seqan-std-string-test-2.0.cc -o seqan-std-string-test-2.0 -g
(env)mcrusoe@athyra:~/khmer/gl-master$ ./seqan-std-string-test-1.4 tests/test-data/random-20-a.fa | head
35      CGCAGGCTGGATTCTAGAGGCAGAGGTGAGCTATAAGATATTGCATACGTTGAGCCAGC
16      CGGAAGCCCAATGAGTTGTCAGAGTCACCTCCACCCCGGGCCCTGTTAGCTACGTCCGT
46      GGTCGTGTTGGGTTAACAAAGGATCCCTGACTCGATCCAGCTGGGTAGGGTAACTATGT
40      GGCTGAAGGAGCGGGCGTACGTGTTTACGGCATGATGGCCGGTGATTATGGGGGACGGG
33      GCAGCGGCTTTGAATGCCGAATATATAACAGCGACGGGGTTCAATAAGCTGCACATGCG
98      ACCAGATGCATAGCCCAACAGCTGAGACATTCCCAGCTCGCGAACCAAGACGTGAGAGC
17      CCCTGTTAGCTACGTCCGTCTAAGGATATTAACATAGTTGCGACTGCGTCCTGTGCTCA
89      GCGAGATACTAGCAAAGGTTCATCAACAGCTACACCCGACGAACCCCGAGAAATTGGGA
30      GTTATGGTCCAGGATGAATGCGCGTACCGGGCGCCTATCACTCCTCTTGTCATTCAGAA
82      ATGCACTATATTTAAGAGGTCTAGAGTGTAAAAAGTGTACCCTTCGGGGTGGAGCTGTT
(env)mcrusoe@athyra:~/khmer/gl-master$ ./seqan-std-string-test-2.0 tests/test-data/random-20-a.fa
[...]

PR #870, opened 2015-02-25
[FIXED-#866] Tokenization: Disabled chunking for std::string.

rrahn:
Changes are:
   * Overloads the ```Chunk``` metafunction for ```std::basic_string``` to disable chunking when reading a file into this type.
   * Adds forward declaration for basic append value.
   * Adapts test.

### DIFF
--- a/include/seqan/sequence/adapt_std_string.h
+++ b/include/seqan/sequence/adapt_std_string.h
@@ -163,6 +163,21 @@ struct DefaultOverflowImplicit< std::basic_string<TChar, TCharTraits, TAlloc> >
 template <typename TChar, typename TCharTraits, typename TAlloc>
 struct IsSequence<std::basic_string<TChar, TCharTraits, TAlloc> > : True {};
 
+// ----------------------------------------------------------------------------
+// Metafunction Chunk
+// ----------------------------------------------------------------------------
+
+// Chunk interface for std::basic strings.
+template <typename TChar, typename TCharTraits, typename TAlloc, typename TSpec>
+struct Chunk<Iter<std::basic_string<TChar, TCharTraits, TAlloc>, AdaptorIterator<TChar*, TSpec> > >
+{
+    typedef typename Chunk<std::basic_string<TChar, TCharTraits, TAlloc> >::Type Type;
+};
+
+template <typename TChar, typename TCharTraits, typename TAlloc, typename TSpec>
+struct Chunk<Iter<std::basic_string<TChar, TCharTraits, TAlloc> const, AdaptorIterator<TChar*, TSpec> > > :
+    Chunk<Iter<std::basic_string<TChar, TCharTraits, TAlloc>, AdaptorIterator<TChar*, TSpec> > > {};
+
 // ===========================================================================
 // Functions
 // ===========================================================================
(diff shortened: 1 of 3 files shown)


include/seqan/sequence/adapt_std_string.h
```diff
@@ -163,6 +163,21 @@ struct DefaultOverflowImplicit< std::basic_string<TChar, TCharTraits, TAlloc> >
...
+// ----------------------------------------------------------------------------
+// Metafunction Chunk
+// ----------------------------------------------------------------------------
+
+// Chunk interface for std::basic strings.
+template <typename TChar, typename TCharTraits, typename TAlloc, typename TSpec>
+struct Chunk<Iter<std::basic_string<TChar, TCharTraits, TAlloc>, AdaptorIterator<TChar*, TSpec> > >
+{
+    typedef typename Chunk<std::basic_string<TChar, TCharTraits, TAlloc> >::Type Type;
+};
```
weese: This will work, but is not very efficient. ```Chunk<std::basic_string<TChar, TCharTraits, TAlloc> >::Type``` will evaluate to ```Nothing``` and disables chunking for the std::string. Instead the slower appendValue will be used. It is fine for now, but we should leave the corresponding bug report open and put a remark there to actually provide real chunking for std::string.
I think many of the functionality is already covered by the chunking support for the ```Iter<TContainer, AdaptorIterator<TValue*, TSpec> > ``` in basic/iterator_adaptor.h but it seems that some functions behave different or are not used for std::string.

esiragusa: But do you have any idea of what's the problem of chunking with std::basic_string<char>? Does it work with std::vector<char> instead?

rrahn: The problem, as far as I could see it, is, that in the current way we exploit the fact that we can resize the SeqAn String by moving the end pointer without really allocating memory. But in conjunction with the ```reserveChunk``` this is fine since then the new memory block is allocated. However, while the chunk is filled while reading in the end we simply advance the string again to cover the filled values. This is something we cannot do with the std::string as we have no control over the private members. Instead one need a complete different strategy, that covers the same behaviour by resizing the std::sting. My guess is, that the std::vector causes the same problem for the named issues, but I didn't try it yet. That's why I decided to fix it first by disabling the chunking. 

There is also another problem. For example when using ```readRecords``` with std::string the compiler fails since the FormattedFileContext uses seqan::CharString causing an invalid construction of the std::string with a seqan::CharString. Probably, the context value types must be adaptable?

weese: Mmh. I think we can do chunking for std::vector and std::string in the same way without using private members. The only problem I see is that before C++11 you couldn't rely to get contiguous memory chunks from the string, i.e. you cannot simply convert the iterators into a char* (In practice all standard STL implementations use contiguous memory, but at least the TBB not). The second problem might be that it would be not compliant to write in reserved/not-resized storage. 

Hence reserveChunk must use resize and I have to think about how to implement advanceChunk then.

rrahn: Oh, I didn't know the first problem. But anyway, the interfaces advance and reverseChunk need to be overloaded. I agree. 
So should we wait for the patch or can we merge the current state into master/develop as a hot-fix. At least one can use std::strings without a compiler error. :yum:

weese:
I'll merge it. Can you leave the corresponding bug report open, refer to this conversion and put a remark that we have to actually provide real chunking for std::string.

esiragusa:
@rrahn If you're still on this issue, would you please have a quick look also at std::vector? I'm afraid of similar problems...

rrahn:
@weese I'll do
@esiragusa I'll do

PR #875, opened 2015-02-26
[feature] add new exception for recoverable parse errors (develop)

h-2:
and fix the dox for the existing ones.

A recoverable parse error can be thrown if something unexpected happened, but the program can continue to function.

PR #876, opened 2015-02-26
[feature] add new exception for recoverable parse errors (master)

h-2:
same as #875 but for master branch


esiragusa:
Why is it an hotfix?

h-2:
ah, should be 2.1.0 instead

PR #881, opened 2015-02-27
[fix] remove const from type (develop)

h-2:
this makes readUntil work on const sequences as well.


h-2:
failed tests unrelated (I think). 

PS why is razers being built again?

PR #890, opened 2015-03-03
[FIX,INTERNAL] make concat with delimiter, strSplit and some internals work with std::strings (master)

h-2:
this makes it possible to
  * assign an infix of an std::string to an std::string
  * make the `concat()` version with a delimiter work with std::strings

PS the commit incorrectly speaks of `strSplit()`, but it means the opposite (i.e. `concat()`)

### DIFF
--- a/include/seqan/basic/array_construct_destruct.h
+++ b/include/seqan/basic/array_construct_destruct.h
@@ -251,7 +251,8 @@ struct ValueConstructor_
     {
         typedef typename Value<TIterator>::Type    TValue;
         typedef typename RemoveConst<TValue>::Type TNonConstValue;
-        new( (void*) & value(it) ) TNonConstValue(SEQAN_FORWARD(TParam, param_));
+        new( (void*) & value(it) ) TNonConstValue();
+        assign(*it, SEQAN_FORWARD(TParam, param_));
     }
 
 #ifndef SEQAN_CXX11_STANDARD
--- a/include/seqan/basic/array_construct_destruct.h
+++ b/include/seqan/basic/array_construct_destruct.h
@@ -251,8 +251,7 @@ struct ValueConstructor_
     {
         typedef typename Value<TIterator>::Type    TValue;
         typedef typename RemoveConst<TValue>::Type TNonConstValue;
-        new( (void*) & value(it) ) TNonConstValue();
-        assign(*it, SEQAN_FORWARD(TParam, param_));
+        new( (void*) & value(it) ) TNonConstValue(SEQAN_FORWARD(TParam, param_));
     }
 
 #ifndef SEQAN_CXX11_STANDARD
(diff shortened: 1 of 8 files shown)


include/seqan/basic/array_construct_destruct.h
```diff
@@ -251,7 +251,8 @@ struct ValueConstructor_
     {
         typedef typename Value<TIterator>::Type    TValue;
         typedef typename RemoveConst<TValue>::Type TNonConstValue;
-        new( (void*) & value(it) ) TNonConstValue(SEQAN_FORWARD(TParam, param_));
+        new( (void*) & value(it) ) TNonConstValue();
```
esiragusa: Why?

include/seqan/sequence/string_base.h
```diff
@@ -1316,7 +1316,8 @@ struct AppendValueToString_
         TSize me_length = length(me);
         if (capacity(me) <= me_length)
         {
-            TTargetValue temp_copy(SEQAN_FORWARD(TValue, _value)); //temp copy because resize could invalidate _value
+            TTargetValue temp_copy;
+            assign(temp_copy, SEQAN_FORWARD(TValue, _value)); //temp copy because resize could invalidate _value
```
esiragusa: Why?

include/seqan/sequence/string_set_base.h
```diff
@@ -1598,10 +1598,10 @@ concat(StringSet<TString, TSpec> const & constMe)
 }
 
 template <typename TStrings, typename TDelim>
-inline String<typename Value<typename Value<TStrings>::Type>::Type>
+inline typename Value<TStrings>::Type
```
esiragusa: When is this specialization called? For a String of Strings? And what does Concatenator<TSTrings> return?

include/seqan/basic/array_construct_destruct.h
h-2: std::string has no constructor for Segment<std::strng ...> but it does have an overloaded assign.

include/seqan/sequence/string_base.h
h-2: std::string has no constructor for Segment<std::strng ...> but it does have an overloaded assign.

include/seqan/sequence/string_set_base.h
h-2: this specialization is called when you concatenate with a delimeter. Maybe `Concatenator<TSTrings>`  would be better, I don't know. I also don't know why it was done in the way it was before.

h-2:
This apparently breaks some stuff that is not immediately related to strings, because some other types have constructor for args, for which they don't have assignment overloads. I will refactor this later.

h-2:
build succeeds on gcc; error on clang is independent I think.

h-2:
0 errors, 0 warnings, all pass on gcc [clang had a timeout error on git-clone].

@esiragusa can you review and merge this? also #910 ? Thx

PR #891, opened 2015-03-03
[FIX,INTERNAL] make concat with delimiter, strSplit and some internals work with std::strings (develop)

h-2:
this makes it possible to
  * assign an infix of an std::string to an std::string
  * make the `concat()` version with a delimiter work with std::strings

PS the commit incorrectly speaks of `strSplit()`, but it means the opposite (i.e. `concat()`)

### DIFF
--- a/include/seqan/basic/array_construct_destruct.h
+++ b/include/seqan/basic/array_construct_destruct.h
@@ -251,7 +251,8 @@ struct ValueConstructor_
     {
         typedef typename Value<TIterator>::Type    TValue;
         typedef typename RemoveConst<TValue>::Type TNonConstValue;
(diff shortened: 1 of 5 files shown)


h-2:
wow, this actually breaks some stuff, I will have a look at it.

h-2:
same as with master; only microrazers fails which fails on other builds as well.

esiragusa:
@h-2 please don't merge it as it doesn't look good. better discuss it with me or @rrahn at the uni. you can keep it in your local branch for now.

esiragusa:
Hannes, can you please recreate this PR for develop using #890's h-2:fix/strSplit_for_stdstring branch?

h-2:
@esiragusa will do.

PR #895, opened 2015-03-04
[DOC] replaced workflow_plugin_dir.zip which was xml file

temehi:


esiragusa:
@temehi does this file contain a template to create a new generic knime node?

temehi:
Yes, It has all the files mentioned in the tutorial

Issue #904, opened 2015-03-11
 bool open(file, fileName) (fileFormat): doesn't compile if fileName is a `char*` type


CVroland:
In seqan 2.0.0, the function open(file, fileName) from FormattedFile doesn't compile if fileName  is a `char*` type. It need a `const char*` type. 
however, in seqan 1.4.1 open function for SequenceStream worked fine with a `char*` type.

rrahn:
Fixed by #913 #914

PR #913, opened 2015-03-19
[FIXES-#904] Fixes issue #904

rrahn:
Enables generic open function only if ```StreamConcept``` is fulfilled.
Otherwise ```char *``` are not converted to ```const char *``` as the generic ```TStream &``` is chosen by the compiler.


weese:
good job, lgtm.

rrahn:
Build errors not related.

PR #914, opened 2015-03-19
[FIXES-#904] Fixes issue #904

rrahn:
Enables generic open function only if ```StreamConcept``` is fulfilled.
Otherwise ```char *``` are not converted to ```const char *``` as the generic ```TStream &``` is chosen by the compiler.

Hotfix applied to develop as well.


rrahn:
See issue #904

rrahn:
build errors not related.

PR #915, opened 2015-03-20
[FEATURE] Reboot of interval tree implementation.

holtgrewe:
Static interval trees are better implemented as arrays than
pointer-based trees. This is an attempt at an implementation.

Includes test and demos, docs are missing.

Major difference to the current implementation in misc is that
the intervals are 1-based instead of 0-based.

@esiragusa @h-2 @rrahn @weese I realize that the documentation is missing but I want to push this in early for a review. The difference in coordinate systems used (0-based here 1-based in AKE's implementation) is a feature. I have not done any benchmarking but I assume that this implementation is not only simpler (and thus easier to understand and maintain) but also faster.

The same trick could be used to implement a dynamic `IntervalTree<>` based on RB-trees, for example. However, I do not see its direct use and RB-trees are somewhat tricky to implement efficiently.

Note that the `Interval<>` class is only here for convenience and any other type implementing `beginPos()` and `endPos()` can be used. Further, not the cargos are returned but pointers to the `TValue`s of the `IntervalTree<>`.

Please point out missing functions for this to be a `ContainerConcept`. Note that for some reason, I did not succeed in writing generic `Iterator<>` functions so I would welcome any help here. After review, I will update the PR to also have documentation.

### DIFF
--- a/demos/interval_tree_genome_region.cpp
+++ b/demos/interval_tree_genome_region.cpp
@@ -1,0 +1,66 @@
+// Demo for the updated interval trees with the GenomicRegion class
+// that implements the IntervalConcept.
+//
+// In this demo, we use Interval class with unsigned cargo and int position
+// type.
+
+#include <vector>
+
+#include <seqan/basic.h>
+#include <seqan/interval_tree.h>
+#include <seqan/sequence.h>
+#include <seqan/seq_io/genomic_region.h>
+#include <seqan/stream.h>
+
+using namespace seqan;
+
+int main()
+{
+    // Using vector here, but any container will do.
+    std::vector<GenomicRegion> intervals;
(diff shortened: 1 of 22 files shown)


h-2:
@holtgrewe We are planning 2.1.0 for late January and would like to include this. Can you (a) provide some benchmarks over the existing implementation, and (b) make sure that tests pass?

Also, can you review where and how the existing implementation is used and whether it can be completely replaced by the new one? We'd rather not have competing implementations in the library.
Thanks!

holtgrewe:
Sorry, I do not have sufficient time to invest in this project.

h-2:
@holtgrewe Ok, then I will close it for now. Thanks in any case.

Issue #919, opened 2015-03-27
cant join StringSet<DnaString> to another StringSet<DnaString> using AppendValue()


fari-99:
is it appendValue thread safe? 

i'm using OpenMP parallel for sequencing. but i kinda stuck in appendValue for stored my sequence. it's always say "segmentation fault". So i tried to parting my variable for storing into 4 part (ex: spectrumLocal[thread] with thread = 4). but i can't join them using appendValue() (ex: appendValue(spectrumGlobal, spectrumLocal[id]). 

can someone give solution or alternative code? 

thanks.
best regards.

fadhlan.

note: 
i already using appendValue(spectrumGlobal, value), but its critical, so the performance slow down, if not critical then the program will crash because segmentation fault. 

i already use ConcurrentQueue for using it, it's very good, but i cant store the value, btw the value is string not a number (ex: "ACGT").

h-2:
To concatenate two sequences, please use `append()` and not `appendValue()`.

`append()` is used to append a sequence to a sequence
`append()` is used to append a value to a sequence

PS: no types are thread safe ;)

fari-99:
Hey, thanks for replying so fast.

yeah, i tried to use append() for it (like this: append(spectrumGlobal,
spectrumLocal[id_thread])). but i get an error like this.

/usr/include/seqan/sequence/sequence_interface.h|957|error:
no matching function for call to
‘append(seqan::StringSet<seqan::String<seqan::SimpleType<unsigned char,
seqan::Dna_>, seqan::Alloc<> > >&,
seqan::StringSet<seqan::String<seqan::SimpleType<unsigned char,
seqan::Dna_>, seqan::Alloc<> > >&,
seqan::DefaultOverflowImplicit<seqan::StringSet<seqan::String<seqan::SimpleType<unsigned
char, seqan::Dna_>, seqan::Alloc<> > > >::Type)’|

i dont understand with that error. can you help me?

thanks.

best regards.

2015-03-27 15:04 GMT+07:00 Hannes Hauswedell <notifications@github.com>:

h-2:
Hm, that's strange. Can you give me a more complete error log?

fari-99:
i'm sorry maybe i'm making you confusing with my explanation earlier.

so my program looks like this (not exactly like this but i made it simpler
though).

int main()
{
    StringSet<DnaString> pool;
    StringSet<DnaString> sekuens;

    //no parallel
    for(int i = 0; i < 12; i++)
    {
[...]

PR #922, opened 2015-03-31
[FIXED-#919] fix and activate StringSet tests as well as other StringSet related fixes and additions

h-2:
Apparently it was not possible to append a non-concat-direct owner stringset to another one. This is a quick fix for #919, although something with tests and a more general solution would ne nicer (the trivial more general solutions do not work, I tried).

### DIFF
--- a/include/seqan/sequence/string_set_base.h
+++ b/include/seqan/sequence/string_set_base.h
@@ -1783,12 +1783,11 @@ inline SEQAN_FUNC_ENABLE_IF(And<Is<ContainerConcept<TSequences> >,
                                 Is<ContainerConcept<typename Value<TSequences>::Type > > >, void)
 write(TTarget & target, TSequences const & seqs, TDelim const & delimiter)
 {
-    typedef typename Size<TSequences const>::Type TSize;
     typedef typename Iterator<TSequences const>::Type TSourceIt;
 
     if (SEQAN_UNLIKELY(empty(seqs)))
         return;
-    for (TSourceIt it = begin(seqs), itBack = (end(seqs) - 1); it != itBack; ++it)
+    for (TSourceIt it = begin(seqs, Standard()), itBack = (end(seqs, Standard()) - 1); it != itBack; ++it)
     {
         write(target, *it);
         write(target, delimiter);
@@ -1817,9 +1816,9 @@ append(TSequences1 & me, TSequences2 const & obj, Tag<TExpand>)
 {
     typedef typename Iterator<TSequences2 const>::Type TSourceIt;
 
-    unsigned oldLength = length(me);
+    typename Size<typename Value<TSequences1>::Type>::Type oldLength = length(me);
     resize(me, oldLength + length(obj), Tag<TExpand>());
-    for (TSourceIt it = begin(obj), itEnd = end(obj); it != itEnd; ++it)
+    for (TSourceIt it = begin(obj, Standard()), itEnd = end(obj, Standard()); it != itEnd; ++it)
         assignValue(me, oldLength++, *it);
 }
 
--- a/include/seqan/sequence/string_set_base.h
+++ b/include/seqan/sequence/string_set_base.h
@@ -1775,7 +1775,7 @@ operator!=(StringSet<TLeftString, TLeftSpec> const & left,
 }
 
 // ----------------------------------------------------------------------------
-// Function write(StringSet)
+// Function write() (works on any container of container)
 // ----------------------------------------------------------------------------
 
 template <typename TTarget, typename TSequences, typename TDelim>
@@ -1784,11 +1784,13 @@ inline SEQAN_FUNC_ENABLE_IF(And<Is<ContainerConcept<TSequences> >,
 write(TTarget & target, TSequences const & seqs, TDelim const & delimiter)
 {
     typedef typename Size<TSequences const>::Type TSize;
+    typedef typename Iterator<TSequences const>::Type TSourceIt;
+
     if (SEQAN_UNLIKELY(empty(seqs)))
         return;
-    for (TSize i = 0; i < length(seqs) - 1; ++i)
+    for (TSourceIt it = begin(seqs), itBack = --end(seqs); it != itBack; ++it)
     {
-        write(target, seqs[i]);
+        write(target, *it);
         write(target, delimiter);
     }
     write(target, back(seqs)); // no delimiter after last
@@ -1800,6 +1802,25 @@ inline SEQAN_FUNC_ENABLE_IF(And<Is<ContainerConcept<TSequences> >,
 write(TTarget & target, TSequences const & seqs)
 {
     write(target, seqs, '\n');
+}
+
+// ----------------------------------------------------------------------------
+// Function append() (works on any container of container)
+// ----------------------------------------------------------------------------
+
+template <typename TSequences1, typename TSequences2, typename TExpand >
+inline SEQAN_FUNC_ENABLE_IF(And<And<Is<ContainerConcept<TSequences1> >,
+                                    Is<ContainerConcept<typename Value<TSequences1>::Type > > >,
+                                And<Is<ContainerConcept<TSequences2> >,
+                                    Is<ContainerConcept<typename Value<TSequences2>::Type > > > >, void)
+append(TSequences1 & me, TSequences2 const & obj, Tag<TExpand>)
+{
+    typedef typename Iterator<TSequences2 const>::Type TSourceIt;
+
+    unsigned oldLength = length(me);
+    resize(me, oldLength + length(obj), Tag<TExpand>());
+    for (TSourceIt it = begin(obj), itEnd = end(obj); it != itEnd; ++it)
+        assignValue(me, oldLength++, *it);
 }
 
 }  // namespace seqan
--- a/include/seqan/sequence/string_set_base.h
+++ b/include/seqan/sequence/string_set_base.h
@@ -1788,7 +1788,7 @@ write(TTarget & target, TSequences const & seqs, TDelim const & delimiter)
 
     if (SEQAN_UNLIKELY(empty(seqs)))
         return;
-    for (TSourceIt it = begin(seqs), itBack = --end(seqs); it != itBack; ++it)
+    for (TSourceIt it = begin(seqs), itBack = (end(seqs) - 1); it != itBack; ++it)
     {
         write(target, *it);
         write(target, delimiter);
--- a/include/seqan/sequence/string_set_concat_direct.h
+++ b/include/seqan/sequence/string_set_concat_direct.h
@@ -141,22 +141,6 @@ public:
     }
 };
 
-// --------------------------------------------------------------------------
-// Functional Adder_ (used in second signature of append())
-// --------------------------------------------------------------------------
-
-struct Adder_ : public std::unary_function<unsigned, unsigned>
-{
-    unsigned plus;
-    Adder_(unsigned _plus) : plus(_plus) {}
-
-    template <typename TValue>
-    unsigned operator() (TValue const & val) const
-    {
-        return val + plus;
-    }
-};
-
 // ============================================================================
 // Metafunctions
 // ============================================================================
@@ -365,7 +349,7 @@ append(StringSet<TString, Owner<ConcatDirect<TSpec> > > & me,
 {
     typedef typename Iterator<TStrings2 const>::Type TIt;
 
-    reserve(me.concat, lengthSum(me) + length(obj), Tag<TExpand>());
+    reserve(me.concat, lengthSum(me) + lengthSum(obj), Tag<TExpand>());
     reserve(me.limits, length(me.limits) + length(obj), Tag<TExpand>());
 
     for (TIt it = begin(obj), itEnd = end(obj); it != itEnd; ++it)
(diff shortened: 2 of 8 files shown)


weese:
@h-2: 
Strangely enough you seem to be the first to implement append for StringSets. Can you please provide a more generic version that works for any TSource and not only for the Owner-Stringet as target. Compare with lines 1260... in string_base.h

h-2:
@weese Yes, I was kind of surprised, as well. 
I can add a more general solution for the Owner-StringSet, but I don't have time to write tests now or fix the other StringSet types, as this was only meant as a quick fix for an externally reported bug.

h-2:
please hold for now

h-2:
Ok, originally I only wanted to fix `append()`, then I realised that **none of the tests** for StringSets are currently executed. Then I spent two days trying to fix most of them.
The bad news: it's still mess.
The good news: close to 500 tests now build and pass on StringSets and the mess isnt worse than before.

Oh, and `append()` now works for some StringSets.

Please review and merge.

include/seqan/sequence/string_set_base.h
```diff
@@ -1784,11 +1784,13 @@ inline SEQAN_FUNC_ENABLE_IF(And<Is<ContainerConcept<TSequences> >,
 write(TTarget & target, TSequences const & seqs, TDelim const & delimiter)
 {
     typedef typename Size<TSequences const>::Type TSize;
+    typedef typename Iterator<TSequences const>::Type TSourceIt;
+
     if (SEQAN_UNLIKELY(empty(seqs)))
         return;
-    for (TSize i = 0; i < length(seqs) - 1; ++i)
+    for (TSourceIt it = begin(seqs), itBack = (end(seqs) - 1); it != itBack; ++it)
     {
```
rrahn: Nice catch! But I experienced a lot of troubles if you are not exact with the Iterator type. Maybe add the Standard() tag to begin and end so it won't mess up.

include/seqan/sequence/string_set_base.h
```diff
@@ -1802,6 +1804,25 @@ write(TTarget & target, TSequences const & seqs)
...
+template <typename TSequences1, typename TSequences2, typename TExpand >
+inline SEQAN_FUNC_ENABLE_IF(And<And<Is<ContainerConcept<TSequences1> >,
+                                    Is<ContainerConcept<typename Value<TSequences1>::Type > > >,
+                                And<Is<ContainerConcept<TSequences2> >,
+                                    Is<ContainerConcept<typename Value<TSequences2>::Type > > > >, void)
+append(TSequences1 & me, TSequences2 const & obj, Tag<TExpand>)
+{
+    typedef typename Iterator<TSequences2 const>::Type TSourceIt;
+
+    unsigned oldLength = length(me);
```
rrahn: Please use Size type for TSequences1.

include/seqan/sequence/string_set_base.h
```diff
@@ -1802,6 +1804,25 @@ write(TTarget & target, TSequences const & seqs)
...
+    resize(me, oldLength + length(obj), Tag<TExpand>());
+    for (TSourceIt it = begin(obj), itEnd = end(obj); it != itEnd; ++it)
+        assignValue(me, oldLength++, *it);
```
rrahn: Use Standard() here as well.

include/seqan/sequence/string_set_concat_direct.h
```diff
@@ -139,6 +141,22 @@ class StringSet<TString, Owner<ConcatDirect<TSpec> > >
...
+    unsigned plus;
+    Adder_(unsigned _plus) : plus(_plus) {}
+
+    template <typename TValue>
+    unsigned operator() (TValue const & val) const
+    {
+        return val + plus;
+    }
+};
+
```
rrahn: Do we still need this? I see that the corresponding code is commented. So this will not be used. If it will be used, please use a generic Size type instead of the unsigned. But if this is not needed anymore, please remove it.

include/seqan/sequence/string_set_concat_direct.h
```diff
@@ -334,6 +352,46 @@ template <typename TString, typename TSpec >
...
+//         return;
+// 
+//     unsigned const oldLimLength = length(me.limits);
+//     Adder_ op(back(me.limits));
+// 
+//     append(me.concat, obj.concat, Tag<TExpand>());
+//     append(me.limits, suffix(obj.limits, 1), Tag<TExpand>());
+// 
+//     std::transform(begin(me.limits) + oldLimLength, end(me.limits), op);
+// }
```
rrahn: Can this overload be removed?

include/seqan/sequence/string_set_concat_direct.h
```diff
@@ -334,6 +352,46 @@ template <typename TString, typename TSpec >
...
+append(StringSet<TString, Owner<ConcatDirect<TSpec> > > & me,
+       TStrings2 const & obj,
+       Tag<TExpand>)
+{
+    typedef typename Iterator<TStrings2 const>::Type TIt;
+
+    reserve(me.concat, lengthSum(me) + length(obj), Tag<TExpand>());
+    reserve(me.limits, length(me.limits) + length(obj), Tag<TExpand>());
+
+    for (TIt it = begin(obj), itEnd = end(obj); it != itEnd; ++it)
```
rrahn: Use Standard() here again.

include/seqan/sequence/string_set_concat_direct.h
```diff
@@ -334,6 +352,46 @@ template <typename TString, typename TSpec >
...
+template <typename TString, typename TSpec, typename TStrings2, typename TExpand >
+inline SEQAN_FUNC_ENABLE_IF(And<Is<ContainerConcept<TStrings2> >,
+                                Is<ContainerConcept<typename Value<TStrings2>::Type > > >, void)
+append(StringSet<TString, Owner<ConcatDirect<TSpec> > > & me,
+       TStrings2 const & obj,
+       Tag<TExpand>)
+{
+    typedef typename Iterator<TStrings2 const>::Type TIt;
+
+    reserve(me.concat, lengthSum(me) + length(obj), Tag<TExpand>());
```
rrahn: Does length(obj) does not return the number of strings contained in obj? Should it not be lengthSum(obj) to determine the total size of the concatenated string set?

include/seqan/sequence/string_set_concat_direct.h
```diff
@@ -429,7 +487,7 @@ template <typename TString, typename TSpec, typename TSize, typename TExpand >
 inline typename Size<StringSet<TString, Owner<ConcatDirect<TSpec> > > >::Type
 resize(StringSet<TString, Owner<ConcatDirect<TSpec> > > & me, TSize new_size, Tag<TExpand> tag)
 {
-    if (new_size < length(me.limits))
+    if (static_cast<typename Size<StringSet<TString, Owner<ConcatDirect<TSpec> > > >::Type>(new_size) < length(me.limits))
```
rrahn: I know this is kind of pedantic, but it should be the Size type of StringSetLimits of StringSet ...
I know it is basically the same, but who knows what will be changed in the future.
:trollface:

include/seqan/sequence/string_set_base.h
```diff
@@ -1802,6 +1804,25 @@ write(TTarget & target, TSequences const & seqs)
...
+                                    Is<ContainerConcept<typename Value<TSequences1>::Type > > >,
+                                And<Is<ContainerConcept<TSequences2> >,
+                                    Is<ContainerConcept<typename Value<TSequences2>::Type > > > >, void)
+append(TSequences1 & me, TSequences2 const & obj, Tag<TExpand>)
+{
+    typedef typename Iterator<TSequences2 const>::Type TSourceIt;
+
+    unsigned oldLength = length(me);
+    resize(me, oldLength + length(obj), Tag<TExpand>());
+    for (TSourceIt it = begin(obj), itEnd = end(obj); it != itEnd; ++it)
```
rrahn: Why not using arrayCopyForward?

include/seqan/sequence/string_set_owner.h
```diff
@@ -141,6 +141,21 @@ class StringSet<TString, Owner<Default> >
...
+inline void append(StringSet<TString, Owner<Default> > & me,
+                   StringSet<TString2, Owner<Default> > const & obj,
+                   Tag<TExpand>)
+{
+    // we rather invalidate limits here to allow to do modify appended strings:
+    me.limitsValid = false;
+    append(me.strings, obj.strings, Tag<TExpand>());
+}
+
+// --------------------------------------------------------------------------
```
rrahn: Why do we need this overload?
This is essentially the same as the generic append version for string sets.
The difference is that limits is not valid, but we also have dependent string sets, which do need to set their limits accordingly. I think it would be better to add a helper function setLimitsState which does nothing for STL containers and otherwise sets the limits to ```false```.

tests/sequence/test_stringset_v2.cpp
```diff
@@ -0,0 +1,61 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ==========================================================================
+// Author: Jochen Singer <jochen.singer@fu-berlin.de>
+// ==========================================================================
```
rrahn: Maybe your name ;)

tests/sequence/test_stringset_v2.cpp
```diff
@@ -0,0 +1,61 @@
...
+// /*
+//  * Not tested: append(), appendValue(), capacity(), erase(), eraseBack(),
+//  * insert(), insertValue(), reserve().
+//  */
+//
+// swap() is not working
+/*
+ * Not tested: capacity(), eraseBack(), replace(), insert(), insertValue(), replace().
+ */
+
```
rrahn: Please, check if the commented code is still needed.

include/seqan/sequence/string_set_base.h
h-2: will do.

include/seqan/sequence/string_set_base.h
h-2: will do.

include/seqan/sequence/string_set_base.h
h-2: Because it might not be a sequential container at all?
[this is the most generic case, owner stringsets have specializations]

include/seqan/sequence/string_set_base.h
h-2: I wasn't aware that this is such an issue. I will add Standard() of course, but we should fix this another way, as it is very cumbersome.

include/seqan/sequence/string_set_concat_direct.h
h-2: true!

include/seqan/sequence/string_set_concat_direct.h
h-2: Ah, I forgot about this one. I will reactivate it and fix it.

include/seqan/sequence/string_set_concat_direct.h
h-2: Yes, thats right.

include/seqan/sequence/string_set_owner.h
h-2: To be honest, I am not sure this is completely the same, as this intricacies of `AppendString_` and the entire arrayCopy* is not very straight-forward... I would prefer not do any major changes there now, as we want to re-evaluate the whole container situation anyway, right?

tests/sequence/test_stringset_v2.cpp
h-2: yep

tests/sequence/test_stringset_v2.cpp
h-2: will do

include/seqan/sequence/string_set_concat_direct.h
```diff
@@ -334,6 +336,51 @@ template <typename TString, typename TSpec >
...
+
+    TSize const oldLimLength = length(me.limits);
+    TSize const oldLength = back(me.limits);
+
+    append(me.concat, obj.concat, Tag<TExpand>());
+    append(me.limits, suffix(obj.limits, 1), Tag<TExpand>());
+
+    for (TIt it = begin(me.limits, Standard()) + oldLimLength, itEnd = end(me.limits, Standard()); it != itEnd; ++it)
+        *it += oldLength;
+}
```
h-2: I tried doing this with `seqan::transform` but ran into a ridiculous amount of issues in different places which I don't have the time to fix(I couldn't fix them in 2h).

h-2:
If there are no objections until 17:00 today, I will merge it :)

PR #948, opened 2015-04-13
Fix/898

### LINKED ISSUE #898: Impossible to read GFF with comments or directives with GffFileIn (seqan 2.0.0.0)
Trying to read Gff with comment (a line begin with '#' char according to http://www.sequenceontology.org/gff3.shtml) or directives (begin with "##") throws an error : 
```
Unexpected character '\n' found.
```

You can use the Gff_IO demo code (gff_stream_read.cpp) : 
``` c++
#include <seqan/basic.h>
#include <seqan/gff_io.h>

using namespace seqan;

// USAGE: gff_stream_read IN.gff
//
// Read GFF file and print the positions (reference, position) to stdout.

int main(int argc, char ** argv)
{
    if (argc != 2)
    {
        std::cerr << "USAGE: " << argv[0] << " IN.gff\n";
        return 1;
    }

    try
    {
        GffFileIn gffIn(argv[1]);
        GffRecord record;
        while (!atEnd(gffIn))
        {
            readRecord(record, gffIn);

            // Note that we print the position 1-based since we use text output
            // whereas it is 0-based in the GffRecord.
            std::cout << record.ref << "\t" << (record.beginPos + 1) << "\n";
        }
    }
    catch (Exception & e)
    {
        std::cerr << e.what() << std::endl;
        return 1;
    }

    return 0;
}
```

with a GFF3 file with comment (or directive in this case) : 
```
##gff-version 3
ctg123	.	mRNA	1300	9000	.	+	.	ID=mrna0001;Name="sonichedgehog;hehe"
ctg123	.	exon	1300	1500	.	+	.	ID=exon00001;Parent=mrn a0001;Name
ctg123	.	exon	1050	1500	.	+	.	ID=exon00002;Name;Parent=mrna0001
```
I am not sure but I think the problem is in gff_io_base.h in function `readRecord`, at line 354  
``` c++
skipUntil(iter, NotFunctor<OrFunctor<EqualsChar<'#'>, IsWhitespace> >());  //skip commments and empty lines
```

In the case of a comment (or directive), it just skip '#' chars without going to the next line and so the next part of the function try to parse the comment as an GFF entry.

xenigmax:
This will close #898 . The lines staring with '#' will be passed.


xenigmax:
I will make another pull request including Enrico's suggestion about test cases.

PR #962, opened 2015-04-16
[FEATURE] add numGaps and alignmentLength members

h-2:
This adds two members to the AlignmentStats class, the absolute number of gap characters and the length of the alignment. The addition of the latter is self-explanatory, but why do we need numGaps, if numGaps always equal numGapOpens + numGapExtensions?
Well, most applications don't have information on gap openings and it would be very confusing to copmute around with them.
FWIW, I think numGapExtensions could be removed, as the opposite is true: every application that has information about the number of openings and the number of extensions, automatically has information about the total number of gap characters. However, since that would break API, I didn't remove it, yet.

since the file is @holtgrewe 's baby, I assigned it to him :)

### DIFF
--- a/include/seqan/align/evaluate_alignment.h
+++ b/include/seqan/align/evaluate_alignment.h
@@ -67,6 +67,9 @@ namespace seqan {
  *
  * All members are initialized to <tt>0</tt>.
  *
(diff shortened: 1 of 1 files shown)


holtgrewe:
:+1:

Issue #828, opened 2015-02-05
Segfault when writing SAM with an empty context.


rrahn:
When the ```BamIOContext``` of the ```BamFileOut``` is empty and one tries to write the file in SAM format it causes a segfault, when trying to access the empty members of the context.
Instead one should throw an assertion, if one tries to write the SAM file with an empty context.

It is even more dangerous with BAM files, since writing a BAM file with an empty context apparently works, though the result is not correct and trying to read this file could cause some real troubles that might be falsely associated with reading and not with writing the file. So here as well we need an assertion.

rrahn:
Fixed by #999 and #1000.

PR #964, opened 2015-04-16
[FIXED-#828] Adds assertion to writeRecord of Bam/Sam IO

rrahn:
Given the nature of the FormattedFile it would be in general possible to create a Bam/Sam file with an empty IO context. By the Sam specifications this might be valid (contains only unmapped reads), so we need a way to assert an incomplete IO context, which can happen easily as demonstrated by the following example taken from demos/tutorial/bam_io/solution1.cpp:

```cpp
BamFileIn bamIn("example.bam")
// some error checking.
BamFileOut bamOut(cout, Sam());
// Reading bamFile and writing sam file to console (as demonstrated by the demo above.) 
```
The code is valid but yields an error, since the context (contig names) of the ```bamOut``` is not set. This leads to a misleading error when writing Sam and even worse no error when writing in Bam format. The error would first appear, when one tries to read in from the erroneous bam file. 

This fix simply adds an assertion when writing Sam/Bam file that checks if the ```rID``` and the ```rNextID``` points to a valid contig name according to the Sam specifications.
This way the user can detect the error more easily.

### DIFF
--- a/include/seqan/bam_io/write_bam.h
+++ b/include/seqan/bam_io/write_bam.h
@@ -244,6 +244,7 @@ void write(TTarget & target,
     // Check for valid IO Context.
     SEQAN_ASSERT_LT_MSG(record.rID, static_cast<__int32>(length(contigNames(context))), "BAM IO Assertion: Unknown REF ID!");
     SEQAN_ASSERT_LT_MSG(record.rNextId, static_cast<__int32>(length(contigNames(context))), "BAM IO Assertion: Unknown NEXT REF ID!");
+    ignoreUnusedVariableWarning(context);
 
     // Update internal lengths
     __uint32 size = updateLengths(record);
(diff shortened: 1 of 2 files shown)


esiragusa:
Hey, https://github.com/seqan/seqan/blob/master/demos/tutorial/bam_io/solution1.cpp is okay... where did you get the example?

rrahn:
Oh sorry, for the misleading link.
The demo has been fixed already (see: https://github.com/seqan/seqan/commit/d7b4c07afb01b9eee030b320eb46ee861c1b0b5d). But the general problem persists, as it is easily overseen to set the context during construction and might confuse other users as well.

esiragusa:
@rrahn Can you please make another PR for master? Then we can merge them both.

rrahn:
Closed in favour of #999.

PR #938, opened 2015-04-10
blast module preview

h-2:
This is a preview of the almost-final blast module. Lots of sweat and tears. I rewrote most of it last year and in the last weeks again, so that it behaves like the new IO from @weese and @esiragusa . Now everything is contexitified and there are FormattedFile interfaces that offer almost all of the functionality. The code base is a lot smaller now as well, but it does have a few more if-checks than it used to.

There are still a bunch of smaller TODOs, some missing examples and a missing Tutorial that I want to write, but maybe you can have a look at it sometime soon anyway, @esiragusa @rrahn .

[just to make it clear: this pull shall not be merged, it's just for getting comments from you; there are also some related commits that I need to get merged before this can go in]

PR #979, opened 2015-04-29
[FEATURE] blast module & activation of demo builds & fixes to demos 

h-2:
see #938 for a description.

@rrahn + @esiragusa although there are still some TODOs, please comment!


h-2:
The only big thing in here not blast is the #773, please review that first.

demos/blast/blast_in_example.cpp
```diff
@@ -0,0 +1,43 @@
...
+int main()
+{
+    typedef Align<String<AminoAcid>, ArrayGaps> TAlign;
+    typedef BlastMatch<CharString, CharString, uint32_t, TAlign> TBlastMatch;
+    typedef BlastRecord<CharString, CharString, uint32_t, TAlign> TBlastRecord;
+    typedef BlastIOContext<Blosum62> TContext;
+
+    std::string inPath = std::string(SEQAN_PATH_TO_ROOT()) + "/tests/blast/plus_header_defaults.blast";
+
+    BlastTabularIn<TContext> in(toCString(inPath));
```
esiragusa: Should be called just BlastFileIn.

demos/blast/blast_in_example.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <seqan/blast.h>
+
+using namespace seqan;
+
+int main()
+{
+    typedef Align<String<AminoAcid>, ArrayGaps> TAlign;
+    typedef BlastMatch<CharString, CharString, uint32_t, TAlign> TBlastMatch;
+    typedef BlastRecord<CharString, CharString, uint32_t, TAlign> TBlastRecord;
+    typedef BlastIOContext<Blosum62> TContext;
```
esiragusa: Do you have to know in advance the correct scoring scheme? Is Blosum62 reasonable as default?

demos/blast/blast_in_example.cpp
```diff
@@ -0,0 +1,43 @@
+#include <iostream>
+#include <seqan/blast.h>
+
+using namespace seqan;
+
+int main()
```
esiragusa: If you add a tutorial in the manual, please reuse this demo - don't make a new one.

demos/blast/blast_in_example.cpp
```diff
@@ -0,0 +1,43 @@
...
+            std::cout << "There were non critical errors when reading the record:\n";
+            write(std::cout, context(in).conformancyErrors);
+            std::cout << "\n";
+        }
+
+        std::cout << "\n";
+    }
+
+    return 0;
+}
```
esiragusa: Please add a new line at the end of all your files.

demos/blast/blast_in_example.cpp
```diff
@@ -0,0 +1,43 @@
+#include <iostream>
+#include <seqan/blast.h>
+
+using namespace seqan;
+
+int main()
+{
+    typedef Align<String<AminoAcid>, ArrayGaps> TAlign;
+    typedef BlastMatch<CharString, CharString, uint32_t, TAlign> TBlastMatch;
```
esiragusa: Please change all types from (u)intXX_t to __(u)intXX - see e.g. https://github.com/seqan/seqan/blob/master/include/seqan/platform/platform_gcc.h

demos/blast/blast_in_example.cpp
esiragusa: You should catch the exceptions thrown by the constructor.

demos/blast/blast_in_example.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+        // print some diagnostics
+        std::cout << "Record of query sequence \"" << record.qId << "\"\n"
+                  << "==========================================\n"
+                  << "Number of HSPs: " << length(record.matches) << "\n";
+        if  (length(record.matches) > 0)
+            std::cout << "E-Value of best HSP: " << front(record.matches).eValue << "\n";
+
+        // if there is anything unexpected, tell the user about it
+        if (length(context(in).conformancyErrors) > 0)
```
esiragusa: !empty(context(in).conformancyErrors)

tests/blast/test_blast_statistics.h
```diff
@@ -0,0 +1,280 @@
...
+        // TEST get*
+        SEQAN_ASSERT_EQ(getLambda(scheme), 0.267);
+        SEQAN_ASSERT_EQ(getKappa(scheme),  0.041);
+        SEQAN_ASSERT_EQ(getH(scheme),      0.140);
+        SEQAN_ASSERT_EQ(getAlpha(scheme),  1.900);
+        SEQAN_ASSERT_EQ(getBeta(scheme), -30.000);
+    }
+
+    // Blosum45
+    {
```
esiragusa: You could have refactored this test...

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+#define NOHEADER_DEFAULTS      "/tests/blast/noheader_defaults.m8"
+// same for HEADER and NOHEADERS:
+#define NOHEADER_CUSTOM        "/tests/blast/noheader_custom.m8"
+
+using namespace seqan;
+
+template <typename TNum,
+          typename std::enable_if<Is<NumberConcept<TNum>>::VALUE, int>::type = 0>
+inline void
+clear(TNum & num)
```
esiragusa: Why here in the test file?

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+{
+    num = 0;
+}
+
+template <typename TArg,
+          typename... TArgs>
+inline void
+clear(TArg & arg, TArgs & ... args)
```
esiragusa: As above.

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+{
+    clear(arg);
+    clear(args...);
+}
+
+void _test_blast_read_tabular_match(std::string const & path,
```
esiragusa: We use camel case for functions...

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+                                    bool const defaults,
+                                    bool const withheader,
+                                    bool const islegacy)
+{
+    // only used when defaults == false
+    typedef BlastMatchField<> TField;
+
+    std::string inPath = std::string(SEQAN_PATH_TO_ROOT()) + std::string(path);
+
+    std::ifstream ifstream(toCString(inPath),
```
esiragusa: Please, all tests should test the public FormattedFile interface only.

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+                // actually BITSCORE is also in the file, but we are checking
+                // that the implementation copes with less requested fields
+                // than available ones which we defined as legal.
+            }
+        };
+        context.fields = customFields;
+    }
+
+    // read header of file
+    readHeader(context, it, BlastTabular());
```
esiragusa: Why does readHeader() need a format tag?

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+    {
+        // no strings here to take the strings
+        readMatch(it, BlastTabularLL(), field3, field4, field5, field6, field7, field8, field9, field10, field11,
+                  field12);
+    } SEQAN_CATCH (BadLexicalCast const & e)
+    {
+        exceptComment = e.what();
+    }
+
+    SEQAN_ASSERT(startsWith(exceptComment, "Unable to convert"));
```
esiragusa: Do not assert assertion comments.

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+            TField::Enum::S_END,
+            TField::Enum::FRAMES,
+            TField::Enum::BIT_SCORE
+        }
+    };
+
+    std::string inPath = std::string(SEQAN_PATH_TO_ROOT());
+    inPath += (custom ? PLUS_HEADER_CUSTOM : PLUS_HEADER_DEFAULTS);
+
+    std::ifstream ifstream(toCString(inPath),
```
esiragusa: As above, please test through the public FormattedFile interface only.

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+
+    // read header of file
+    readHeader(context, it, BlastTabular());
+    SEQAN_ASSERT(getBlastTabularSpec(context) == BlastTabularSpec::HEADER);
+
+    // first line is record header
+    SEQAN_ASSERT(!onMatch(context, BlastTabular()));
+    SEQAN_ASSERT(onRecord(context, BlastTabular()));
+
+    readRecordHeader(r, it, context, BlastTabular());
```
esiragusa: A function called readRecordHeader() seems very confusing, as we have already readRecord() and readHeader(). Also, does it need to be public?

tests/blast/test_blast_input.h
```diff
@@ -0,0 +1,1166 @@
...
+{
+    _test_blast_read_tabular_with_header(false);
+}
+
+SEQAN_DEFINE_TEST(test_blast_read_header_customfields_tabular_with_header)
+{
+    _test_blast_read_tabular_with_header(true);
+}
+
+SEQAN_DEFINE_TEST(test_blast_read_header_tabular_with_header_legacy)
```
esiragusa: Can't you refactor all these tests?

esiragusa: Please remember to:
* Test only public interfaces
* Stick to **small and generic** unit tests

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+template <typename TFile, typename TScore, typename TString, typename TRecords, BlastProgram p, BlastTabularSpec h>
+inline void
+test_blast_write_do(TFile & file,
+                    BlastIOContext<TScore, TString, p, h> & context,
+                    TRecords const & records,
+                    int const format,
+                    int const,
+                    BlastReport const & /**/)
+{
+    if (format < 3)
```
esiragusa: What is format < 3? Do you mean <3? :trollface:

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+    {
+        writeHeader(file, context, BlastReport());
+        for (auto const & r : records)
```
esiragusa: Don't we have writeRecord**s**()?

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+            writeRecord(file, context, r, BlastReport());
+        writeFooter(file, context, BlastReport());
+    } else
+    {
+        BlastReportOut<BlastIOContext<TScore, TString, p, h>> out(context, file, BlastReport());
```
esiragusa: Should be called BlastFileOut.

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+template <typename TFile, typename TScore, typename TString, typename TRecords, BlastProgram p, BlastTabularSpec h>
+inline void
+test_blast_write_do(TFile & file,
+                    BlastIOContext<TScore, TString, p, h> & context,
+                    TRecords const & records,
+                    int const format,
+                    int const custom,
+                    BlastTabular const & /**/)
+{
+    switch (format)
```
esiragusa: :bomb:

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+            if ((format == 1) && (custom <= 1))
+                writeHeader(file, context, BlastTabular()); // noop for TABULARs
+            for (auto const & r : records)
+            {
+                if ((format == 1) && (custom <= 1))
+                    writeRecordHeader(file, context, r, BlastTabular());
+
+                for (auto const & m : r.matches)
+                {
+                    if (custom <= 1)
```
esiragusa: :spaghetti:

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+    appendValue(qIds, "Query_Numero_Tres with args");
+
+    setScoreGapOpenBlast(context.scoringScheme, -11);
+    setScoreGapExtend(context.scoringScheme, -1);
+    SEQAN_ASSERT(isValid(context.scoringScheme));
+
+    String<TBlastRecord> records;
+    resize(records, 3);
+
+    context.dbName = "The Foo Database";
```
esiragusa: Note that other FormattedFileContextes (e.g. http://docs.seqan.de/seqan/develop/class_BamIOContext.html, http://docs.seqan.de/seqan/develop/class_VcfIOContext.html) already have standardized accessors for these properties. Please stick to that interface.

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+    appendValue(qIds, "Query_Numero_Uno with args");
+    appendValue(qIds, "Query_Numero_Dos with args");
+    appendValue(qIds, "Query_Numero_Tres with args");
+
+    setScoreGapOpenBlast(context.scoringScheme, -11);
+    setScoreGapExtend(context.scoringScheme, -1);
+    SEQAN_ASSERT(isValid(context.scoringScheme));
+
+    String<TBlastRecord> records;
+    resize(records, 3);
```
esiragusa: FYI this will call the Generous resize.

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+
+    // sort by bit-score
+    for (auto & r : records)
+        r.matches.sort();
+
+    test_blast_write_do(file, context, records, format, customFields, TFormat());
+}
+
+template <typename TScore, typename TString, BlastProgram p, BlastTabularSpec h>
+void test_blast_write_tabular_impl(int const format, // 0 only matches; 1 matches+headers; 2 records; 3 formattedFile
```
esiragusa: Please don't use these numbers.

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+    versionLine.append("[I/O Module of SeqAn-");
+    versionLine.append(std::to_string(SEQAN_VERSION_MAJOR));
+    versionLine.append(".");
+    versionLine.append(std::to_string(SEQAN_VERSION_MINOR));
+    versionLine.append(".");
+    versionLine.append(std::to_string(SEQAN_VERSION_PATCH));
+    versionLine.append(", http://www.seqan.de]\n");
+
+    std::string fieldsLine = "# Fields: ";
+    if (context.legacyFormat)
```
esiragusa: The following 100 lines of code are hard to understand... :spaghetti:

tests/blast/test_blast_output.h
```diff
@@ -0,0 +1,814 @@
...
+                break;
+            }
+        }
+    }
+    SEQAN_ASSERT_EQ(contents, compString);
+}
+
+// TEST MATCH, WITHOUT HEADER, DEFAULTS FIELDS
+
+SEQAN_DEFINE_TEST(test_blast_write_match_tabular_run_time_context_args)
```
esiragusa: We have a generic testing framework. The idea was to avoid repeating tests like in the following 200 lines.

include/seqan/blast/blast_tabular_out.h
```diff
@@ -0,0 +1,799 @@
...
+            write(s, match.sLength);
+            break;
+        case BlastMatchField<>::Enum::Q_START:
+        {
+            TPos effectiveQStart    = match.qStart;
+            TPos effectiveQEnd      = match.qEnd;
+            _untranslateQPositions(effectiveQStart, effectiveQEnd, match.qFrameShift, match.qLength,
+                                   context.blastProgram, BlastProgramTag<p>());
+            write(s, effectiveQStart);
+        } break;
```
esiragusa: Breaks go on a new line.

include/seqan/blast/blast_tabular_out.h
```diff
@@ -0,0 +1,799 @@
...
+            TPos effectiveSEnd      = match.sEnd;
+            _untranslateSPositions(effectiveSStart, effectiveSEnd, match.sFrameShift, match.sLength,
+                                   context.blastProgram, BlastProgramTag<p>());
+            write(s, effectiveSEnd);
+        } break;
+//         case ENUM::Q_SEQ: write(s,  * ); break;
+//         case ENUM::S_SEQ: write(s,  * ); break;
+        case BlastMatchField<>::Enum::E_VALUE:
+        {
+            std::string formatString;
```
esiragusa: Please (re)use a buffer in the context.

include/seqan/blast/blast_tabular_out.h
```diff
@@ -0,0 +1,799 @@
...
+            else if (match.eValue < 10.0)
+                formatString = "%2.1lf";
+            else
+                formatString = "%5.0lf";
+
+            write(s, FormattedNumber<double>(formatString.c_str(), match.eValue));
+        } break;
+        case BlastMatchField<>::Enum::BIT_SCORE:
+        {
+            std::string formatString;
```
esiragusa: See above.

include/seqan/blast/blast_tabular.h
```diff
@@ -0,0 +1,661 @@
...
+        "Query Coverage Per Subject",
+        "Query Coverage Per HSP"
+    };
+
+    /*!
+     * @var static_constexpr_bool_const BlastMatchField::implemented[]
+     * @brief An array of bools revealing whether the Blast I/O module supports printing this field
+     */
+    //TODO(c++14): change to std::bitset that is initialized with binary literal
+    static constexpr bool const implemented [] =
```
esiragusa: I think there is a false missing at line 610 :trollface:

include/seqan/blast/blast_tabular.h
```diff
@@ -0,0 +1,661 @@
...
+        "Query id, Subject id, % identity, alignment length, mismatches, gap openings, q. start, q. end, s. start, s."
+         " end, e-value, bit score"
+    };
+
+    /*!
+     * @var static_constexpr_char_const_*_const BlastMatchField::columnLabels[]
+     * @brief An array of CStrings representing the <b>column label</b> of each possible field; for the
+     * @link BlastIOContext::legacyFormat @endlink, use @link BlastMatchField::legacyColumnLabels @endlink instead.
+     */
+    static constexpr char const * const columnLabels [] =
```
esiragusa: Why constexpr??

include/seqan/blast/blast_tabular.h
```diff
@@ -0,0 +1,661 @@
...
+ * <tr><td>45</td><td>Q_COV_HSP</td><td>qcovhsp</td><td>% hsp coverage</td><td>Query Coverage Per HSP</td><td>&#9744;</td></tr>
+ * </table></span>
+ * @endhtmlonly
+ *
+ * More fields will likely be implemented in the future.
+ */
+
+
+template <typename TVoidSpec = void>
+struct BlastMatchField
```
esiragusa: I think we should discuss together how to represent this stuff in a better way.

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+ * @val BlastProgram BlastProgram::TBLASTN
+ * @brief Protein Query VS translated Nucleotide Subject
+ *
+ * @val BlastProgram BlastProgram::TBLASTX
+ * @brief translated Nucleotide Query VS translated Nucleotide Subject
+ *
+ * @val BlastProgram BlastProgram::UNKNOWN
+ * @brief Unkown type. Used internally and to signify that the type could not be inferred from the file.
+ */
+enum class BlastProgram : uint8_t
```
esiragusa: These should have been file format tags.

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+// ----------------------------------------------------------------------------
+
+/*
+ * Given both an enum value and a tag, return the tag's value unless it is UNKNOWN
+ * in which case the enum's value is returned. Can be used in implementations that
+ * want to cover compile-time constants and run-time settings both.
+ */
+
+// this is always evaluated at compile-time
+template <BlastProgram p>
```
esiragusa: Template argument must be prefixed by T, e.g. TBlastProgram.

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+constexpr BlastProgram
+getBlastProgram(BlastProgram const, BlastProgramTag<p> const &)
+{
+    return p;
+}
+
+// this will likely be evaluated at run-time
+constexpr BlastProgram
+getBlastProgram(BlastProgram const p, BlastProgramTag<BlastProgram::UNKNOWN> const &)
```
esiragusa: What's different from the above function?

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+ * If both a tag and an enum are specified, the tag is given preference, unless it is set to
+ * @link BlastProgram::UNKNOWN @endlink.
+ *
+ * @return false for @link BlastProgram::BLASTP @endlink and @link BlastProgram::TBLASTN @endlink
+ * @return true otherwise
+ * @headerfile <seqan/blast.h>
+ */
+
+constexpr bool
+qHasRevComp(BlastProgram const p)
```
esiragusa: These should have been metafunctions.

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+{
+    return ((p==BlastProgram::BLASTP) || (p==BlastProgram::TBLASTN))
```
esiragusa: return !(p == BlastProgram::BLASTP || p == BlastProgram::TBLASTN)

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+ * If both a tag and an enum are specified, the tag is given preference, unless it is set to
+ * @link BlastProgram::UNKNOWN @endlink.
+ *
+ * @return true for @link BlastProgram::BLASTX @endlink and @link BlastProgram::TBLASTX @endlink
+ * @return false otherwise
+ * @headerfile <seqan/blast.h>
+ */
+
+constexpr bool
+qIsTranslated(BlastProgram const p)
```
esiragusa: Previous comments apply to all functions below. In addition, the code below is extremely repetitive...

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+_programTagToString<BlastProgram::TBLASTX>()
+{
+    return "TBLASTX";
+}
+
+// if not compile time fixed, than we can't operate with const char * and need std::string instead
+inline std::string const
+_programTagToString(BlastProgram const _p)
+{
+    switch(_p)
```
esiragusa: What about mapping the types to a static const char array?

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+
+template <BlastProgram p>
+constexpr bool
+sNumFrames(BlastProgram const _p, BlastProgramTag<p> const &)
+{
+    return sNumFrames(getBlastProgram(_p, BlastProgramTag<p>()));
+}
+
+// ----------------------------------------------------------------------------
+// _programTagToString()
```
esiragusa: In general, I don't see what you are trying to do here with _programTagToString() and why it is so complicated...

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+
+// ----------------------------------------------------------------------------
+// _programStringToTag()
+// ----------------------------------------------------------------------------
+
+template <typename TString>
+inline BlastProgram
+_programStringToTag(TString const & str)
+{
+    if (str == _programTagToString<BlastProgram::BLASTN>())
```
esiragusa: What about searching the string back in the static const char array?

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+
+// ----------------------------------------------------------------------------
+// Function _firstOcc
+// ----------------------------------------------------------------------------
+
+template <typename TString>
+inline typename Size<TString>::Type
+_firstOcc(TString const & str, typename Value<TString>::Type const & val)
+{
+    typedef typename Size<TString>::Type S;
```
esiragusa: Typenames start with T

include/seqan/blast/blast_base.h
```diff
@@ -0,0 +1,669 @@
...
+        _untranslatePositions(effectiveStart, effectiveEnd, frameShift, length, False(), False());
+}
+
+// ----------------------------------------------------------------------------
+// Function _firstOcc
+// ----------------------------------------------------------------------------
+
+template <typename TString>
+inline typename Size<TString>::Type
+_firstOcc(TString const & str, typename Value<TString>::Type const & val)
```
esiragusa: Isn't this std::find?

include/seqan/blast/blast_io_context.h
```diff
@@ -0,0 +1,383 @@
...
+ * context(outfile).dbName = "The Foo Database";
+ * context(outfile).dbTotalLength = length(concat(subjects));
+ * context(outfile).dbNumberOfSeqs = length(subjects);
+ * @endcode
+ *
+ * See @link BlastTabularOut @endlink and @link BlastReportOut @endlink for more complete examples of usage.
+ */
+
+template <typename TScore_ = Blosum62,
+          typename TString_ = std::string,
```
esiragusa: Is it important to change TString?

include/seqan/blast/blast_io_context.h
```diff
@@ -0,0 +1,383 @@
...
+          BlastProgram _p = BlastProgram::UNKNOWN,
```
esiragusa: These should have been file format tags.

include/seqan/blast/blast_io_context.h
```diff
@@ -0,0 +1,383 @@
...
+    TString versionString;
+    void _setDefaultVersionString()
+    {
+        clear(versionString);
+        append(versionString, _programTagToString(blastProgram, TBlastProgram()));
+        append(versionString, " 2.2.26");
+        if (!legacyFormat)
+            append(versionString, "+");
+        append(versionString, " [I/O Module of SeqAn-");
+        append(versionString, std::to_string(SEQAN_VERSION_MAJOR));
```
esiragusa: This doesn't work if TString is seqan::CharString.

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+/*!
+ * @class BlastMatch
+ * @implements AssignableConcept
+ * @implements CopyConstructibleConcept
+ * @implements DefaultConstructibleConcept
+ * @implements EqualityComparableConcept
+ * @implements LessThanComparableConcept
+ * @headerfile <seqan/blast.h>
+ * @signature struct BlastMatch<TQId, TSId, TPos, TAlign> { ... };
+ * @brief An data structure to hold a blast match, also known as high-scoring segment pair (HSP)
```
esiragusa: Why don't you call it HSP then?

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+ *
+ * @tparam TQId  Type of qId, defaults to std::string
+ * @tparam TSId  Type of sId, defaults to std::string
+ * @tparam TPos  Position type of the sequences, defaults to <tt>uint32_t</tt>
+ * @tparam TAlign Type of the @link Align @endlink member, defaults to
+ * <tt>Align<CharString, ArrayGaps></tt>
+ *
+ */
+
+template <typename TQId = std::string,
```
esiragusa: Do we need all these template parameters? Maybe only the ArrayGaps spec?

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+struct BlastMatch
+{
+    /*!
+     * @var TQId BlastMatch::qId;
+     * @brief The verbose Id of the query.
+     *
+     * @var TSId BlastMatch::sId;
+     * @brief The verbose Id of the subject.
+     */
+    TQId            qId;
```
esiragusa: Why don't you factorize out query and subject information? You repeat id, start, end, length, frameshift for both...

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+     * To facilitate fast sorting of matches in a @link BlastRecord @endlink, only the bit-score is compared. Also
+     * large bit-score are sorted to front (i.e. operator< on BlastMatch checks operator>= on the bitScores).
+     */
+
+    inline bool operator< (BlastMatch const & bm2) const
+    {
+        return (bitScore >= bm2.bitScore);
+    }
+
+    inline void _clear()
```
esiragusa: Why do you define it as a member function?

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+     * @var TPos BlastRecord::qLength;
+     * @brief length of the query sequence
+     */
+    TPos            qLength;
+
+    /*!
+     * @var std::list<TBlastMatch> BlastRecord::matches;
+     * @brief list of the contained matches
+     */
+    std::list<TBlastMatch>  matches;
```
esiragusa: Really a std::list? :scream:

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+_memberIsSet(CharString const & in)
+{
+    return (in != "not init");
+}
+
+template <typename TNumber>
+inline SEQAN_FUNC_ENABLE_IF(Is<NumberConcept<TNumber> >, bool)
+_memberIsSet(TNumber const & in)
+{
+    return (in != std::numeric_limits<TNumber>::max());
```
esiragusa: FYI you don't need parentheses here nor above...

include/seqan/blast/blast_record.h
```diff
@@ -0,0 +1,384 @@
...
+ * @brief A record of blast-matches (belonging to one query).
+
+ * @tparam TQId  Type of qId, defaults to std::string
+ * @tparam TSId  Type of sId, defaults to std::string
+ * @tparam TPos  Position type of the sequences, defaults to <tt>uint32_t</tt>
+ * @tparam TAlign Type of the @link Align @endlink member, defaults to
+ * <tt>Align<CharString, ArrayGaps></tt>
+ */
+
+template <typename TQId = std::string,
```
esiragusa: The same here for these template parameters.

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+ *
+ * The file generated in /tmp/output.blast looks like this:
+ *
+ * @include demos/blast/blast_out_example.report
+ *
+ * @see BlastRecord
+ */
+
+template <typename TBlastIOContext = BlastIOContext<> >
+using BlastReportOut = FormattedFile<BlastReport, Output, TBlastIOContext>;
```
esiragusa: This will become a simple typedef when the BlastIOContext goes away.

esiragusa: Should be called BlastFileOut.

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+
+// ----------------------------------------------------------------------------
+// Class MagicHeader
+// ----------------------------------------------------------------------------
+
+template <typename T>
+struct MagicHeader<BlastReport, T> :
+    public MagicHeader<Nothing, T> {};
```
esiragusa: Really? Isn't it a string like BLAST*?

esiragusa: FYI the public keyword is superfluous when you inherit a struct.

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+// ============================================================================
+
+// ----------------------------------------------------------------------------
+// Function guessFormat()
+// ----------------------------------------------------------------------------
+
+template <typename TSpec>
+inline bool guessFormat(FormattedFile<BlastReport, Output, TSpec> &)
+{
+    return true;
```
esiragusa: So there's no format guessing?

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+{
+    return (number == 0) ? 1 : static_cast<T>(std::floor(std::log10(number))+1);
+}
+
+// ----------------------------------------------------------------------------
+// some string constants
+// ----------------------------------------------------------------------------
+
+constexpr const char *
+_blastReference()
```
esiragusa: Why don't you make them static const strings then?

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+          typename TMatch,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline void
+_writeStatsBlock(TStream & stream,
+                 BlastIOContext<TScore, TConString, p, h> const &,
+                 TMatch const & m,
+                 BlastReport const &)
+{
+    char buffer[512] = "";
```
esiragusa: Why do you do = "" ?

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+
+    sprintf(buffer," Score =  %.1f bits (%d), Expect =  %.1g\n"
```
esiragusa: Why do you use a buffer + sprintf()?

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+    auto const & row1        = row(m.align, 1);
+
+    TPos const   maxPos      = std::max({effQStart, effQEnd, effSStart, effSEnd});
+    // max # digits in pos's
+    unsigned char const numberWidth = _numberOfDigits(maxPos);
+
+    while (aPos < m.alignStats.alignmentLength)
+    {
+        // Query line
+        sprintf(buffer, "Query  %-*d  ", numberWidth, qPos + effQStart);
```
esiragusa: Again, why do you use a buffer + sprintf()?

include/seqan/blast/blast_report_out.h
```diff
@@ -0,0 +1,893 @@
...
+                    BlastReport const & /*tag*/)
+{
+    if (length(m.sId) == 66) // it fits
+    {
+        write(stream, m.sId);
+    }
+    else if (length(m.sId) < 66) // needs to be padded with ' '
+    {
+        write(stream, m.sId);
+        for (unsigned char i = 0; i < 66 -length(m.sId); ++i)
```
esiragusa: Use simply unsigned.

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+// not implemented in blast
+
+/// PAM250
+template <typename TSpec>
+struct KarlinAltschulValues<Pam250, TSpec>
+{
+    typedef uint8_t size_t;
+
+    /* statics */
+    static constexpr size_t nParams = 8;
```
esiragusa: Why not just const??

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+// not implemented in blast
+
+/// PAM200
+// not implemented in blast
+
+/// PAM250
+template <typename TSpec>
+struct KarlinAltschulValues<Pam250, TSpec>
+{
+    typedef uint8_t size_t;
```
esiragusa: TSize?

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+ */
+template <typename TScore>
+struct BlastScoringScheme
+{
+    typedef typename Value<TScore>::Type TValue;
+    typedef typename KarlinAltschulValues<TScore, void>::size_t TNumValues;
+
+    /* related to ScoringScheme conversion */
+    TValue _gapOpenBlast = 0;
+    TScore _internalScheme;
```
esiragusa: Shouldn't there be a constructor to pass a scoring scheme?

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+/* NOTE(h-2): Maybe this should specialize Score<> instead of being a wrapper.
+ * This would however involve adding a new template parameter to score and
+ * changing lots of code. Also it would become confusing as the shortcuts
+ * like Blosum62 would no longer work...
+ */
+template <typename TScore>
+struct BlastScoringScheme
+{
+    typedef typename Value<TScore>::Type TValue;
+    typedef typename KarlinAltschulValues<TScore, void>::size_t TNumValues;
```
esiragusa: Use Size/TSize.

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+
+    /* related to ScoringScheme conversion */
+    TValue _gapOpenBlast = 0;
```
esiragusa: Why is it different from the base scoring scheme gap open?

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+    TScore _internalScheme;
+
+    /* parameter selection */
+    TNumValues parameterIndex = std::numeric_limits<TNumValues>::max();
```
esiragusa: What's that?

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+/*!
+ * @fn BlastScoringScheme#getLambda
+ * @headerfile <seqan/blast.h>
+ * @brief Get the &lambda; value of the Karlin-Altschul parameters.
+ * @signature double getLambda(blastScoringScheme);
+ */
+
+template <typename TMatrixSpec>
+inline double
+getLambda(BlastScoringScheme<Score<int, ScoreMatrix<AminoAcid, TMatrixSpec>>> const & scoringScheme)
```
esiragusa: Why do you have **two** versions of all these getters for ScoreMatrix and Simple?

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+    SEQAN_ASSERT(isValid(scoringScheme));
+    return TKAValues::VALUE[scoringScheme.parameterIndex][8];
+}
+
+// ----------------------------------------------------------------------------
+// Function _lengthAdjustment
+// ----------------------------------------------------------------------------
+
+// Computes the length adjustment for E-value computation
+// Based on the public domain NCBI BLAST code by Tom Madden.
```
esiragusa: Please add the original license statement.

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+ *
+ * @section Remarks
+ *
+ * The reference returned is always const, because you should not modify the internal score object
+ * directly. Use @link BlastScoringScheme @endlink's interface instead.
+ */
+
+template <typename TScore>
+inline TScore const &
+seqanScheme(BlastScoringScheme<TScore> const & bscheme)
```
esiragusa: getScoringScheme()?

include/seqan/blast/blast_statistics.h
```diff
@@ -0,0 +1,1076 @@
...
+          typename TConString,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline double
+computeEValue(TBlastMatch & match,
+              BlastIOContext<TScore, TConString, p, h> & context)
+{
+    uint64_t ql = match.qLength; // convert to 64bit, once
+    // length adjustment not yet computed
+    if (context._cachedLengthAdjustments.find(ql) == context._cachedLengthAdjustments.end())
```
esiragusa: Is ql the key or the value? Before you call container.find(ql) and then contaner[ql]...

include/seqan/blast/blast_tabular.h
```diff
@@ -0,0 +1,661 @@
...
+// ----------------------------------------------------------------------------
+// Class FileExtensions
+// ----------------------------------------------------------------------------
+
+template <typename T>
+struct FileExtensions<BlastTabular, T>
+{
+    static constexpr char const * VALUE[5] =
+    {
+        ".blast",
```
esiragusa: Hmm are we using the same set of extensions for different formats?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+ *
+ * @return bool true or false
+ */
+
+template <typename TScore,
+          typename TString,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline bool
+onRecord(BlastIOContext<TScore, TString, p, h> & context,
```
esiragusa: Private?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+ *
+ * @return    true or false
+ */
+
+template <typename TScore,
+          typename TString,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline bool
+onMatch(BlastIOContext<TScore, TString, p, h> & context,
```
esiragusa: Private?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+ *
+ * @throw IOError On low-level I/O errors.
+ * @throw ParseError On high-level file format errors.
+ *
+ * @return bool true or false
+ */
+
+template <typename TContext>
+inline bool
+onRecord(BlastTabularIn<TContext> & formattedFile)
```
esiragusa: Private?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+// Function goNextLine()
+// ----------------------------------------------------------------------------
+
+template <typename TScore,
+          typename TString,
+          typename TFwdIterator,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline void
+goNextLine(BlastIOContext<TScore, TString, p, h> & context,
```
esiragusa: Private?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+        setBlastTabularSpec(context, BlastTabularSpec::HEADER);
+
+    int queryLinePresent = 0;
+    int dbLinePresent = 0;
+    int fieldsLinePresent = 0;
+    int hitsLinePresent = 0;
+
+    do
+    {
+        if (std::regex_search(seqan::begin(context._lineBuffer),
```
esiragusa: Do you read/buffer the file line by line? Seriously? :scream:

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+ * @headerfile seqan/blast.h
+ */
+
+template <typename TFwdIterator,
+          typename TScore,
+          typename TString,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline void
+skipRecordHeader(TFwdIterator & iter,
```
esiragusa: Private?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+          BlastTabularSpec h>
+inline void
+readMatch(BlastMatch<TQId, TSId, TPos, TAlign> & match,
+          TFwdIterator & iter,
+          BlastIOContext<TScore, TString, p, h> & context,
+          BlastTabular const &)
+{
+    if (context.legacyFormat)
+    {
+        if (SEQAN_ENABLE_DEBUG)
```
esiragusa: Why isn't this a preprocessor directive?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+ * @throw ParseError On high-level file format errors.
+ */
+
+template <typename TFwdIterator,
+          typename TScore,
+          typename TString,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline void
+skipMatch(TFwdIterator & iter,
```
esiragusa: What's the purpose of this function?

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+          typename TSId,
+          typename TPos,
+          typename TAlign,
+          typename TFwdIterator,
+          typename TScore,
+          typename TString,
+          BlastProgram p,
+          BlastTabularSpec h>
+inline void
+readRecordHeader(BlastRecord<TQId, TSId, TPos, TAlign> & r,
```
esiragusa: This function name is confusing.

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+_readRecordNoHeader(BlastRecord<TQId, TSId, TPos, TAlign> & blastRecord,
+                    TFwdIterator & iter,
+                    BlastIOContext<TScore, TString, p, h> & context,
+                    BlastTabular const &)
+{
+    ++context.numberOfRecords;
+
+    clear(blastRecord);
+
+    auto it = seqan::begin(context._lineBuffer); // move into line below when seqan supports && properly
```
esiragusa: Why do you use explicitly the namespace seqan? The iter tag is missing.

include/seqan/blast/blast_tabular_in.h
```diff
@@ -0,0 +1,1295 @@
...
+    {
+        blastRecord.matches.emplace_back();
+        // read remainder of line
+        readMatch(back(blastRecord.matches), iter, context, BlastTabular());
+
+        if (!startsWith(context._lineBuffer, curIdPlusTab)) // next record reached
+            break;
+    }
+
+    if (length(blastRecord.matches) == 0)
```
esiragusa: Maybe empty() ?

demos/blast/blast_in_example.cpp
h-2: > Should be called just BlastFileIn.

There are two formats currently implemented: BlastTabular and BlastReport. I could rename the FormattedFiles from BlastTabularOut to BlastTabularFileOut a.s.o
Do you mean that?

> You should catch the exceptions thrown by the constructor.

Yes!

esiragusa: There has to be one BlastFile which supports the two formats TabularBlast and ReportBlast. Like BamFile which supports Sam and Bam.

demos/blast/blast_in_example.cpp
h-2: > Do you have to know in advance the correct scoring scheme?

Yes, it is a required spec for the BlastScoringScheme which is in turn required for Blast IO.

> Is Blosum62 reasonable as default?

Since all BLAST programs other than BLASTN score in protein space and Blosum62 is the default there, I think it should be here, as well.

demos/blast/blast_in_example.cpp
h-2: Yes, I will.

demos/blast/blast_in_example.cpp
h-2: Ah, that one escaped me. Of course all files should end on on newline.

demos/blast/blast_in_example.cpp
h-2: I can change this for now, but I think we should discuss changing it instead in all other places to use the official typedef, since all of our target compilers now have this, I think.

demos/blast/blast_in_example.cpp
h-2: +1

tests/blast/test_blast_statistics.h
h-2: But all ther numbers are actually different? How would you have refactored this?

tests/blast/test_blast_input.h
h-2: Was a leftover from before that isn't needed anymore.

tests/blast/test_blast_statistics.h
esiragusa: Use statically initialized vectors of double?

tests/blast/test_blast_input.h
h-2: Except for test names, you mean? ^^
But true, this is not a test but a function that is called by a test.

tests/blast/test_blast_input.h
h-2: Hm, I can change this, but then we loose a lot of currently tested functionality...

tests/blast/test_blast_input.h
h-2: Because its different for BlastTabular and BlastReport?

tests/blast/test_blast_input.h
h-2: Ok.

tests/blast/test_blast_input.h
h-2: It was publicly documented before, so it is still public. I can change it to private, but I don't see whats wrong with the name. Every record has a header; since SeqAn defines the header to be file-global, I used RecordHeader to show that it is the header of the record. This is also explained in current dox.

tests/blast/test_blast_output.h
h-2: ;)
It is an internal variable that is used to re-use code for different tests. If I remove everything except the FormattedFile interface than this will go away.

tests/blast/test_blast_output.h
h-2: Only SeqFileOut has this I think. At least it is not in the dox for other formats...

tests/blast/test_blast_output.h
h-2: Not really. It has to be templatized somewhere. Only templatizing the context at least moves the variability to one place.
Since downstream functions don't work on the FormattedFile but on the context and the iterator, obviously we can't move sepcializations from the context to the FormattedFile.

tests/blast/test_blast_output.h
h-2: Then we need to discuss these accessors because obviously they were not intended to be general. I don't have contigs, and I can't expect users to know that they have to say "contig" when they mean "protein in protein database".
[But I do agree that have the same accessors is very much desirable]

tests/blast/test_blast_output.h
h-2: see above, it will go away if we only test formattedfile.

tests/blast/test_blast_output.h
h-2: Well, its the construction of the expected output. It can be done with less :spaghetti: but it will mean lots of duplication.

tests/blast/test_blast_output.h
h-2: I haven't done testing on this scale before, can you point me to the relevant documents describing how to do this in a nicer way?

include/seqan/blast/blast_tabular_out.h
h-2: ok.

include/seqan/blast/blast_tabular_out.h
h-2: For strings this short this is guaranteed to be storage from the stack; accessing a potentially larger buffer and overwriting it could actually be slower.
But if we have a policy for this I can change it of course.

include/seqan/blast/blast_tabular.h
h-2: ?

include/seqan/blast/blast_tabular.h
h-2: Because I can :gangnamstyle:

include/seqan/blast/blast_tabular.h
h-2: Hm, sure.

include/seqan/blast/blast_base.h
h-2: As I previously said: before there was one BlastFormat which had multiple template arguments for the fileformate (Tabular, TabularWithHeader, Report, poossibly XML...), the program type (BlastN, BlastP...), the generation (blastall vs blast+). You said that it was super confusing and that it needed to change.
Then I treied it with the TagSelector but obiously this then results in huge confusing number of possible combinations that all get a tag... so instead I want with the above :)

include/seqan/blast/blast_base.h
h-2: But only if its a type, or not? This is a value.

include/seqan/blast/blast_base.h
h-2: one returns left arg, the other returns the right arg?

include/seqan/blast/blast_base.h
h-2: Maybe we should discuss this. As we are targetting c++11 anyway, I don't see think it is better do make things needlessly more complex (however I do see value in consistency).

include/seqan/blast/blast_base.h
h-2: nice catch!

include/seqan/blast/blast_base.h
h-2: You are right, this would be a much better solutation. I will change this.

include/seqan/blast/blast_base.h
h-2: Yes, see above.

include/seqan/blast/blast_base.h
h-2: Yes, see above.

include/seqan/blast/blast_base.h
h-2: Yes.

include/seqan/blast/blast_base.h
h-2: True, I will replace it.

include/seqan/blast/blast_io_context.h
h-2: Not important. Using std::string instead of CharString is a little more efficient in some places, but I though maybe someone would want to use CharString for some reasons.

include/seqan/blast/blast_io_context.h
h-2: see above

include/seqan/blast/blast_record.h
h-2: I thought "match" is a more general term, as it used in other palces in SeqAn, too.

include/seqan/blast/blast_record.h
h-2: In some situations you will want to change the types, e.g. when you have all IDs in a ConcatDirect StringSet, it makes much more sense to also use Infixes for the IDs, so you don't need to copy.

include/seqan/blast/blast_record.h
h-2: I thought it would be more confusing to add another layer of members. And I am not sure it would actually reduce the lines of code or increase its maintainability. But I can change it, sure!

include/seqan/blast/blast_record.h
h-2: could be changed.

include/seqan/blast/blast_record.h
h-2: Well the BlastMatches are rather fat objects and usually not too many. And  you always want to sort your matches before printing;my tests showed that this is measurably faster than with vectors.

include/seqan/blast/blast_record.h
h-2: but it makes for nicer readability, or not?

include/seqan/blast/blast_report_out.h
h-2: > This will become a simple typedef when the BlastIOContext goes away.

I don't see any way of doing this without templatizing the final FormattedFile.

include/seqan/blast/blast_report_out.h
h-2: > Really? Isn't it a string like BLAST*?

It is '> BLAST*' or '> TBLAST*' for tabular with header, it is not defined for tabular without header, it  is 'BLAST*' or 'TBLAST*' for report.

> FYI the public keyword is superfluous when you inherit a struct.

Ah, I didn't know, but it makes sense of course.

include/seqan/blast/blast_report_out.h
h-2: It is the same really, the only difference is that to prevent global inclusion you would have to make the static const cstring a member of a templated struct (which is a hack by itself).

demos/blast/blast_in_example.cpp
esiragusa: Yep we have them with C++11 (see http://www.cplusplus.com/reference/cstdint/). But please change it back for now.

include/seqan/blast/blast_report_out.h
h-2: I think I did something else before and need it is 0-initialized... it can go now.

include/seqan/blast/blast_report_out.h
h-2: Because it would take 3 pages of code to do the same by writing FormattedNumber :o

tests/blast/test_blast_input.h
esiragusa: Which functionality is outside of the FormattedFile interface? Is it relevant? This additional complexity blows up not only tests but also dox and manual.

include/seqan/blast/blast_statistics.h
h-2: Well anything that could be, should be constexpr. If all our types had proper constructors (and the compiler supports c++14), it would be possible to select the KarlinAltschulValues for a given scoring scheme at compile time :)

include/seqan/blast/blast_statistics.h
h-2: yes

include/seqan/blast/blast_statistics.h
h-2: Because of the possibly confusing conversion of open scores I would rather have the people use the set*() functions instead.

include/seqan/blast/blast_base.h
esiragusa: I mean compile time format traits...

include/seqan/blast/blast_base.h
esiragusa: Ah... did you (kind of) reimplement the tag selector for enums?

include/seqan/blast/blast_io_context.h
esiragusa: Then let's remove it from the main template parameters, one could still tweak it through Fibre/Member, or one last optional config/traits object.

include/seqan/blast/blast_record.h
esiragusa: But why do you need a list rather than a vector? Do you insert elements in the middle?

include/seqan/blast/blast_report_out.h
esiragusa: Both score matrix and blast program type have to be format traits rather than io context template parameters!

include/seqan/blast/blast_report_out.h
esiragusa: Global inclusion into what?

include/seqan/blast/blast_report_out.h
esiragusa: Then we should come up with something better...

include/seqan/blast/blast_base.h
h-2: Yes, sort of. But in my case there are not ever any needless checks, if values are selected at compile-time. And even if they are dynamic, there is no recursion (which is bad because it can't be inlined and its also bad for branch prediction I think).

Seeing that there are 5 blast program formats, 3 implemented file types and legacy and plus formats for each there would already be 30 combinations of fileformats to check whenever a record a is written. This is totally not feasible.

h-2:
All the warning that now appear in cash are related to demos. AFAICT the demos where not built automatically before at all, or we would have seen these warning already. I will fix what I can and include it in the pull request.

include/seqan/blast/blast_record.h
h-2: > Do you insert elements in the middle?

No, but swapping is instant in a linked list, becaus only the pointers to the elements are changed. So for fat obejcts this is better than swapping in a vector.

include/seqan/blast/blast_base.h
h-2: > These should have been metafunctions.

They are used both as regular and as constexpr functions, so any other solution would have meant more code, less clarity :)

include/seqan/blast/blast_statistics.h
h-2: I have added a more prominent notice at the top of the file, but a public domain declaration implies that I don't have to do anything, it is not a license and comes with no terms.

include/seqan/blast/blast_statistics.h
h-2: Because the values are in different positions in the matrices.

include/seqan/blast/blast_statistics.h
h-2: I am using the term BlastScoringScheme and SeqanScoringScheme to differentiate between the behaviours, so it has to be clear that this function returns an object with SeqAn behaviour.

include/seqan/blast/blast_report_out.h
h-2: In any case, since the magic header is always ambiguous, I will leave it as "nothing".

include/seqan/blast/blast_report_out.h
h-2: I have removed this in all places, except the one above, because this behaviour cannot be achieved with our `FormattedNumber`.

h-2:
@esiragusa I have fixed lots of stuff and removed comments for issues that we discussed f2f. Can you have a look at the remaining points? Can you also look at the tests again, I have refactored them to use the FormattedFile interface only (except for the two low-level tests).
Maybe this can still be done more generic, but I didn't completely understand it from the sequence_v2 tests.

h-2:
As far as I am concerned this is final, now that the very extensive tutorial is done, as well. 
@esiragusa please comment on any open issues, I want to merge this ASAP. Thanks!

h-2:
If @esiragusa  is busy, can @rrahn have a look? I really want to merge this today so I have tomorrow to fix possible issues discovered during tonights' runs.

PR #910, opened 2015-03-18
[FIX,INTERNAL] make concat with delimiter, strSplit and some internals work with std::strings (develop)

h-2:
see #890

PR #983, opened 2015-04-30
[FIX] add implicit conversions similiar to infix

h-2:
make changes from #910 work for prefixes and suffixes, as well (not only infixes).

Will merge if there are no errors.


esiragusa:
:+1:

PR #997, opened 2015-05-13
Feature/better stl adapt

h-2:
This is a preliminary pull for the new stl adaptation. 

It has support for all non-associative stl containers (also deque, forward_list, array). It has support for move semantics. It should be a lot more consistent and clear, although it obviously did not get shorter than the old one.

Open issues:
  * some warning on gcc if no cxx11 is available
  * the v2 tests for std::deque crash on clang 
These will be fixed soon.

Many things in the sequence tests had to be changed, no more "=" and no more "==", but only assign() now. This also means that tests for const strings are currently disabled. There are currently no tests for std::array because it is doesn't fulfill our SequenceConcept.


h-2:
This should also fix some outstanding issues for the old code. I will make a pull for develop, once I have your feedback on this.

h-2:
beside some weird error in ctd support, this is done from my POV.

include/seqan/sequence/string_packed.h
```diff
@@ -678,8 +678,8 @@ getObjectId(String<TValue, Packed<THostspec> > const & me)
 template <typename TValue, typename THostspec, typename TPos>
 inline typename Iterator<String<TValue, Packed<THostspec> >, Standard>::Type
 iter(String<TValue, Packed<THostspec> > & me,
-     TPos pos,
-     Standard)
+     TPos const pos,
+     Standard const &)
 {
```
h-2: correct whitespace

tests/basic/test_basic_allocator.cpp
```diff
@@ -25,7 +25,7 @@
 // SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 // CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
 // LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
-// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+    // OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 // DAMAGE.
 //
```
h-2: remove

h-2:
@rrahn I think I have changed all the stuff you recommended, can you have a look again?

include/seqan/sequence/string_packed.h
h-2: why did we comment on this?

include/seqan/sequence/adapt_array_pointer.h
```diff
@@ -102,54 +102,30 @@ struct DefaultOverflowExplicit< TValue const [SIZE] >
...
 
 // ----------------------------------------------------------------------------
 // Concept Sequence
 // ----------------------------------------------------------------------------
 
-template <typename TValue>
-SEQAN_CONCEPT_IMPL((TValue *), (ContainerConcept));
+// NOTE(h-2): why is this broken?
+// template <typename TValue>
+// SEQAN_CONCEPT_IMPL((TValue *), (ContainerConcept));
```
h-2: If the tests pass this is the last issue. As far as I am concerned I would just leave it like this, because making a regular pointer a container:
  * makes no sense
  * is not used anywhere
  * is not documented so no one expects it

h-2:
as discussed with @esiragusa I removed the container concept for simple pointers. If this doesnt break anything else I will merge it later today, however only to develop and not to master.

PR #998, opened 2015-05-15
[FIXED-#828] Adds assertion to writeRecord of Bam/Sam IO. 

rrahn:
PR for Master.

See PR for develop #964 for discussions.


rrahn:
Falsely based on develop.

PR #999, opened 2015-05-15
[FIXED-#828] Adds assertion to writeRecord of Bam/Sam IO. 

rrahn:
Given the nature of the FormattedFile it would be in general possible to create a Bam/Sam file with an empty IO context. By the Sam specifications this might be valid (contains only unmapped reads), so we need a way to assert an incomplete IO context, which can happen easily as demonstrated by the following example taken from demos/tutorial/bam_io/solution1.cpp:

```cpp
BamFileIn bamIn("example.bam")
// some error checking.
BamFileOut bamOut(cout, Sam());
// Reading bamFile and writing sam file to console (as demonstrated by the demo above.) 
```
The code is valid but yields an error, since the context (contig names) of the ```bamOut``` is not set. This leads to a misleading error when writing Sam and even worse no error when writing in Bam format. The error would first appear, when one tries to read in from the erroneous bam file. 

This fix simply adds an assertion when writing Sam/Bam file that checks if the ```rID``` and the ```rNextID``` points to a valid contig name according to the Sam specifications.
This way the user can detect the error more easily.


rrahn:
CI build errors not related.

PR #1003, opened 2015-06-02
[FIX] deactivate one test on freebsd

h-2:
this deactivates a test on FreeBSD that otherwise fails because of lower math precision in the FreeBSD c standard library.

Will merge later today, unless someone objects. 

@skrakau is this okay for you? it shouldnt effect anything else.


h-2:
failure unrelated

Issue #644, opened 2014-10-17
combining infixes and modified strings doesn't work right


h-2:
Possibly all combinations of "virtual" strings, that cannot be compressed to a single one are broken (e.g. two infixes are compressed to one).

the problem is with the 'value' function and 'getValue' functions (among other things). When accessing values on a 'infix(modString(somestring))' multiple iterators are created on the way of whome an inner one leaves the stack while outr ones still depend on it, resulting in very erratic behaviour.

b80bd4cbb04693def9d085bd1b805ac1202e9daf fixes this behaviour for 'getValue(infix(modstring)))'  (only this combination), by directly creating the single right iterator and not multiple ones. But this needs a general solution.

h-2:
I will fix this after the blast module is merged and I adapted Lambda to the new codebase (since Lambda triggers this issue).

h-2:
Should be fixed by 429b9b4 / aa3ecea

Issue #538, opened 2014-10-04
Cannot iterator over const ModifiedString


holtgrewe:
Originally from John on Trac

I'm getting a compilation error when I try to iterate over a const ModifiedString. I can iterator over a mutable String, a const String and a mutable ModifiedString. I'm wondering if there is a reason why I'm getting this error. Here is some test code:

```
#include <seqan/sequence.h>
#include <seqan/modifier.h>

/// Defines a type to represent the reverse complement of the template string type.
template< typename StringT >
struct rev_comp {

	typedef typename seqan::Value< StringT >::Type   string_char; ///< The type of a character in the string.

	/// The type of the reverse complement of the string.
	typedef seqan::ModifiedString<
		seqan::ModifiedString<
			StringT,
			seqan::ModView< seqan::FunctorComplement< string_char > >
		>,
		seqan::ModReverse
	> type;
};

int
main( int argc, char * argv[] ) {
	using namespace seqan;

	typedef String< Dna5 >                                string_t;
	typedef rev_comp< string_t >::type                    rev_comp_t;

	string_t                                        str( "ACGTNACGTNACGTN" );
	rev_comp_t                                      rev_comp( str );
	begin( str );
	begin( rev_comp );

	const string_t                                  const_str( "ACGTNACGTNACGTN" );
	const rev_comp_t                                const_rev_comp( str );
	begin( const_str );
	begin( const_rev_comp ); // <-- CAUSES COMPILATION ERROR
}
```

I'm using seqan svn revision 5028 with gcc 4.4.1 on Ubuntu.

h-2:
This now builds fine.

PR #1007, opened 2015-06-03
Fix/modified iterators

h-2:
This is a fix for #644 and hopefully also for #538. For develop it is included in #997, because it was triggered by changes there.  For master it comes by way of this pull request.

Previous tests showed that this effects no apps or tests other than rabema which is fixed by one of the diffs below. The corresponding part in rabema is covered by app tests (which pass), but maybe @holtgrewe wants to double-check it?


holtgrewe:
As long as the RABEMA tests run through I'm fine.

h-2:
@holtgrewe yes they pass.

Issue #1008, opened 2015-06-03
BAM IO test failure


h-2:
One of the BAM IO tests fails with:
`Uncaught exception of type std::ios_base::failure: BGZF block too short.` (gcc)
and 
`Terminating due to uncaught exception 0x805c10020 of type std::__1::ios_base::failure
Attempting to rethrow an exception that doesn't exist!
Attempting to rethrow an exception that doesn't exist!
...` (clang)

This happes **always** on FreeBSD and randomly (not often) on Linux (especially on Travis). See current cdash for more details.

esiragusa:
Fixed by #1053 #1054.

rrahn:
This is not completely

rrahn:
I postpone this to when we have actual freeBSD support.

h-2:
we do have actual freebsd support, this is the only remaining os-specific problem and I want to update the offical FreeBSD port to 2.x (http://www.freshports.org/biology/seqan).

h-2:
@rrahn if possible change the file in the nightlies before 2.1.1 and move fixing the real problem to later. Thanks!

h-2:
fixed via #1791

PR #1011, opened 2015-06-05
[FIX] fixed bgzfstream tell to return the begin offset ... [master]

weese:
... of the next block instead of the end of the previous one, as the samtools do.

### DIFF
--- a/include/seqan/stream/zipstream/bgzfstream.h
+++ b/include/seqan/stream/zipstream/bgzfstream.h
@@ -345,6 +345,7 @@ public:
         TBuffer         buffer;
         off_type        fileOfs;
         int             size;
+        unsigned        compressedSize;
 
         CriticalSection cs;
         Condition       readyEvent;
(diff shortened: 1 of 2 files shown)


h-2:
Hi Dave, do you think this will fix our (more or less random) bam_io test failures?

weese:
I don't know. Didn't know that they failed.

weese:
So, no.

weese:
And I have to fix a test first, before it can be merged.

weese:
The micro_razers error seems to be unrelated and gcc app test errors as well, as they don't appear in the Travis report of my personal branch used for the PR.

My branch succeeds with gcc https://travis-ci.org/weese/seqan/builds/65554170

h-2:
Ok, LGTM.

PS: I was referring to #1008 up there, but thats probably unrelated...

h-2:
If no one objects I will merge it in half an hour.

PR #1027, opened 2015-06-22
[FIX] assigning host to modview

h-2:
make assigning a host to a modview work.

If there are no objections, I will merge tomorrow morning.

should this go into master as well?


include/seqan/modifier/modifier_view.h
```diff
@@ -475,6 +475,24 @@ convert(TSequence const & sequence, TFunctor const &F)
...
+// --------------------------------------------------------------------------
+// Function assign()
+// --------------------------------------------------------------------------
+
+template <typename THost, typename TFunctor>
+void assign(ModifiedString<THost const, ModView<TFunctor> > & target,
+            THost const & host)
+{
+    target._host = &host;
+}
```
rrahn: This is the same signature as the non-const version below if THost is const. Probably this results in an ambiguous function call error? What was the intension of the const version anyway?

tests/modifier/test_modifier_view.h
```diff
@@ -399,4 +399,21 @@ SEQAN_DEFINE_TEST(test_modifier_convert_in_place)
...
+
+    CharString originalStr = "acgtnACGTN";
+    Dna5String EXPECTED_RESULT = "ACGTNACGTN";
+
+    ModifiedString<CharString, ModView<TFunctor> > modifiedStr;
+    assign(modifiedStr, originalStr);
+
+    CharString modifiedStrCopy = modifiedStr;
+    SEQAN_ASSERT_EQ(modifiedStrCopy, EXPECTED_RESULT);
+}
```
rrahn: Please test the const THost version.

esiragusa:
* The function that you named assign() has to be called setHost()
* You have to use _toPointer() - see https://github.com/seqan/seqan/blob/master/include/seqan/modifier/modifier_view.h#L185.

h-2:
> Please test the const THost version.

Yep.

> You have to use _toPointer() - see https://github.com/seqan/seqan/blob/master/include/seqan/modifier/modifier_view.h#L185.

OK!

> The function that you named assign() has to be called setHost()

No, I need assign, because the constructors call assign and you want the following to work
```c++
String<AminoAcid> foo = "ARGLILIL";
TModview bar(foo);
```
Actually it is even more complicated for Concat-Stringset, but having `assign()` do the above is very much expected I think.

esiragusa:
You are assigning the host, not the values.

As far as I understand, after ```assign(a,b)``` it must hold ```a == b```, but with modifiers it would be ```a != b```.

So you need setHost(), not assign(). The ModifiedString constructor will call setHost() if you provide an object of the host type.

h-2:
> As far as I understand, after assign(a,b) it must hold a == b, but with modifiers it would be a != b.

IAMNM a==b will hold because the non-modified view will be cast to a modvie, just the same as:

```c++
CharString a = "ACGTX";
DnaString b;
assign(a, b);
// a == b ?
```

> So you need setHost(), not assign(). The ModifiedString constructor will call setHost() if you provide an object of the host type.

But if I do the same on a ConcatDirect StringSet over a ModString, it will propagate to the Constructor of the StringSet which calls `assign()` on the members....

esiragusa:
The cast to ModifiedString should be forbidden - Is it working? :bomb: 

In your case, you want assign() to set the host, as ModifiedString is not a container i.e. it does not hold the values itself. Could you please have a look at how Segments behave?

h-2:
#1028 solves the problem for me, too, and is maybe less controversial ;)

If you approve of that I would close this one.

PR #1028, opened 2015-06-23
[INTERNAL] add concat-specific copy constructor

h-2:
Make copy construction go through down-stream copy-constructors instead of `assign()`.


h-2:
any comments on this? Otherwise I'd like to merge it and close #1027.

PR #1029, opened 2015-06-23
Fix/mod iterator position

h-2:
The existing functions did not work in certain situations, e.g.
  *  `position(host(me), cont);`
  *  `host(me)` will resolve to an iterator on the original string and `position()` will try to subtract `begin(modstring)` from it which will fail, because it is an iterator on the modstring.
  * this solution does the operations directly on the mod-string
 * doesn't effect things like modreverse which implement their own versions of these functions anyway.


h-2:
Hold this for now, there seems to be more stuff broken since seqan2.0.

h-2:
superseded by #1044

PR #1033, opened 2015-06-25
Fix/iupac alphabet

h-2:
For some undocumented reason that defies all other resources on the topic our IUPAC alphabet did not contain the letter 'U' but instead the letter '=' which is described as `// = or U`. I changed this now to actually be 'U'  so that sequence IO works. Implicit conversion from the `=` character is still included.


h-2:
@rrahn @esiragusa  can one of your peer review this?

Issue #1025, opened 2015-06-19
Only support clang with libc++


h-2:
Right now our settings for c++-standard-libraries are:

                | Linux         | FreeBSD  | Mac
------------ | ---------------|-----------| ---------
**g++**        | libstdc++  | libstdc++ |  libstdc++
**clang++** | libstdc++ |  libc++ |  libstdc++

I propose to change this to:

                | Linux         | FreeBSD | Mac
------------ | ---------------|-----------| ---------
**g++**        | libstdc++  | libstdc++ |  libstdc++
**clang++** | libc++       |  libc++ |  libc++

That is, we should always use libc++ with clang. The reason is that there are implementation problems in c++11 when mixing compilers and stdlibs and it gets ridiculously hard to track in which compiler/stdlib combination a feature works or not. Forcing libstdc++ on Mac supposedly means no c++11 on Mac, at all, so we should definitely give that up. [1]

The only real change would be for Linux where people have to install libc++. However, since clang is not installed by default on Linux anyway, we might as well let people install libc++.

What do you think?

[1] http://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-use-the-flag-stdlib-libstdc

esiragusa:
I see two potential disadvantages:
* On Linux people have to install libc++
* Is GNU parallel STL supported by libc++?

Which are the advantages? As far as I can see, on OSX libstdc++ seems compatible with c++11...

h-2:
> On Linux people have to install libc++

Yes, but on Linux people also have to install clang ;) No, seriously, on Linux it is a very conscious decision to use Clang instead of GCC, so I think it is not a problem to install libc++ in these cases, as well.

> Is GNU parallel STL supported by libc++?

AFAICT gnu parallel is not supported by Clang with libstdc++ either.

> Which are the advantages?

As I said above. Certain things are broken for Clang+libstdc++ , e.g. http://cdash.seqan.de/viewBuildError.php?buildid=133243
(I already had to work around lots of other issues before) Also look at 
https://github.com/seqan/seqan/blob/develop/include/seqan/platform/platform_gcc.h#L177
Right now I have made this library detection code based on the existance and later removal of a header from the code. Very ugly.

> As far as I can see, on OSX libstdc++ seems compatible with c++11...

Maybe you have installed a new g++ version yourself which comes with new libstdc++? Anyway, I can't say for sure for Mac. But, please note, that I just want libc++ for Clang! For GCC, libstdc++ will stay default.

[...]

PR #1036, opened 2015-07-03
[FIXED-#1025] use libc++ with clang always

h-2:
see #1025 for discussion. @rrahn and I think we should merge this into develop if it doesn't break anything.


h-2:
since we will drop support for gcc < 4.9 on seqan 2.1.0 this won'T be necessary.

Issue #1037, opened 2015-07-06
VirtualStream adds wrong file extensions like .fasta.bam


h-2:
I am confused about our file extensions. getFileExtensions actually generates `.fasta.bam` for me, just because there is bz2 on the system. Obviously I do want `.fasta.bz2`, but I don't want `.fasta.bam`.
`.bam` should be part of the BAM definition -- which it also is. Why is this the case and how can we change it?

Also, why are the other fasta extensions deactivated? They are quite common.

h-2:
Also `.z` is the extension for compress (LZW) and not for Gzip, and `.bz` ist the first bzip version which is patented and not compatible with bzip2. IMHO the only valid gzip extension should be `.gz` and the only valid bzip2 extension should be `.bz2`. If for some reason we want to read tarballs we can also add `.tgz` and `.tbz`, but I don't see the necessity right now.

h-2:
If no one objects until tomorrow noon, I will create a pull that 
  * removes extensions from gzip and bzip2 that are not used normally; and
  * adds the fasta extensions back again

esiragusa:
:+1: for removing extensions:
* .z https://github.com/seqan/seqan/blob/aedd69b4507649ef17fa8381442efe31d2d17c6f/include/seqan/stream/stream_base.h#L131
* .bz https://github.com/seqan/seqan/blob/aedd69b4507649ef17fa8381442efe31d2d17c6f/include/seqan/stream/stream_base.h#L161

To remove fasta.bam and fasta.bzgf you should remove BgzfFile from the list CompressedFileTypes used by SeqFile's VirtualStream. See:
* CompressedFileTypes https://github.com/seqan/seqan/blob/aedd69b4507649ef17fa8381442efe31d2d17c6f/include/seqan/stream/virtual_stream.h#L71
* FileFormat https://github.com/seqan/seqan/blob/aedd69b4507649ef17fa8381442efe31d2d17c6f/include/seqan/stream/virtual_stream.h#L318

What do you mean by "adds the fasta extensions back again"?

h-2:
@esiragusa Please review the pull request once tests pass, thanks!

> What do you mean by "adds the fasta extensions back again"?
see b5c27b00d636517789b9428402f920f9cbf4f7b8

h-2:
@temehi  we can discuss this together..

PR #1045, opened 2015-07-23
fix extensions (master)

h-2:
see #1037

### DIFF
--- a/include/seqan/bam_io/bam_file.h
+++ b/include/seqan/bam_io/bam_file.h
@@ -359,6 +359,24 @@ writeRecords(FormattedFile<Bam, Output, TSpec> & file, TRecords const & records)
         write(file.iter, buffers[i]);
 }
 
+// ----------------------------------------------------------------------------
+// Function getFileExtensions()
+// ----------------------------------------------------------------------------
+// NOTE(h-2): this is overloaded so we get Bgzf in addition to other
+// compressions which is crucial for Bam
+
+template <typename TDirection, typename TSpec>
+static std::vector<std::string>
+getFileExtensions(FormattedFile<Bam, TDirection, TSpec> const & file)
+{
+    std::vector<std::string> extensions;
+    _getCompressionExtensions(extensions,
+                              typename FormattedFile<Bam, TDirection, TSpec>::TFileFormats(),
+                              CompressedFileTypesWithBgzf_(),
+                              false);
+    return extensions;
+}
+
 }  // namespace seqan
 
 #endif // SEQAN_BAM_IO_BAM_FILE_H_
--- a/include/seqan/bam_io/bam_file.h
+++ b/include/seqan/bam_io/bam_file.h
@@ -372,7 +372,7 @@ getFileExtensions(FormattedFile<Bam, TDirection, TSpec> const & file)
     std::vector<std::string> extensions;
     _getCompressionExtensions(extensions,
                               typename FormattedFile<Bam, TDirection, TSpec>::TFileFormats(),
-                              CompressedFileTypesWithBgzf_(),
+                              CompressedFileTypes(),
                               false);
     return extensions;
 }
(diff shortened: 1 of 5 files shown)


h-2:
@esiragusa Now some of the tests involving bam fail. I think the transparent decompression is not applied... do you have any quick thoughts on this?

esiragusa:
Hey, in #1037 I meant to overload the FileFormat metafunction for the VirtualStream of SeqFile and remove BgzfFile from the list. In this way you disabled it for all file formats!

esiragusa:
And, in my opinion, the name FileFormat is misleading! FileFormat of FormattedFile lists file formats (it's okay) but FileFormat of VirtualStreams lists stream encodings (maybe it should be called StreamFormat or similar).

h-2:
> Hey, in #1037 I meant to overload the FileFormat metafunction for the VirtualStream of SeqFile and remove BgzfFile from the list. In this way you disabled it for all file formats!

Hm, shouldn't it be the other way around, that bzgf is especially enabled for some, but not in general?

esiragusa:
I don't think so, bzgf is a general compression format (like bzip2).

There seems to be a bit of confusion with formats/extensions: we don't want to have fastq.bam but we might want to have fastq.bzgf (e.g. for parallel decompression or random seeks).

h-2:
> I don't think so, bzgf is a general compression format (like bzip2).

I can find no documentation on this. Google barely yields any results and those that it does all point to BAM. 

But I can  overload FileFormat for SeqFile as suggested early on and maybe we can rediscuss this a later point.

esiragusa:
Google doesn't help because I mistyped the name, it's BGZF! :tada:

h-2:
ROFL, okay!

h-2:
> And, in my opinion, the name FileFormat is misleading! FileFormat of FormattedFile lists file formats (it's okay) but FileFormat of VirtualStreams lists stream encodings (maybe it should be called StreamFormat or similar).

It all seems kind of confusing to me. I don't see how I can overload the compressed formats for any FormattedFile, since VirtualStream cannot be specialized for this.

We have different FormattedFiles which each have different Formats which each have different extensions and each extension has again multiple compressed versions. And the mechanisms for defining what is legal in the end are different for each layer of abstraction. Plus there is (nearly) no documentation -> I don't want to complain, but its really not nice.

Any hints for solving the current issue, i.e. not having the bgzf extensions added to SeqFormats?

Thanks!

h-2:
And why is `getFileExtensions()` a member function of `FormattedFile`? It means that I can't overload it, otherwise I could sneak in a different CompressedFormats Taglist there...

I could overload `_getCompressionExtensions()` for SeqFormat and always ignore/replace the copmressionformats argument, but thats a hack at best.

h-2:
Ok, I have an intermediate solution. First of all, I would like to say that I think Bgzf should just not define `.bam` as of its extensions. This should only every be defined by the bam-format. However it is apperently not easy to do with our IO-code, at least not for me.

The current solution does the following:
  * define one set of CompressedFileTypes with and one without Bgzf
  * the `getFileExtensions()` by default returns combinations without, but for `FormattedFile<Bam...` returns combinations with

This achieves the desired behaviour with minimal changes and is easy to adapt, once a more thorough solution should be developed.

@rrahn can you review?

rrahn:
First of all, why should fasta.bgzf be a problem? Look here: http://blastedbio.blogspot.de/2011/11/bgzf-blocked-bigger-better-gzip.html
So biopython can do this on any type of file :v:
So in general as @esiragusa already mentioned, this is just another compression format.
So why all the hustle to get it off of other sequence file formats?

rrahn:
So is the issue, that we don't have a compressed format called ```bgzf``` but only ```bam```?

h-2:
> So is the issue, that we don't have a compressed format called bgzf but only bam?

The issue is that `.bam` is accepted as general extension for any bgzf'ed file. As discussed I will merge this and we can reevaluate the issue at a later point in #1037

PR #1044, opened 2015-07-22
modified iterators and modview (again)

h-2:
some modview related fixes and new tests.

Don't commit yet, some more cleanup needed.

PR #1047, opened 2015-07-24
modified iterators and modview (again) [develop]

h-2:
see #1044

### DIFF
--- a/include/seqan/basic/iterator_interface.h
+++ b/include/seqan/basic/iterator_interface.h
@@ -350,7 +350,7 @@ position(TIterator const & it,
 
 template <typename T, typename TContainer>
 inline SEQAN_HOST_DEVICE bool
-atBegin(T const & it, TContainer const & cont)
+atBegin(T & it, TContainer & cont)
 {
     SEQAN_CHECKPOINT;
     return it == begin(cont, Standard());
@@ -369,8 +369,8 @@ atBegin(T const & it)
 
 template <typename T, typename TContainer>
 inline SEQAN_HOST_DEVICE bool
-atEnd(T const & it,
-      TContainer const & cont)
+atEnd(T & it,
+      TContainer & cont)
 {
     SEQAN_CHECKPOINT;
     return it == end(cont, Standard());
--- a/include/seqan/basic/iterator_interface.h
+++ b/include/seqan/basic/iterator_interface.h
@@ -356,37 +356,6 @@ atBegin(T const & it, TContainer const & cont)
     return it == begin(cont, Standard());
 }
 
-template <typename T, typename TContainer>
-inline SEQAN_HOST_DEVICE bool
-atBegin(T const & it, TContainer & cont)
-{
-    SEQAN_CHECKPOINT;
-    return it == begin(cont, Standard());
-}
-
-template <typename T, typename TContainer>
-inline SEQAN_HOST_DEVICE bool
-atBegin(T & it, TContainer & cont)
-{
-    SEQAN_CHECKPOINT;
-    return it == begin(cont, Standard());
-}
-
-template <typename T, typename TContainer>
-inline SEQAN_HOST_DEVICE bool
-atBegin(T & it, TContainer const & cont)
-{
-    SEQAN_CHECKPOINT;
-    return it == begin(cont, Standard());
-}
-
-template <typename T>
-inline SEQAN_HOST_DEVICE bool
-atBegin(T & it)
-{
-    return atBegin(it, container(it));
-}
-
 template <typename T>
 inline SEQAN_HOST_DEVICE bool
 atBegin(T const & it)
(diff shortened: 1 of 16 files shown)


include/seqan/basic/iterator_interface.h
```diff
@@ -350,43 +350,12 @@ position(TIterator const & it,
...
-inline SEQAN_HOST_DEVICE bool
-atBegin(T const & it, TContainer & cont)
-{
-    SEQAN_CHECKPOINT;
-    return it == begin(cont, Standard());
-}
-
-template <typename T, typename TContainer>
-inline SEQAN_HOST_DEVICE bool
 atBegin(T & it, TContainer & cont)
```
rrahn: Why keeping the non-const version instead of the const-version?
atBegin does neither change the iterator nor the container and returns a bool. Hence I would rather keep the very first one?

include/seqan/basic/iterator_interface.h
```diff
@@ -401,47 +370,12 @@ atBegin(T const & it)
...
-atEnd(T & it,
-      TContainer & cont)
-{
-    SEQAN_CHECKPOINT;
-    return it == end(cont, Standard());
-}
-
-template <typename T, typename TContainer>
-inline SEQAN_HOST_DEVICE bool
-atEnd(T const & it,
```
rrahn: Same here!?

include/seqan/modifier/modifier_iterator.h
```diff
@@ -130,14 +130,8 @@ struct Value<ModifiedIterator<THost, TSpec> const> : Value<ModifiedIterator<THos
...
 
-//NOTE(h-2): ModifiedStringIterators always return by value since some
-// modified strings result in values being created anyway (e.g. ModViews)
-// and depending on scope these might be lost.
-// For example iterators over infixes of ModViews would otherwise reference
-// part of the stack that were freed.
-
 template <typename THost, typename TSpec>
-struct GetValue< ModifiedIterator<THost, TSpec> > : Value<THost>
+struct GetValue< ModifiedIterator<THost, TSpec> > : GetValue<THost>
```
rrahn: Is this assumption you noted and deleted above not valid anymore?

include/seqan/modifier/modifier_iterator.h
```diff
@@ -538,11 +530,20 @@ position(ModifiedIterator<THost, TSpec> const & me)
     return position(host(me));
 }
 
+// rooted overload
+template <typename TContainer1, typename TIterator, typename TSpec1, typename TSpec2, typename TContainer2>
+inline typename Position<ModifiedIterator<Iter<TContainer1, AdaptorIterator<TIterator, TSpec1> >, TSpec2> const>::Type
+position(ModifiedIterator<Iter<TContainer1, AdaptorIterator<TIterator, TSpec1> >, TSpec2> const & me,
+        TContainer2 const &)
```
rrahn: Please indent by one.

rrahn: Is there no global rooted iterator interface that could be called instead of overloading it?

include/seqan/modifier/modifier_iterator.h
```diff
@@ -538,11 +530,20 @@ position(ModifiedIterator<THost, TSpec> const & me)
...
+{
+    return position(me); // rooted has container
+}
+
 template <typename THost, typename TSpec, typename TContainer>
 inline typename Position<ModifiedIterator<THost, TSpec> const>::Type
 position(ModifiedIterator<THost, TSpec> const & me, TContainer const &cont)
 {
-    return position(host(me), cont);
+    return me - begin(cont, Standard());
```
rrahn: Again, this seems very generic to me? Is there no default function that would be called instead. We always have something like that.

include/seqan/modifier/modifier_iterator.h
```diff
@@ -593,68 +594,56 @@ operator>(ModifiedIterator<THost, TSpec> const & a, ModifiedIterator<THost, TSpe
...
 
-template <typename THost, typename TSpec>
+template <typename THost, typename TSpec, typename TContainer>
 inline bool
-atEnd(ModifiedIterator<THost, TSpec> const & me)
+atEnd(ModifiedIterator<THost, TSpec> & me,
+      TContainer & container)
 {
-    return atEnd(host(me));
+    return me == end(container, Standard());
```
rrahn: Why needing this overload?

include/seqan/basic/iterator_interface.h
h-2: The non-const Version implies both a cinst interface and a non-const interface.

Having only a const interface on the other hand results in comparisons between const iterators and non const iterators (if we had an explicit `cbegin()` this could be used). Doing a cast might also work but makes the code less readable.

include/seqan/modifier/modifier_iterator.h
h-2: No, I have moved the special logic to modview so that it doesn't effect all mod strings, but only the modviews.

include/seqan/modifier/modifier_iterator.h
h-2: Unfortunately not, and our rooted iterator doesn't even fulfill our rooted iterator concept :S
Even if it this were fixed it is not possiblr to do globally, because the modified iterator of a rooted iterator is a specialization of the modified iterator and not a specialization of the rooted iterator.

include/seqan/modifier/modifier_iterator.h
h-2: Sure there is a generic function, but it doesn't work ;) 
... Because the host iterator might be on a completely different alphabet then the container (and the mod iterator).

include/seqan/modifier/modifier_iterator.h
h-2: See above. The extra rioted overload is needed, because we cannot subtract rooted and standard iterators.

include/seqan/basic/iterator_interface.h
rrahn: Well, this could happen with the non-const interface as well. 
Secondly, this shouldn't cause any problems as the comparison of two iterators should be independent of there const-ness, as it is checked if they point to the same element requiring them to be in the same domain. Since the comparison does not change the state it should be const. 
I mean what is the benefit of the non-const version?
Moreover, something like this could not work:
```cpp
if (atEnd(std::find(begin(cont), end(cont), f), cont))
{
    // do stuff
}
```
Since the rvalue cannot bind to the lvalue reference, but to a lvalue const reference.
I know the example is not very intuitive, but since iterators are considered lightweight objects they might as well be returned by value in some functions. So the const version does not harm and is safer in usage.
The same for the other case. Assume, that the container of the iterator is lightweight and I call a function to get the container by value, the non-const version wouldn't work.
Or am I overlooking something here?
As far as I know, the comparison is implemented on const references anyway. So why shouldn't that work?

PR #1052, opened 2015-07-29
fix extensions [develop]

h-2:


esiragusa:
IMHO this is not a hotfix and should have got into develop only...

PR #1057, opened 2015-07-30
modstrings again... (master)

h-2:
followup to #1044


h-2:
hold for now :(

h-2:
Now includes a fix for:
http://cdash.seqan.de/viewBuildError.php?buildid=137652
And backports another commit that was previously only in develop (but now is needed in master, too).

PR #1062, opened 2015-08-03
Trie construction

esiragusa:
Made trie and radix tree construction part of the public indexCreate() API.


esiragusa:
Failing tests are unrelated.

PR #1070, opened 2015-08-05
Feature/set host

rrahn:
Adds setHost with move semantics to replace old non-const and const version which was misused in many places, especially in the Pattern implementation If 
If c++11 is enabled! The setHost of Pattern delegates now correctly to a Holder, meaning that temporaries are copied and lvalue refs are set dependent.

### DIFF
--- a/include/seqan/align/dp_matrix.h
+++ b/include/seqan/align/dp_matrix.h
@@ -122,21 +122,6 @@ public:
     {
         create(data_host);
     }
-//
-//    DPMatrix_(DPMatrix_ const & other) :
-//        data_host(other.data_host) {}
-//
-//    ~DPMatrix_() {}
-//
-//    DPMatrix_ & operator=(DPMatrix_ const & other)
-//    {
-//        if (this != &other)
-//        {
-//            data_host = other.data_host;
-//        }
-//        return *this;
-//    }
-
 };
 
 // ============================================================================
--- a/include/seqan/align/dp_matrix.h
+++ b/include/seqan/align/dp_matrix.h
@@ -60,6 +60,13 @@ struct DefaultScoreMatrixSpec_;
 // ============================================================================
 
 // ----------------------------------------------------------------------------
+// Tag MatrixMember
+// ----------------------------------------------------------------------------
+
+struct DPMatrixMember_;
+typedef Tag<DPMatrixMember_> DPMatrixMember;
+
+// ----------------------------------------------------------------------------
 // Tag SparseDPMatrix
 // ----------------------------------------------------------------------------
 
@@ -106,29 +113,29 @@ class DPMatrix_<TValue, FullDPMatrix>
 {
 public:
 
-    typedef Matrix<TValue, 2> THost;
+    typedef typename Member<DPMatrix_, DPMatrixMember>::Type THost;
 
-    Holder<THost>   _dataHost;  // The host containing the actual matrix.
+    Holder<THost>   data_host;  // The host containing the actual matrix.
 
     DPMatrix_() :
-        _dataHost()
+        data_host()
     {
-        create(_dataHost);
+        create(data_host);
     }
-
-    DPMatrix_(DPMatrix_ const & other) :
-        _dataHost(other._dataHost) {}
-
-    ~DPMatrix_() {}
-
-    DPMatrix_ & operator=(DPMatrix_ const & other)
-    {
-        if (this != &other)
-        {
-            _dataHost = other._dataHost;
-        }
-        return *this;
-    }
+//
+//    DPMatrix_(DPMatrix_ const & other) :
+//        data_host(other.data_host) {}
+//
+//    ~DPMatrix_() {}
+//
+//    DPMatrix_ & operator=(DPMatrix_ const & other)
+//    {
+//        if (this != &other)
+//        {
+//            data_host = other.data_host;
+//        }
+//        return *this;
+//    }
 
 };
 
@@ -160,21 +167,16 @@ struct DefaultScoreMatrixSpec_<LocalAlignment_<WatermanEggert> >
 // ----------------------------------------------------------------------------
 
 // Returns the type of the underlying matrix.
-template <typename TDPMatrix>
-struct DataHost_ {};
-
 template <typename TValue, typename TMatrixSpec>
-struct DataHost_<DPMatrix_<TValue, TMatrixSpec> >
+struct Member<DPMatrix_<TValue, TMatrixSpec>, DPMatrixMember>
 {
-    typedef DPMatrix_<TValue, TMatrixSpec> TDPMatrix_;
-    typedef typename TDPMatrix_::THost Type;
+    typedef Matrix<TValue, 2> Type;
 };
 
 template <typename TValue, typename TMatrixSpec>
-struct DataHost_<DPMatrix_<TValue, TMatrixSpec> const>
+struct Member<DPMatrix_<TValue, TMatrixSpec> const, DPMatrixMember>
 {
-    typedef DPMatrix_<TValue, TMatrixSpec> TDPMatrix_;
-    typedef typename TDPMatrix_::THost const Type;
+    typedef Matrix<TValue, 2> const Type;
 };
 
 // ----------------------------------------------------------------------------
@@ -190,7 +192,7 @@ template <typename TValue, typename TMatrixSpec>
 struct SizeArr_<DPMatrix_<TValue, TMatrixSpec> >
 {
     typedef DPMatrix_<TValue, TMatrixSpec> TDPMatrix_;
-    typedef typename DataHost_<TDPMatrix_>::Type TDataHost_;
+    typedef typename Member<TDPMatrix_, DPMatrixMember>::Type TDataHost_;
     typedef typename SizeArr_<TDataHost_>::Type Type;
 };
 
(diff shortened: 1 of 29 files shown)


include/seqan/find/find_hamming_simple.h
```diff
@@ -78,11 +78,26 @@ class Pattern<TNeedle, HammingSimple> {
...
+#ifdef SEQAN_CXX11_STANDARD
+    template <typename TNeedle2>
+    Pattern(TNeedle2 && ndl,
+            int k = -1,
+            SEQAN_CTOR_DISABLE_IF(IsSameType<typename std::remove_reference<TNeedle2>::type const &, Pattern const &>))
+                :   maxDistance(-1),
+                    distance(0),
+                    matchNFlags(0)
+    {
+        setHost(*this, std::forward<TNeedle2>(ndl), k);
```
rrahn: Actually setting the distance should not be done using the setHost interface.
I think there are interfaces to set the distance. We rather keep a clean setHost interface and set another feature with another function, especially as the current interface does not require the distance parameter for the initialisation of the Pattern.

include/seqan/find/find_multiple_bfam.h
```diff
@@ -110,16 +110,29 @@ class Pattern<TNeedle, MultiBfam<TSpec> >
...
     {
         SEQAN_CHECKPOINT
         setHost(*this, ndl);
     }
+# endif
 
     ~Pattern()
-    {
-    }
+    {}
```
rrahn: Remove default destructor.

include/seqan/find/find_myers_ukkonen.h
```diff
@@ -615,109 +674,100 @@ _patternMatchNOfFinder(Pattern<TNeedle, Myers<TSpec, THasState, void> > & patter
...
+//_myersSetHost(Pattern<TNeedle, Myers<TSpec, THasState, TFindBeginPatternSpec> > &, TNeedle2 const &)
+//{
+//}
+//
+//template <typename TNeedle, typename TSpec, typename TFinderCSP, typename TPatternCSP, typename THasState, typename TFindBeginPatternSpec, typename TNeedle2>
+//inline void
+//_myersSetHost(Pattern<TNeedle, Myers<AlignTextBanded<TSpec, TFinderCSP, TPatternCSP>, THasState, TFindBeginPatternSpec> > & pattern, TNeedle2 const & ndl)
+//{
+//    setValue(pattern.data_host, ndl);
+//}
```
rrahn: remove commented block

include/seqan/align/dp_matrix.h
```diff
@@ -106,29 +113,29 @@ class DPMatrix_<TValue, FullDPMatrix>
...
+//    ~DPMatrix_() {}
+//
+//    DPMatrix_ & operator=(DPMatrix_ const & other)
+//    {
+//        if (this != &other)
+//        {
+//            data_host = other.data_host;
+//        }
+//        return *this;
+//    }
```
rrahn: Remove commented block.

include/seqan/find/find_myers_ukkonen.h
```diff
@@ -615,109 +674,100 @@ _patternMatchNOfFinder(Pattern<TNeedle, Myers<TSpec, THasState, void> > & patter
...
+//
+//template <typename TNeedle, typename TSpec, typename THasState, typename TFindBeginPatternSpec>
+//inline TNeedle
+//host(Pattern<TNeedle, Myers<TSpec, THasState, TFindBeginPatternSpec> > & pattern)
+//{
+//SEQAN_CHECKPOINT
+//    typedef Pattern<TNeedle, Myers<TSpec, THasState, TFindBeginPatternSpec> > TPattern;
+//    return host(const_cast<TPattern const &>(pattern));
+//}
```
rrahn: remove.

include/seqan/find/find_pattern_base.h
```diff
@@ -221,36 +228,99 @@ struct ScoringScheme<TNeedle const>:
...
+        TNeedle2 && ndl)
+{
+    SEQAN_ASSERT(!empty(ndl));
+    setValue(_dataHost(me), std::forward<TNeedle2>(ndl));  // Set generic setHost function.
+    _reinitPattern(me);
+}
+
+template <typename TNeedle, typename TSpec, typename TNeedle2, typename TError>
+inline void
+setHost(Pattern<TNeedle, TSpec> & me,
```
rrahn: remove this interface as discussed above.

include/seqan/find/find_pattern_base.h
```diff
@@ -221,36 +228,99 @@ struct ScoringScheme<TNeedle const>:
...
 {
-     me.data_host = ndl; //assign => Pattern haelt eine Kopie => doof!
+    SEQAN_ASSERT(!empty(ndl));
+    setValue(_dataHost(me), ndl);  // Set generic setHost function.
+    _reinitPattern(me);
+}
+
+template <typename TNeedle, typename TSpec, typename TNeedle2, typename TError>
+inline void
+setHost(Pattern<TNeedle, TSpec> & me,
```
rrahn: remove this interfaces with the distance parameter.

include/seqan/find/find_shiftand.h
```diff
@@ -138,63 +150,64 @@ setHost(Pattern<TNeedle, ShiftAnd> & me, TNeedle2 const & needle)
...
+//    return host(pattern);
+//}
+//
+//template <typename TNeedle>
+//inline TNeedle
+//needle(Pattern<TNeedle, ShiftAnd> & pattern)
+//{
+//SEQAN_CHECKPOINT
+//    return host(const_cast<Pattern<TNeedle, ShiftAnd> const &>(pattern));
+//}
```
rrahn: remove commented block

include/seqan/find/find_shiftor.h
```diff
@@ -94,15 +105,17 @@ class Pattern<TNeedle, ShiftOr> {
 // Functions
 //////////////////////////////////////////////////////////////////////////////
 
-template <typename TNeedle, typename TNeedle2>
+// TODO(rrahn): Change to rvalue reference for TNeedle2 as soon as all pattern store pointer to the needle.
```
rrahn: remove comment.

include/seqan/find/find_shiftor.h
```diff
@@ -138,63 +151,63 @@ setHost(Pattern<TNeedle, ShiftOr> & me, TNeedle2 const & needle)
...
+//    return host(pattern);
+//}
+//
+//template <typename TNeedle>
+//inline TNeedle
+//needle(Pattern<TNeedle, ShiftOr> & pattern)
+//{
+//SEQAN_CHECKPOINT
+//    return host(const_cast<Pattern<TNeedle, ShiftOr> const &>(pattern));
+//}
```
rrahn: remove commented block.

include/seqan/index/find_backtracking.h
```diff
@@ -675,6 +680,7 @@ void _moveIteratorAtRoot(Pattern<Index<TNeedle, TSpec>, Backtracking<TDistance,
     me.index_iterator = TIndexIterator(host(me));
 }
 
+// TODO(rrahn): I am not sure about the index version. Should it always be copied? Seems very error prone to me!
```
rrahn: Adapt to also allow setting a reference instead of always copying.

esiragusa:
What is broken with non-const/const ref setHost()? What do you fix with rvalue ref setHost()?

rrahn:
When calling setHost, it always does a const-cast on the needle, causing the underlying holder always to take a copy of the needle instead of the storing a pointer to the resource (unless the Holder is over a const TNeedle type, which I haven't seen so far.) 
If this is the intended behaviour of the pattern class then it makes no sense to use a holder at all. With universal reference, it either takes a lvalue reference and propagates it as such or it can get a temporary and propagates as such, so that by passing a rvalue the holder copies the resource and it stores only a pointer to the lvalue reference.

esiragusa:
Thanks for clarifying! This is now fixed with C++11 only, right? Apart from Pattern, is this bug affecting any class using Holder + setHost()? To solve it properly, shouldn't we replace Holders with pointers?

rrahn:
Right, this only fixes this issue for c++11.
There where only a few other classes that uses a holder as well:
  * Matrix
  * DPMatrix
  * JournaledString
  * JournaledSet
  * BamTags something
But for them it is just a cosmetic change in order to avoid overloading just one instead of two interfaces or simply use the default setHost.
I rather tend to explicitly manage resources outside of the Pattern class and only use weak pointer or raw pointer to the resource. But I think this involves quite some changes. We planned to refactor find module anyway. I think we can then drop the Holder member?

esiragusa:
Ah okay, the problem arises only when one overloads setHost() and does a const cast, right?
For the find modules, it's fine like that, It has to be refactored!

rrahn:
Yes, only the Pattern implementation did the const_cast of the non-const needle for what ever reason?! :fearful:

esiragusa:
OK! Ship it if CI builds are fine :+1:

rrahn:
@esiragusa I also changed some of the pattern implementations, i.e. all patterns now support a Holder.
This wasn't the case for Myers<> unbanded, ShiftAnd and ShiftOr. They encoded the needle within the bit masks and did not hold a copy of the needle. Of course it wasn't necessary, because the needle was always copied. But I don't know where and how Myers is used and if it could affect the performance of some tools with that change. For the future if we completely move back from holders to raw pointers this might not a problem anymore. What do you think?

esiragusa:
Yep, in the new find module we will use pointers instead of holders. But it's not worth to fiddle with the old module...

rrahn:
ok, than I will role back the changes to these 3 instances to cover the old behaviour and we change it then with the new find module.

esiragusa:
Well if there are no issues with your changes you can keep them... :+1:

rrahn:
Well I think all tests run through! :feelsgood:

PR #1072, opened 2015-08-05
Support Visual Studio 2015

catkira:
everything except test_sequence_v2 and test_index_stree compiles


apps/razers2/outputFormat.h
```diff
@@ -785,7 +785,7 @@ void dumpMatches(
 
 		case 2:	// Eland Format
 			_sep_ = '\t';
-			for(unsigned readNo = 0; readNo < length(store.readSeqStore); ++readNo)
+			for (unsigned readNo = 0; it != itEnd; ++readNo, it++)
```
esiragusa: RazerS 2 app test now fails, probably because of this change. Anyway, this app is deprecated in favor of RazerS 3, so I would't spend any time on it :walking: We can disable VS2015 in its CMakeLists.txt

apps/stellar/stellar.cpp
```diff
@@ -581,6 +581,9 @@ void _setParser(ArgumentParser & parser)
 
 // TODO(holtgrew): Move this into a SeqAn misc module.
 
+// not supported anymore in vc2015
+// https://msdn.microsoft.com/en-us/library/bb531344.aspx
+#if _MSC_VER < 1900
```
esiragusa: Isn't sufficient to replace ```#ifdef PLATFORM_WINDOWS_VS``` below with ```#if _MSC_VER < 1900```?

apps/stellar/stellar.cpp
```diff
@@ -673,7 +676,9 @@ int mainWithOptions(TOptions & options, String<TAlphabet>)
 int main(int argc, const char * argv[])
 {
     // Makes sure that printing doubles in scientific notation is normalized on all platforms.
+#if _MSC_VER < 1900
```
esiragusa: See above and remove this.

include/seqan/basic/iterator_adapt_std.h
```diff
@@ -55,6 +55,17 @@ namespace std
         typedef typename seqan::Value<TIter>::Type * pointer; // nolint
         typedef typename seqan::Reference<TIter>::Type reference; // nolint
     };
+
+	// there is a bug in vc2015 stl, it doesnt check the iterator_traits correctly
+	// I have reported this bug to microsoft already (filed as DevDiv#1208117). 
+	// For now, this is a workaround.
+#if _MSC_VER == 1900
+	template<class _Ty, class Tag>
+	struct _Is_iterator<typename seqan::Iter<_Ty, Tag>>
```
esiragusa: Please use ```> >``` otherwise it's not valid C++03!

catkira: okay i will do next time, but this code is only active in vs2015 - i will change it anyway to make it more consistent :)

apps/razers2/outputFormat.h
catkira: okay, I will disable it for vs2015

apps/stellar/stellar.cpp
catkira: I dont know if i understand you correctly. But this _set_output_format function is supported in vs2013 and below, therefore I did not want to change tested code.

include/seqan/basic/allocator_interface.h
```diff
@@ -204,6 +204,7 @@ allocate(T const &,
     if (data)
         SEQAN_PROADD(SEQAN_PROMEMORY, count * sizeof(TValue));
 #endif
+	memset((void*)data, 0, count * sizeof(TValue));
```
esiragusa: Why do you need to zero the memory in allocate()?

BTW these SeqAn internals are :scream:

catkira: yes very scarry :)
test_sequence failed. I could narrow it down to seqan::string's internal memory not being initialized correctly (I compared the internal data after creation in vs2015 and gcc, it was different). After I added this memset, the unit test was ok for vs2015 and the internal data was the same between gcc and vs2015.
I did not understand all the internals completely, so probably only the designer can say if this memset should be neccessary or not. There is the possibility that a compiler bug in vs2015 makes this memset neccessary. But I think its more likely, that gcc and clang set the allocated memory to zero - even if they dont have to.

apps/stellar/stellar.cpp
esiragusa: Hmm _set_output_format() was necessary in older VS and is now deprecated in VS2015, right? If so, ```#ifdef PLATFORM_WINDOWS_VS``` becomes ```#if _MSC_VER < 1900```.

include/seqan/basic/allocator_interface.h
esiragusa: Hmm the problem must be somewhere above... allocate() does not have to initialize the memory.

See http://docs.seqan.de/seqan/develop/class_Allocator.html#Allocator#allocate:
"These objects are not constructed by allocate."

include/seqan/platform/platform_windows.h
```diff
@@ -71,6 +71,11 @@
 #pragma warning( disable : 4345 )
 #pragma warning( disable : 4351 )
 
+// allow multiple constructors
+#pragma warning( disable : 4521 )
+// allow functions like strncpy without warning
+#define _CRT_SECURE_NO_WARNINGS
```
esiragusa: Where's strncpy() used? :bomb:

catkira: many places, for example in \include\seqan\basic\debug_test_system.h

include/seqan/platform/platform_windows.h
```diff
@@ -71,6 +71,11 @@
 #pragma warning( disable : 4345 )
 #pragma warning( disable : 4351 )
 
+// allow multiple constructors
+#pragma warning( disable : 4521 )
```
esiragusa: Why do you need to suppress this warning? I don't see it in our nightlies (e.g. http://cdash.seqan.de/viewBuildError.php?type=1&buildid=138468) and is not specific to VS2015, right?

include/seqan/basic/allocator_interface.h
catkira: I think maybe some higher layers rely on it being initialized to 0 even if they should not. One way to test it, would be to do a memset 0xFF and see if the unit_tests on GCC and CLANG are still ok.

include/seqan/platform/platform_windows.h
catkira: I get this warning many times on vs2015, if it doesnt appear on any other version of vs, its probably vs2015 specific. Do you have the goal to do warning free builds on any supported compiler?

apps/stellar/stellar.cpp
catkira: ah yes, ok I will change it.

include/seqan/sequence/adapt_stl_container.h
```diff
@@ -585,6 +585,10 @@ capacity(std::array<TChar, N> const & me)
 // Function empty
 // ----------------------------------------------------------------------------
 
+// VC2015 implements some C++17 functions which would collide for 
+// applications that do using namespace std
+// http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4280.pdf
+#if _MSC_VER < 1900
```
esiragusa: I think that VS2015 does not imply any of C++11/14/17.
You should be able to instruct VC to respect the C++11 (or 14) standard, like the flag -std=... for gcc, right?
I don't have/use VC... maybe this helps: https://msdn.microsoft.com/en-us/library/jj851139.aspx.

include/seqan/stream/buffered_stream.h
```diff
@@ -232,14 +232,14 @@ class BufferedStreamBuf :
             numPutback * sizeof(TValue));
 
         // read new characters
-        size_t numRead = streamBufPtr->sgetn(
+		std::streamsize numRead = streamBufPtr->sgetn(
```
esiragusa: :+1:

include/seqan/stream/buffered_stream.h
```diff
@@ -232,14 +232,14 @@ class BufferedStreamBuf :
...
                             &buffer[putbackSize],
                             buffer.size() - putbackSize);
 
         // reset buffer pointers
         setg(&buffer[putbackSize - numPutback],     // beginning of putback area
              &buffer[putbackSize],                  // read position
-             &buffer[putbackSize + numRead]);       // end position
+             &buffer[putbackSize + static_cast<unsigned int>(numRead)]);       // end position
```
esiragusa: static_cast<size_t> ?

include/seqan/basic/iterator_adapt_std.h
catkira: done

apps/stellar/stellar.cpp
catkira: done

include/seqan/platform/platform_windows.h
esiragusa: Hmm it seems quite old: https://msdn.microsoft.com/en-us/library/e02fwk4s(v=vs.120).aspx

catkira: I wonder why this warning did not show up in vs2013, anyway it should be disabled or what do you think?

include/seqan/sequence/adapt_stl_container.h
catkira: I did not find an option in vs to set a specific c++ standard, looks like other people had this problem too:
http://stackoverflow.com/questions/23386472/disable-c11-features-in-vs2013

include/seqan/stream/buffered_stream.h
catkira: done

include/seqan/basic/allocator_interface.h
catkira: I did some test commits to see the GCC and Clang results, test_sequence fails on Clang if I do 	memset((void*)data, 0xFFFFFFFF, count * sizeof(TValue));
So the must be a bug somewhere in seqan (seqan relying on zero initialized memory). I dont know how much performance penalty this additional memset is, maybe somebody who know this low level code should look into it???

catkira: Now after I changed it back to memset 0, razers and microrazers tests are failing on Clang, but they were ok yesterday with the same code. How can this be?

esiragusa: Hmm razers or razers3? The microrazers problem might be unrelated...

catkira: https://travis-ci.org/seqan/seqan/builds/74372145
razers and micro_razers, but in GCC its all fine now. Very strange

esiragusa: Okay, we will investigate this issue. But for now, please remove the calls to memset from this PR. Thanks!

catkira: ok, its removed

include/seqan/sequence/adapt_stl_container.h
esiragusa: OMG :scream: OK let's keep it like this :+1:

include/seqan/platform/platform_windows.h
esiragusa: Could you please add ```#pragma warning( disable : 4521 )``` to your project only for now? I would better open an issue and investigate it whenever we have VS2015 installed.

catkira: ok done, I also removed the #define _CRT_SECURE_NO_WARNINGS. 
I noticed that most of my warnings appear when i build all the demos. The demos are not built on travis. Why are they excluded there?

catkira: The nightly builds also show all the warnings for older visual studio versions
http://cdash.seqan.de/viewBuildError.php?type=1&buildid=138611
How to handle those warnings?

esiragusa: Oh the warnings happen on develop but not on master... @rrahn @h-2 any ideas why?

esiragusa:
Failing CI is unrelated (travis went out of memory).

PR #1075, opened 2015-08-06
Added pkg-config support and augmented the SeqAn build system

SoapZA:
Dear SeqAn maintainers,
I have integrated pkg-config support now. In addition, I have excluded building the demos, apps and dox when users specify "-DSEQAN_BUILD_SYSTEM=SEQAN_RELEASE_MINIMAL" in order to prevent unnecessary building of apps. My take is that most linux distributions and macports only want to install the headers for their users without the apps.

Furthermore, the standard SeqAn cmake build system does not install the headers, which is non-standard and requires further intervention on the side of the user. Standard best practice should be to always install headers. Macports even copies them manually:
https://trac.macports.org/browser/trunk/dports/science/seqan/Portfile?order=name

Finally, I think it would be a good idea to have a crude API versioning scheme in place: namely instead of stuffing all headers in ${include_dir}/seqan, they should be put into ${include_dir}/seqan2, with the effect that multiple major version SeqAn installations can be used in parallel. Changing this should be trivial, and given the degree to which interfaces changed between 1.4 and 2.0, I think such a forward-looking approach is desirable for the sake of the stability of older software requiring certain interfaces without impeding the progress of the library overall. This is for instance how gtk-2 and gtk-3 can co-exist peacefully on one system. As such, I have called the .pc file seqan-2.pc on purpose, to allow for such changes to happen gracefully in the future.


h-2:
The question of whether the default should install the headers has been discussed in other issues and the current stance is that we have different targets for building the apps or installing the library. Also the global includedir cannot be changed anymore for seqan-2.* but it will possibly be changed for seqan3 to be include/seqan3

Regarding the pkg-config, can we merge that for 2.0.1, @rrhan?

CMakeLists.txt
```diff
@@ -163,3 +163,18 @@ include (package)
 # ===========================================================================
 
 include (SeqAnCtdSetup)
+
+# ===========================================================================
+# Install all C++ header files
+# ===========================================================================
+
+install (DIRECTORY include/seqan DESTINATION "${CMAKE_INSTALL_PREFIX}/include")
+
```
rrahn: Can you please revert the changes above and only keep the additions for the pkg-config.
We have special build flags for packaging. See here: http://seqan.readthedocs.org/en/master/BuildManual/UsingTheSeqAnBuildSystem.html#packaging-library-releases

SoapZA:
I've restored the package defaults and all edits now are solely to integrate pkg-config support.

esiragusa:
Shouldn't the pkg-config file be simply called seqan.pc? Does pkg-config derive the install path from the .pc filename?

esiragusa:
Note that we would need different paths for 1.4, 2.0, 2.1, etc. Only minor releases (eg. 2.0.x) are stable.

SoapZA:
hey Enrico,
as Hannes probably agrees, a versioning scheme is a good idea, it's just too late to be used for seqan-2. Instead of changing the includedir, we can at least version the seqan.pc file. Looking forward, I imagine things going like this:
SeqAn-2.X will get installed in include/seqan and have seqan-2.pc
SeqAn-3.X will get installed in include/seqan3 and have seqan-3.pc
This will allow for two (and more) major versions of SeqAn to be installed at the same time, which is in my opinion highly desirable. If you only use seqan.pc, you will not be able to version the pkg-config file going forward. On which level you actually version it is up for debate (if you think the API changes drastically between 2.0 and 2.1, then I guess seqan-2.0.pc and seqan-2.1.pc is more sensible. ).

Finally, the install path of pkg-config is encoded in the file (not in the filename, except for the major.minor version).

h-2:
@SoapZA Thanks again for the pull request! I would like to get this in for 2.0.2! Can you move the pkg-config folder into util? So the file would be at `util/pkg-config/seqan-2.pc.in` ?
Also can you make sure that thew whole thing is only done when `(("${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE") OR   ("${SEQAN_BUILD_SYSTEM}" STREQUAL SEEQAN_RELEASE_LIBRARY"))` ?
Because we don't want this to be installed when building an app...

esiragusa:
Note that the file has to be called **seqan.pc**. Why should otherwise the filename contain a version number? The content of the file will be updated, from release to release, to point to a folder called like **seqan-1.4** or **seqan-2.0**, as our workflow enforces api stability across minor releases.

h-2:
Since the (rough) plan for seqan3 is to possibly coexist with seqan2 for a while, I think it would be good, if we use seqan-2 here, and then later have seqan-3.

esiragusa:
Are you referring to the installation path or .pc filename?

SoapZA:
I agree, this is like gtk+-3 and gtk+-2, they can only coexist because their names are different. If the APIs between Seqan-2 and 3 are likely to be so different that recompilation of even the simplest packages between both versions will fail, then a filename versioning is absolutely necessary. Again, this is good practice and advised, see for instance

https://tecnocode.co.uk/2014/12/09/a-checklist-for-writing-pkg-config-files/
https://developer.gnome.org/programming-guidelines/unstable/parallel-installability.html.en

I will change the file layouts later.

h-2:
> Are you referring to the installation path or .pc filename?

One implies the other.

> Again, this is good practice and advised, 

Yes, I agree.

> I will change the file layouts later.

Great, thanks!

esiragusa:
> > Are you referring to the installation path or .pc filename?

> One implies the other.

OK I'll tell you tomorrow... apparently I am not able to explain it to you guys :scream:

SoapZA:
Let's have a look at GTK then:
You might be aware that GTK+2 and GTK+3 are generally API incompatible, i.e., you cannot compile apps designed around GTK+2 against GTK+3, due to GTK+3's use of opaque pointers for encapsulation. Nonetheless, GTK+2 and GTK+3 can perfectly co-exist on one system. How does this work? Let's have a look at their .pc files:

**/usr/lib/pkgconfig/gtk+-2.0.pc**:
<snip>
```
Libs: -L${libdir} -lgtk-${target}-2.0
Cflags: -I${includedir}/gtk-2.0
```
</snip>

**/usr/lib/pkgconfig/gtk+-3.0.pc**:
<snip>
```
Libs: -L${libdir} -lgtk-3
Cflags: -I${includedir}/gtk-3.0
```
</snip>

Such API versioning is simply not possible with just one **seqan.pc** file. You require at least some seqan-**MAJOR**.pc scheme.

esiragusa:
@SoapZA thank you for providing additional input. @h-2 and me will discuss it in person so that we understand can each other :sweat_smile:

esiragusa:
@SoapZA @h-2 please see #1278.

PR #1078, opened 2015-08-07
[FIX,DOC] Jst: Fixes broken dox entries.

rrahn:


rrahn:
Build errors unrelated. dox is fixed.

PR #1079, opened 2015-08-07
New Application: Flexbar++ (complete rewrite, new features, architecture improvement, 100% speedup)

catkira:
… n bases barcodes at the 5` end will be removed and stored in the fasta id for further processing.

[CLEANUP] removed some redundant code, its still not perfect but should be maintanable now. Speed optimization is not so important here, since the main bottleneck is the zlib write routine.

Idea: adding a multithreaded zlib outputstream to seqan would also benefit many other projects


esiragusa:
Thank you for this PR.

My first comments:
* Please follow our fascist code style guide at http://seqan.readthedocs.org/en/master/StyleGuide/Cpp.html (e.g. use 4 spaces instead of tabs)
* Better remove old code instead of commenting it out
* Why did you add a new command line option ```-tn``` specific to chip-nexus barcodes? What's the difference from standard trimming?

I'll comment more on the diffs.

.gitignore
```diff
@@ -5,3 +5,4 @@ sandbox
 !sandbox/CMakeLists.txt
 *.pyc
 *~
+apps/razers2/tests/razers2.exe
```
esiragusa: Please remove this one.

apps/seqan_flexbar/adapter_trimming.h
```diff
@@ -87,12 +87,17 @@ struct Mode {};
 struct Auto : Mode {};
 
 //Tagging struct representing the the match algorithm working
-//with values supplied by the user. Saves those values as members.
+//with values supplied by the user. Saves those  values as members.
```
esiragusa: OMG This was extremely ugly :scream:

Is it much work to change it? I have no clue as it's the first time that I look into this app... :weary: 
I wouldn't do any cast nor pointer/new. Maybe use only the struct User for options and add a boolean auto to it? Where is tag dispatching needed?

apps/seqan_flexbar/adapter_trimming.h
```diff
@@ -343,7 +348,7 @@ void alignAdapter(seqan::Pair<unsigned, seqan::Align<TSeq> >& ret, TSeq& seq, TA
 }
 
 //Version for automatic matching options
-inline bool isMatch(int overlap, int mismatches, const Auto &)
+inline bool isMatch(int overlap, int mismatches, const Auto *)
```
esiragusa: Please pass arguments by reference only.

apps/seqan_flexbar/adapter_trimming.h
```diff
@@ -357,13 +362,16 @@ inline bool isMatch(int overlap, int mismatches, const Auto &)
 	return mismatches <= errors;
 }
 //Overload for user-definied matching options
-inline bool isMatch(int overlap, int mismatches, const User& userOptions)
+inline bool isMatch(int overlap, int mismatches, const User* userOptions)
```
esiragusa: Same here and below.

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
 }
 
 
-template<typename TSeqs, typename TIds> //Version for single end data
-void preTrim(TSeqs& seqs, TIds& ids, unsigned head, unsigned tail, unsigned min, GeneralStats& stats)
+template<typename TSeqs, typename TIds>
+void _preTrimRemove(const std::vector<TSeqs*> &seqsVector, const std::vector<TIds*> &idsVector, StringSet<bool> const& rem, DemultiplexingParams& demultiplexParams, GeneralStats& stats)
```
esiragusa: * Avoid pointers and vectors of pointers.
* Use full template types whenever possible, e.g. TSeqsVector, TIdsVector, ...

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
-            ++ex;
-        }
-    }
-    if (ex != 0)
-    {
-        resize(seqs, limit - ex);
-        resize(ids, limit - ex);
-        stats.removedSeqsShort += ex;
-    }
+	assert(!seqsVector.empty() && !idsVector.empty());
```
esiragusa: Use SEQAN_ASSERT_* (see http://seqan.readthedocs.org/en/master/Tutorial/WritingTests.html?highlight=assert#assertions)

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
+	assert(seqsVector.size() == idsVector.size());
+	unsigned ex = 0;
+	unsigned limit = length(**seqsVector.begin());
```
esiragusa: ??

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
+	bool demultiplex = (length(demultiplexParams.multiplexFile) > 0);
+	for (int j = limit - 1; j >= 0; j--)
+	{
+		if (rem[j])
```
esiragusa: I do not understand the code below. Could you please add a comment explaining what it does?

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
+		stats.removedSeqsShort += ex;
+	}
 }
 
-//Overload for single end data with multiplex barcodes
-template<typename TSeqs, typename TIds, typename TMulti>
-void preTrim(TSeqs& seqs, TIds& ids, TMulti& multiplex, unsigned head, unsigned tail, unsigned min, GeneralStats& stats)
+// main preTrim function
+template<typename TSeqs, typename TIds>
+void _preTrim(TSeqs& seqs, TIds& ids, unsigned head, unsigned nexus, unsigned tail, unsigned min, StringSet<bool>& rem)
```
esiragusa: Please use a typename TRem. By the way, StringSet works over container types only, e.g. ```StringSet<String<bool>>```. Do you mean ```String<bool>```?

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
-        resize(seqs, limit - ex);
-        resize(ids, limit - ex);
-        resize(multiplex, limit - ex);
-        stats.removedSeqsShort += ex;
-    }
+	resize(rem, limit);
+
+	// disable this omp, because making a local copy of ids for every threat is not worth it
+	//SEQAN_OMP_PRAGMA(parallel for default(shared) private(i) schedule(static))
+	for (i = 0; i < limit; i++)
```
esiragusa: Why would you need a local copy of ids? Threads would work on mutually exclusive ```i``` thus ```ids[i]```, isn't it?

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
-    {
-        resize(seqs, limit - ex);
-        resize(seqsRev, limit - ex);
-        resize(ids, limit - ex);
-        resize(idsRev, limit - ex);
-        stats.removedSeqsShort += (2 * ex);
-    }
+	StringSet<bool> rem;
+	_preTrim(seqs, ids, head, nexus, tail, min, rem);
+	std::vector<TSeqs*> seqsVector;
```
esiragusa: I didn't get why you use these vectors of pointers instead of seqs/ids (see above)...

apps/seqan_flexbar/general_processing.h
```diff
@@ -483,474 +483,240 @@ void processN(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TMulti& mult
...
+	seqsVector.emplace_back(&seqs);
+	idsVector.emplace_back(&ids);
+	_preTrimRemove(seqsVector, idsVector, rem, demultiplexParams, stats);
+}
 
+// overload for paired end multiplex
+template<typename TSeqs, typename TIds, typename TDemultiplexingParams>
+void preTrim(TSeqs& seqs, TIds& ids, TSeqs& seqsRev, TIds& idsRev, TDemultiplexingParams&& demultiplexParams, unsigned head, unsigned nexus, unsigned tail, unsigned min, GeneralStats& stats)
+{
+	StringSet<bool> rem1, rem2;
```
esiragusa: ```String<bool>```?

apps/seqan_flexbar/seqan_flexlib.cpp
```diff
@@ -257,11 +271,17 @@ void ArgumentParserBuilder::addAdapterTrimmingOptions(seqan::ArgumentParser & pa
         "np", "no-paired", "Trim paired-end input with single-end trimming method.");
     addOption(parser, pairedModeOpt);
 
-    seqan::ArgParseOption rateOpt = seqan::ArgParseOption(
+    seqan::ArgParseOption errorOpt = seqan::ArgParseOption(
```
esiragusa: Isn't the error-rate parameter convenient?

.gitignore
catkira: yes, that was lame :+1:  
Its removed now

apps/seqan_flexbar/seqan_flexlib.cpp
catkira: There Error-rate parameter that existed was not really a "rate" parameter, it should be called maximum number of allowed errors. However most papers specifiy an error rate, which is the -e option in the cutadapt tool. It is (maximum number of allowed errors)/overlap. In order not to break existing use cases of flexbar, i called the new option -er.

apps/seqan_flexbar/general_processing.h
catkira: ok, changed

apps/seqan_flexbar/general_processing.h
catkira: It does what this code did before:
-    unsigned ex = 0;
-    for (int j = length(rem) - 1; j >= 0; --j)
-    {
-        if (rem[j])
-        {
-            swap(seqs[j], seqs[limit - ex - 1]);
-            swap(seqsRev[j], seqsRev[limit - ex - 1]);
-            swap(ids[j], ids[limit - ex - 1]);
-            swap(idsRev[j], idsRev[limit - ex - 1]);
-            ++ex;
-        }
-    }
-    if (ex != 0)
-    {
-        resize(seqs, limit - ex);
-        resize(seqsRev, limit - ex);
-        resize(ids, limit - ex);
-        resize(idsRev, limit - ex);
-        stats.removedSeqsShort += (2 * ex);
-    }

catkira: I seperated this code out and created another function for it, so i was able to remove a lot of copy-pasted code.

apps/seqan_flexbar/general_processing.h
catkira: hm yes thats true. I dont know the internal design of the underlaying structure, if its designed so that different threads can work on different elements at the same time, then it should be ok. I guess deleting or inserting would be problematic, but changing the content is ok right?

apps/seqan_flexbar/general_processing.h
catkira: Because then I can use the same _preTrimRemove function, for single and double strands.

For double strands I do:
	seqsVector.emplace_back(&seqs);
	seqsVector.emplace_back(&seqsRev);
	idsVector.emplace_back(&ids);
	idsVector.emplace_back(&idsRev);
	_preTrimRemove(seqsVector, idsVector, rem1, demultiplexParams, stats);

Actually it would be better to make a vector of THasIteratorObjects, then I could do:
	objectsVector.emplace_back(&seqs);
	objectsVector.emplace_back(&seqsRev);
	objectsVector.emplace_back(&ids);
	objectsVector.emplace_back(&idsRev);
        if (demultiplex)
              objectsVector.emplace_back(&demultiplexParams);
	_preTrimRemove(seqsVector, idsVector, rem1, demultiplexParams, stats);

and the _preTrimRemove would also be much simpler. Is there a common Iterator-Class for seqan::StringSet's ?

apps/seqan_flexbar/general_processing.h
catkira: Ah yes, I have to get used to seqan container types.
I changed it.

catkira:
The tn options says, that there is a random barcode at the 5" end, which should not be just deleted, but the content should also be stored in the id of the read, so that it can be evaluated later after mapping.

apps/seqan_flexbar/adapter_trimming.h
catkira: I changed it, now its not so ugly anymore :)
BTW: does seqan::StringSet have an iterator?

apps/seqan_flexbar/adapter_trimming.h
catkira: changed

apps/seqan_flexbar/adapter_trimming.h
catkira: changed

esiragusa:
Then I would remove ```-tn``` and add instead another option to store barcodes trimmed by ```-tr``` or ```-tl``` in the read id, because such functionality is not specific to chip-nexus reads...

apps/seqan_flexbar/CMakeLists.txt
```diff
@@ -16,6 +16,14 @@ set (SEQAN_APP_VERSION "1.0.2")
 # Dependencies
 # ----------------------------------------------------------------------------
 
+find_package(CXX11 REQUIRED)
+
+# Disable flexbar if no CXX11 available.
+if (NOT CXX11_FOUND)
+    message (STATUS "  flexbar cannot be built with MinGW or with MS Visual C++ 8 and earlier.")
```
esiragusa: A correct message would be: ```C++11 compiler required to build flexbar```.

catkira:
Yes, it would be possible to do a "store trimmed bases" option, but then there would need to be 2 options, one for tl and one for tr. The id string could look like "tl:CATCAT;tr:TACTAC"
The options for flexbar would be "-ttl" and "-ttr", tag trim left/right. Is that good?

esiragusa:
Why not just one ```-tt``` to save them both?

apps/seqan_flexbar/seqan_flexlib.cpp
esiragusa: Ah you added an error rate parameter and renamed the old parameter as (absolute) errors :+1:

catkira:
yes, i can make it -tt, sounds good

apps/seqan_flexbar/seqan_flexlib.cpp
catkira: yes, because the old e parameter was actually about absolute errors, so i left the function the same and just changed the description of it.

apps/seqan_flexbar/CMakeLists.txt
catkira: corrected

h-2:
I think major changes to apps like rewrites, renaming etc should not happen inside the SeqAn tree anymore. Also we now have https://github.com/seqan/flexbar ... How are they related? Should these changes not be applied there?

h-2:
@catkrira, just to followup on this: have you spoken with @jtroehr ? Shall this be integrated into http://github.com/seqan/flexbar ? Or do you want your own seqan/flexbar++ ?

catkira:
I have talked to @jtroehr , our conclusion was that I host flexbar++ as separate app outside seqan/seqan and he might integrate some usefull parts of flexbar++ into his new implementation.
For me both would be ok, move flexbar++ into seqan/flexbar++ or leave it on my private git. Since I dont plan to become a preprocessing guru in the long run, It would be also okay for me if somebody else became maintainer of flexbar++.

h-2:
I can set you up with you own repository, but I think having seqan/flexbar and seqan/flexbar++ will be a little confusing, as it implies that ++ is always the newer version, although both are actively maintained. Can the two of you discuss this and agree on different names, e.g. flexbar_jtr and flexbar_bm, or flexbar and flexbar_cr/flexbar_catkira... ?
Also it would be good if both apps' READMEs state that there are two seqan based flexbars and how they differ.

Thank you!

catkira:
i also think that 2 versions would be confusing and a waste of energy. So I think the plan is that @jtroehr s version will be the official flexbar. In the meantime until flexbar is ready, I can rename my flexbar++ to some other name, to avoid confusion.

h-2:
Ok, thats great. If you have a name (flexbar_catkira?), just let me know and I will create the repository for you. You will then have write access and not need any of us to accept pull requests.

catkira:
I think the underscore in names always look a bit ugly. Is flexcat okay?

h-2:
sure :)

PR #1082, opened 2015-08-11
[FIX] More fixes for windows builds.

rrahn:


rrahn:
Upps, this isn't necessary at all. I'll withdraw this.

Issue #583, opened 2014-10-07
GFF AnnotationTree bug


cfblaeb:
When parsing a NCBI GFF file (I specifically am trying these ftp://ftp.ncbi.nlm.nih.gov/genomes/Cricetulus_griseus/GFF/) using AnnotationTree I noticed that the structure of the tree did not become root-->gene-->mRNA-->exons/cds, but rather root-->gene and root-->mRNA-->exons/cds
The gene and the mRNA have correct parent/id pairs, but if the mRNA line contains a transcript_id then that is used instead of id.
When I remove/rename transcript_id it works correctly.
According to the gff3 specifications, transcript_id should NOT hold any special meaning and only Id should be used in determining parent/child relations. http://www.sequenceontology.org/gff3.shtml

esiragusa:
@cfblaeb This issue should be finally fixed :sweat_smile:

PR #1093, opened 2015-08-13
Fix FragmentStore parsing of Gff3 files

esiragusa:
FragmentStore constructs a wrong AnnotationTree when parsing Gff3 files containing transcript_id fields. See #583.


rrahn:
lgtm!

PR #1104, opened 2015-08-13
Manual: added a BAM index tutorial

xenigmax:
I added explanation section about BAM index according to https://github.com/holtgrewe/seqan-wiki/blob/master/Tutorial%252FSamBamIO.mediawiki


catkira:
It would be nice to explain how to create a bai index with seqan

rrahn:
@catkira please see #1108.

PR #1106, opened 2015-08-13
Vs2015 compiler warnings

catkira:
fix compiler warnings


rrahn:
@catkira I was actually wondering if you could only make a PR for this change catkira@facc487 only as a hotfix into master :wink:. But for this you have to branch from master. VS 14 support will not go into the minor release, you probably know best :sweat_smile:. I will add this change, so don't worry if I close this PR. We can later add this to develop. So your work is not lost. :thumbsup:

PR #1115, opened 2015-08-14
BAM index file (.bai) can be saved via _saveIndex()

### LINKED ISSUE #1081: Saving bai index
Hi,

I have been using the `_saveIndex()` function for a `BamIndex<Bai>` although it is currently only called in a function that starts with (bam_io/bam_index_bai.h, line 614):
```
SEQAN_FAIL("This does not work yet!");
```

The `_saveIndex()` function is almost working fine for me, only line 573
```
__int32 numIntervals = length(index._linearIndices);
```
needs to be fixed to
```
__int32 numIntervals = length(linearIndex);
```
.

Cheers, Birte

xenigmax:
This will close #1081 .
However, this is a internal function(with "_") which should not be used directly. 
We must work on "buildIndex" and other functions to support BAM index properly.

### DIFF
--- a/include/seqan/bam_io/bam_index_bai.h
+++ b/include/seqan/bam_io/bam_index_bai.h
@@ -529,7 +529,7 @@ open(BamIndex<Bai> & index, char * filename)
 
 inline bool _saveIndex(BamIndex<Bai> const & index, char const * filename)
 {
-    std::cerr << "WRITE INDEX TO " << filename << std::endl;
+    //std::cerr << "WRITE INDEX TO " << filename << std::endl;
     // Open output file.
     std::ofstream out(filename, std::ios::binary | std::ios::out);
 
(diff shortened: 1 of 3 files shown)


include/seqan/bam_io/bam_index_bai.h
```diff
@@ -529,7 +529,7 @@ open(BamIndex<Bai> & index, char * filename)
 
 inline bool _saveIndex(BamIndex<Bai> const & index, char const * filename)
```
esiragusa: @xenigmax would you please rename it as ```save()``` and add a doc entry?

esiragusa:
Ideally ```buildIndex()``` should work in memory. I would remove the filename argument and remove the last three lines writing the file.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -529,7 +529,7 @@ open(BamIndex<Bai> & index, char * filename)
...
 {
-    std::cerr << "WRITE INDEX TO " << filename << std::endl;
+    //std::cerr << "WRITE INDEX TO " << filename << std::endl;
```
esiragusa: You can remove these lines!

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -578,7 +578,7 @@ inline bool _saveIndex(BamIndex<Bai> const & index, char const * filename)
     }
 
     // Write the number of unaligned reads if set.
-    std::cerr << "UNALIGNED\t" << index._unalignedCount << std::endl;
+    //std::cerr << "UNALIGNED\t" << index._unalignedCount << std::endl;
```
esiragusa: This too!

PR #1134, opened 2015-08-17

catkira:


esiragusa:
@catkira thanks for the PR! I would make it an hotfix. Then you would have to:
* branch from master
* commit your changes
* open two PRs from your hotfix branch: one to master and one to develop

PS You might want to have a look at https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow and http://seqan.readthedocs.org/en/master/Infrastructure/SeqAnWorkflow.html

PR #1135, opened 2015-08-17
Fixes failing splazers test.

rrahn:
This fixes the splazerS test when running with the GappedXDrop extension.

The problem here is, that the function ```_extendSeedGappedXDropOneDirection ``` expects the query and data base sequence in opposite order to the seed definition.
So what comes first into the seed comes last into the extension function. :scream:


rrahn:
Travis build errors unrelated.

PR #1137, opened 2015-08-18
[improvement] before it was not possible to use the OMP PRAGMA for readRecords, bec…

catkira:
…ause readRecord used the file context buffer internally. With these changes, reading records has about 20% speedup.

In the feature it might be good, to move the pubsetbuf stuff into a wrapper function in FileContext so that its better customizable.


catkira:
didnt work for sam files yet.

PR #1140, opened 2015-08-18

catkira:

(cherry picked from commit 37cbf3c8d37b1232f76fb51003db3bc30c0c4b5b)


include/seqan/bam_io/bam_file.h
```diff
@@ -248,10 +248,10 @@ readRecords(TRecords & records, FormattedFile<Bam, Input, TSpec> & file, TSize m
 {
     String<CharString> & buffers = context(file).buffers;
     if ((TSize)length(buffers) < maxRecords)
-    {
         resize(buffers, maxRecords, Exact());
+    if ((TSize)length(records) < maxRecords)
```
rrahn: Could you please use static_cast according to our style guide http://seqan.readthedocs.org/en/latest/StyleGuide/Cpp.html#static-cast. 
It would be great if you could change it above in line 250 as well.
:+1:

rrahn:
Thanks :+1:

PR #1141, opened 2015-08-18

catkira:

(cherry picked from commit 37cbf3c8d37b1232f76fb51003db3bc30c0c4b5b)


catkira:
I updated it according to style guidelines

PR #1146, opened 2015-08-19
[FIX] Disabling splazers on 32 bit arch.

rrahn:


catkira:
maybe also disable yara on 32 bit app, because it very quickly runs out of memory there and the error message is not very enlightening.

PR #1147, opened 2015-08-20
[FIX] header was not cleared in readBamHeader

catkira:


esiragusa:
Nice catch. The fix looks obviously correct, but MicroRazerS tests failed (see http://cdash.seqan.de/testDetails.php?test=26769722&build=140837). At best, the bam files used for testing are wrong and must be re-generated.

catkira:
If I run the microrazers test locally its all OK. Are you sure it has nothing to do with travis out of memory or something like that?

PR #1150, opened 2015-08-20
[WARNINGS] compiler warnings for master builds

catkira:


rrahn:
Oh, actually, having both variants is not really a good solution here I guess.
I will dig into the build system to see why there are different behaviours for the master and the develop builds.

PR #1152, opened 2015-08-21
Position ModifiedString and small fixes for other lightweight containers

esiragusa:
* Implemented a lightweight Position ModifiedString
* Solved some issues with const/non-const lightweight objects
* Improved the Segment StringSet interface
* Fixed CountingIterator metafunctions
* Wrapped more STL algorithms

### DIFF
--- a/include/seqan/basic/iterator_position.h
+++ b/include/seqan/basic/iterator_position.h
@@ -147,10 +147,19 @@ template <typename TContainer>
 struct Position<Iter<TContainer, PositionIterator> > : Position<TContainer> {};
 
 template <typename TContainer>
+struct Difference<Iter<TContainer, PositionIterator> > : Difference<TContainer> {};
+
+template <typename TContainer>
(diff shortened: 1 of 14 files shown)


include/seqan/basic/iterator_position.h
```diff
@@ -168,10 +177,10 @@ container(Iter<TContainer, PositionIterator> & me)
 }
 
 template <typename TContainer>
-inline typename Parameter_<TContainer>::Type
+inline typename Parameter_<TContainer const>::Type
 container(Iter<TContainer, PositionIterator> const & me)
 {
-    return _toParameter<TContainer>(me.data_container);
+    return _toParameter<TContainer const>(me.data_container);
```
rrahn: Is that correct? To make the container const? I thought, a const-iterator can return a non-const container. The constness of the container type should be independent of the constness of the iterator, right?

PR #1156, opened 2015-08-31
[FIX] nasty little bug in evaluate_alignment

h-2:


esiragusa:
Hmm is there a test for it? Last week I tried to use it, discovered that it doesn't work, and implemented my own stats as this one is really crappy :trollface:

h-2:
Yeah, it works. I fixed the test now as well.

PR #1162, opened 2015-09-01
argument parser updates

### LINKED ISSUE #536: Usage of hidden options is not consistent in seqan apps
Originally from Stephan on Trac

The argument parser can hide options. This is used in two different ways

To hide developer-only/debug options
To hide advanced options
We should change this to hidden and advanced options. In this way way we could exclude the hidden options for the CTD export and tag the advanced options as advanced in the CTD.

h-2:
adds a couple of features to the arg-parser:
  * abilitiy to set options as "advanced", advanced options are hidden when looking at `--help`
  * add `--full-help` which shows the advanced options as well
  * shortCopyright field and citation field which are shown in new LEGAL section of the default `--help`-page
  * longCopyright field which enables a `--copyright` option that display the field (useful for license texts et cetera)
  * SeqAn copyright information is automatically added to both fields if they are used
  * fixes #536

~~Don't merge yet, as the functions for html and man-export are not yet adapted.~~

### DIFF
--- a/include/seqan/arg_parse/arg_parse_ctd_support.h
+++ b/include/seqan/arg_parse/arg_parse_ctd_support.h
@@ -339,6 +339,11 @@ writeCTD(ArgumentParser const & me, std::ostream & ctdfile)
          ++optionMapIterator)
     {
         ArgParseOption const & opt = *optionMapIterator;
+
+        // exclude hidden
+        if (isHidden(opt))
+            continue;
+
         std::string optionIdentifier = _getPrefixedOptionName(opt);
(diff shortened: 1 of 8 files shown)


h-2:
I have fixed the ctd generation and the tests now, too. For CTDs options marked as advanced are now also marked as advanced and hidden options are silently ignored. Since this changes the previous behaviour it would be good if someone familiar with KNIME check if our nodes still work after this is upstreamed.
If the tests pass this can be merged from my POV.

h-2:
@esiragusa can you review this today some time? I need this in develop to release a new version of lambda (as I depend on upstream now). Thanks!

PR #1168, opened 2015-09-09
[FIX] add overloads for const strings, as well

h-2:


esiragusa:
But should't open() be non-const ref only and save() const ref only?

h-2:
In theory yes, but since we have ridiculous global overloads for everything its usually advisable to overwrite both...
in this case however it seems to be ok ;) I will update it.

esiragusa:
Which global overloads? There too, open() must be & only and save() const & only.

h-2:
> Which global overloads?

I was referring to global overloads for things like insert and append which have versions for both const and non-const, although they shouldn't (I have removed many, but oh well). DIfferent topic.

h-2:
@esiragusa can I merge this?

esiragusa:
Well the diff is okay... but why do we need dummy open/save for modified strings?

h-2:
Because the ModifiedString doesn't contain any data, so it shouldnt write any data!

esiragusa:
Waait that's not true... :wink:

PR #1170, opened 2015-09-11
[FIX] make static builds work on freebsd

h-2:
by linking against libelf.


h-2:
this is rather trivial, could someone review it sometime soon? Thanks!

Issue #858, opened 2015-02-20
Move all cpp code blocks from the manual into files


esiragusa:
Many tutorials contain cpp code blocks (directive .. code-block:: cpp) rather than external compilable files (directive .. includefrags:: file.cpp).

Such code blocks are not built in nightly/ci and potentially broken.

See e.g. https://github.com/seqan/seqan/blob/master/manual/source/Tutorial/Indices.rst#assignment-1 or https://github.com/seqan/seqan/blob/master/manual/source/Tutorial/Sequences.rst.

rrahn:
@smehringer Can we close this?

smehringer:
@rrahn Not yet. I was busy otherwise and there are several code snippets that need to be looked at  carefully because they do not resemble an actual function or program. I think Hannes suggested to always  have a "basic.cpp"-file where all snippets are listed in main().

rrahn:
Ah great. Yes, something I also suggest.

smehringer:
Then I'll do that.

rrahn:
Maybe wait with this until next Monday.
We are basically restructuring the layout of the tutorials and the manual page.
We can then discuss the details about this.

Issue #809, opened 2015-01-28
Cleanup the demos


esiragusa:
See also #520.

* Enforce demos structure:
 * demos/tutorial/module/exercise.cpp
 * demos/tutorial/advanced_xxx/exercise.cpp
 * demos/dox/module/exercise.cpp
 * demos/unassigned/module/exercise.cpp

* Check which demos got duplicated, e.g. demos/index_sufarray.cpp and demos/index/index_sufarray.cpp
* Move demos which are not referenced by any tutorial, maybe move them into e.g. demos/[module]_[unassigned]/
* Rename weird target names, e.g. demo_index_index_sufarray to demo_index_sufarray
* Create a .stdout file for each demo

PR #1173, opened 2015-09-16
Restructured demo directory

### LINKED ISSUE #520: Restructure demo directory
The demos that support the API documentation should go into a subdirectory of "demos" that is the same as their module name.

We should also introduce a policy on which demo should go where in the hierarchy.

Here a proposal:

We want the following types of example/demo programs:

* **library driver applications** Complex applications that expose library functionality, e.g. the realignment. If they fit into one .cpp file, they should be put into the "demos" directory. Such a demo DEMO.cpp should be accompanied with a DEMO.README file that gives explanation.
* **API demo** Demo programs that are included by the API documentation. A demo for a function/class/etc. called codeItem from module the_module should be put into "demos/the_module/code_item.cpp".
* **Tutorial/How-To demo** Demo programs used in the wiki tutorial or How-To programs. They should be put into "demos/tutorial/tutorial_name/program.cpp" and "demos/howto/howto_name/program.cpp".
Thus only more complex demos are placed directly in the demos folder.

smehringer:
Fixes #520 #809.

Todo:
* Some .stdout files are empty because the corresponding demos fail to execute.
* Partially fix for #858. Other directive .. code-block:: cpp / console have to be converted into includefrags.


rrahn:
There is an empty file in apps/ called "=" which should be deleted. I cannot at a comment there because the file is empty :smile:

rrahn:
Ok when removing this one file, this can be merged.

PR #1175, opened 2015-09-16
Fix Markov model emitted probabilities calculation for string sets

JohnReid:
Probabilities for bases are multiplied inside Strings but the probabilities for entire strings are added. I believe they should also be multiplied. Also there was an old vim .swp file in the repository that looks like it was unneeded.


JohnReid:
Actually checking this I'm not sure if I understood the code correctly in the first place so I'm closing it. It might be worth looking into though.

PR #1178, opened 2015-09-17
use gaps internally for align_extend and integrateAlign

h-2:
This also fixes some things in alignExtend where it depended on broken behaviour and did unnecessary clears.

### DIFF
--- a/include/seqan/align/alignment_operations.h
+++ b/include/seqan/align/alignment_operations.h
@@ -55,6 +55,47 @@ namespace seqan {
 // Functions
 // ============================================================================
 
+template <typename TSource0, typename TGapSpec0,
+          typename TSource1, typename TGapSpec1,
+          typename TPos>
+inline void
+integrateGaps(Gaps<TSource0, TGapSpec0> & targetRow,
+              Gaps<TSource1, TGapSpec1> const & sourceRow,
+              TPos const viewPos)
+{
+    typedef typename Iterator<Gaps<TSource0, TGapSpec0> >::Type TTargetIt;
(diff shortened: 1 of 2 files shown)


include/seqan/align/alignment_operations.h
```diff
@@ -55,6 +55,48 @@ namespace seqan {
...
+    // This assertion ensures that the number of sequence characters after viewPos is greater than or equal to
+    // the number of source characters in the clipped infix row.
+    SEQAN_ASSERT_GEQ(endPosition(targetRow) - toSourcePosition(targetRow, viewPos),
+                     endPosition(sourceRow) - beginPosition(sourceRow));
+
+    // init iterators
+    TTargetIt it = iter(targetRow, viewPos);
+
+    // walk through Gaps containers and copy gaps
+    for (TSourceIt infixIt = begin(sourceRow, Standard()); !atEnd(infixIt);)
```
rrahn: Please, use the atEnd function consistently with the iterator flag. 
Semantically the code doesn't make sense, because you are getting a Standard iterator but call the rooted atEnd version. See here: http://docs.seqan.de/seqan/develop/?p=RootedIteratorConcept#RootedIteratorConcept%23container it is only concept of the rooted iterator.

rrahn:
Please respond to the minor issue. Otherwise this lgtm.

include/seqan/align/alignment_operations.h
h-2: That was copy'n'paste from before, but of course you are right.

PR #1180, opened 2015-09-18
[DOC] fix dox in develop

h-2:
sorry for breaking it earlier this week....


h-2:
Merged since it only effects dox and develop's dox is broken already (and might be fixed by this).

PR #1207, opened 2015-10-02
[FIX] many warnings on newer compilers

h-2:
fixes many warnings on newer gcc and clang. Note that I think it is kind of a mess that we have the same typedef for `__attribute__ ((unused))` defined so many times, but oh well.

### DIFF
--- a/include/seqan/basic/debug_test_system.h
+++ b/include/seqan/basic/debug_test_system.h
@@ -346,6 +346,16 @@ inline const char * toCString(Demangler<T> const & me)
 #define SEQAN_ENABLE_CHECKPOINTS 0 // SEQAN_ENABLE_TESTING
 #endif  // #ifndef SEQAN_ENABLE_CHECKPOINTS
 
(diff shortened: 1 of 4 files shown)


h-2:
removed the multiple instances of the macro definition.

include/seqan/basic/debug_test_system.h
```diff
@@ -346,6 +346,16 @@ inline const char * toCString(Demangler<T> const & me)
 #define SEQAN_ENABLE_CHECKPOINTS 0 // SEQAN_ENABLE_TESTING
 #endif  // #ifndef SEQAN_ENABLE_CHECKPOINTS
 
+#if (defined(__GNUC__) && ((__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 7)))) || defined(__clang__)
+#  define SEQAN_UNUSED __attribute__((unused))
```
esiragusa: Maybe better in platform.h?

PR #1214, opened 2015-10-06
Hotfix/win build warnings

rrahn:
Modified the cmake lists file to use PARENT_SCOPE for some variables set within function scope.

Also updated CXX_FLAGS for win32 applications instead of using add_definitions.


esiragusa:
@rrahn could you please prepare one PR for develop as well?

esiragusa:
@rrahn I will merge it before #1197 unless you object.

rrahn:
Ok added a PR for develop. You can merge it before.

PR #1219, opened 2015-10-07
[FIX] don't detect openmp on clang

h-2:


esiragusa:
@h-2 We have (had?) nightly builds under clang/openmp. We used this: http://clang-omp.github.io.

h-2:
i have fixed the previous build issues now, but test_parallel  still segfaults and nearly all our apps make clang crash during build:
https://llvm.org/bugs/show_bug.cgi?id=24674

This is fixed for 3.7.1, but since that hasn't shipped yet I propose deactivating clang for now. I will however update this pull, so that the build flag stuff works right, as soon as we reactivate it.

h-2:
> @h-2 We have (had?) nightly builds under clang/openmp. We used this: http://clang-omp.github.io.

long time ago ;)

esiragusa:
Is it only clang 3.7.0? Can't you deactivate only that version?

h-2:
> Is it only clang 3.7.0? Can't you deactivate only that version?

Thats the plan, although I can't verify that 3.7.1 will work, because I don't have a svn version installed that is new enough.

esiragusa:
I didn't see that one demo now fails to compile. See: http://cdash.seqan.de/viewBuildError.php?buildid=147471.

PR #1226, opened 2015-10-09
Hotfix/merge demo fixes from develop

rrahn:
In #979 fixes to the demos have been applied which were out of sync to the master branch.
Now, since we fixed the original issue with building the demos in #1214 for the master we had to fix the issues in master as well. Now, both fixes to master and the previous one to develop are out of sync so I cherrypicked the changes from develop to merge them into master to sync both branches.
Then we can apply remaining hot fixes and some changes to the previous merge in the usual fashion without merge conflicts into develop.


rrahn:
Ok the ci builds look good.

PR #1227, opened 2015-10-09
Hotfix/merge demo from dev

rrahn:
See #1226 for reference.


demos/dox/sequence/segment.cpp
```diff
@@ -75,7 +75,7 @@ int main()
 //![basic operations]
 
 //![metafunction examples]
-    typedef Infix<TInfix>::Type  TInfix2;  // == TInfix
+/*  typedef Infix<TInfix>::Type  TInfix2;  // == TInfix
```
esiragusa: Please remove ```/* */``` comments.

PR #1234, opened 2015-10-13
Hotfix: Removes all unused or unassigned demos.

rrahn:
I remove now the unused demos if no one objects.
In case one can restore them from this commit.


rrahn:
Please wait with this PR, as it is not automatically mergeable into develop.
I have to check what's going on here.

rrahn:
Ok, I fixed the merge conflict manually.

esiragusa:
Perhaps we should have a look at these demos. There might be important functionalities that are not yet described in the dox or manual. One example could be repeat search using suffix trees.

esiragusa:
Stream benchmarks could be moved to seqan/bench.

h-2:
hm, can you guys agree on something? I will review it when you are done :)

rrahn:
@h-2 could you also please have a look to check if there are any demos that might be kept for the future? I made this PR so everyone can have a look at the demos to decide which will be removed and which not. I wasn't quite specific with that, I agree :sweat_smile:

h-2:
@rrahn I have to prepare the lecture and will probably not have time to review any of the PRs, sorry.

esiragusa:
@rrahn I'll have a look at these demos.

esiragusa:
I would postpone this...

Some demos might go into apps, like
* bam utilities (sam2svg, bam*)
* format conversion utilities (gff2gtf, annotation_converter, soft_to_hard_masking)
* seqan/bench (benchmark_stream*)

Other demos could be integrated into existing or new tutorials like
* indexed repeat search (index_*)
* basic/custom I/O (tutorials_custim_io/*)

Ah, why do we have unused demos for tutorial_alignments and tutorial_sequence?

rrahn:
Well, might be that we had some old demos, used for old alignment tutorials which we never deleted in the old trac system and hence there wehere ported to the new manual as well.
I will have a look and remove them if not needed. Otherwise, I will fix the last errors that haven't been fixed for the unused demos.

PR #1247, opened 2015-10-21
[FIX] don't force the use of Skew here

h-2:
but instead use the documented metafunction.


h-2:
please do not merge yet, it is not fully clear whether this fixes the issues I am seeing.

h-2:
> please do not merge yet, it is not fully clear whether this fixes the issues I am seeing.

no, everything is good.

esiragusa:
It's OK for me. Please check TravisCI carefully. Then you can merge.

PR #1254, opened 2015-10-23
Cache TravisCI builds

esiragusa:
**Work in progress**


h-2:
@esiragusa Since this is rather complicated and you have spent time on this: Can you finish this before you go? Thanks!

h-2:
moved to backlog in favor of using ccache for now.

PR #1261, opened 2015-10-27
Fix/gustaf readme

ktrappe:
Created new file Readme.md. Left the old Readme file as it contains more details and file specifications.


esiragusa:
Please, convert your readme to rst (you can use http://pandoc.org/try/), rename it as README.rst and add this line to your CMakeLists.txt https://github.com/seqan/seqan/blob/master/apps/yara/CMakeLists.txt#L192

esiragusa:
Is this a hotfix? If so, you have to apply it to master.

ktrappe:
Why does it have to be rst? Because of the build error?
It is not necessarily a hotfix (it's a readme after all), can I still apply it to master?

esiragusa:
@ktrappe it's better to apply this as a hotfix, so master/develop won't diverge for your app.
@h-2 Then .md or .rst? Let's decide so that @ktrappe knows what to do...

esiragusa:
Closed in favor of #1281, #1282.

PR #1269, opened 2015-11-03
[FIX] adapter iterator for const strings

h-2:
previously the functions below returned Value* even if the container was const. The change directly makes dereferenciation go over the type that was defined for it instead of trying to use Value<>.

### DIFF
--- a/include/seqan/basic/iterator_adapt_std.h
+++ b/include/seqan/basic/iterator_adapt_std.h
@@ -158,13 +158,13 @@ public:
     // Pointer Operators;  Have to be defined within class.
     // ------------------------------------------------------------------------
 
-    typename Value<Iter>::Type *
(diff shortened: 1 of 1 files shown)


include/seqan/basic/iterator_adapt_std.h
```diff
@@ -158,13 +158,13 @@ class Iter<TContainer, StdIteratorAdaptor>
...
 
-    typename Value<Iter>::Type *
+    typename std::iterator_traits<TIterator>::pointer
     operator->()
     {
         return &*data_iterator;
     }
 
-    typename Value<Iter>::Type const *
+    typename std::iterator_traits<TIterator>::pointer const
```
esiragusa: I think you should use ```typename Pointer<Iter const>::Type```...

PR #1278, opened 2015-11-11
Added pkg-config support

esiragusa:
Replacement for #1075.

Note that SeqAn's include path should be something like ```/usr/include/seqan/2.1.0/``` for versioned pkg-config to work. We can add that as well... I just cannot see why it was not done in #1075.


SoapZA:
Dear Enrico,
I like the fact that the installation is moved into a separate directory within the seqan includedir. This will allow for co-existence of multiple header versions. Unfortunately, the pkgconfig file is not versioned, the reasons for which I believed to have illustrated with the example of GTK+2/3.

esiragusa:
It is! :)

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -199,6 +199,11 @@ macro (seqan_build_system_init)
                        README.rst
                        CHANGELOG.rst
                  DESTINATION share/doc/seqan)
+        # Install pkg-config file, except on Windows.
+        if (NOT CMAKE_SYSTEM_NAME MATCHES Windows)
+            configure_file("util/pkgconfig/seqan.pc.in" "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION}.pc" @ONLY)
```
esiragusa: Here automatic versioning happens.

esiragusa:
@SoapZA @h-2 would you please try on your systems too if everything works as expected?

esiragusa:
@SoapZA so you are also proposing to install to e.g. ```/usr/include/seqan/2.1.0/```, correct? Do you see any problems with it? This was not (yet?) in #1075.

SoapZA:
@esiragusa excuse my stupidity, I completely missed the renaming in the cmake. I've pulled your pkgconfig branch and tested it.

# pkg-config
Given that you clearly stated that you **guarantee** API stability between patch versions (i.e. from 2.1.0 to 2.1.1), we have a case of "overversioning". We should version the files, but only as much as necessary for capturing the same API and not more. See for instance, installing the library gives me `share/pkgconfig/seqan-2.1.0.pc`, which instead should be `share/pkgconfig/seqan-2.1.pc`, because we definitely do not need to have coexistence between patch versions. Whether you install the headers into `/usr/include/seqan/2.1.0/` or `/usr/include/seqan/2.1/` is a matter of taste. Choose whichever you prefer, the former might make it nice for the user to determine the seqan version without digging through files.

# Directory/header versioning scheme
## Most correct, but painful solution
Concerning the versioning of the .h files, just putting them into `/usr/include/seqan/2.1.0/` without further modifications is a recipe for disaster, as it will
1. require editing **ALL** `#include` statements in **ALL** .h and .cpp
2. (even worse) would hardcode the version in those files, which will become a maintenance nightmare, as things would go from `#include <seqan/bam_io.h>` to `#include <seqan/2.1.0/bam_io.h>`, which is in no way desirable. The idea should be to encode the version requirement in the build system (Makefile/Autoconf/Cmake/whatever) and have to change it in **one** place only, and **not** in any C++ source code.

I've studied the problem a bit more, and have come to the conclusion (surprise!) that GTK+ again has the **proper** solution:

1. At a minimum, prepend all **public** header filenames (i.e. those to be consumed by users) with `seqan_` i.e., `file.h` -> `seqan_file.h` (whether private headers should be renamed is a matter of debate, I still think it's desirable for reasons mentioned later)
2. Remove `seqan/` from all `#include`'s, i.e., change all source code files from
```
#include <seqan/misc/base.h>
```
to
```
#include <misc/seqan_base.h>
```
and then change in seqan.pc.in the `includedir=${prefix}/include/@SEQAN_VERSION_MAJOR@.@SEQAN_VERSION_MINOR@`

I am well aware that this is an *extremely invasive and painful* operation, which could probably still be mostly automated with `sed`. The reason why I think this needs to be done, is because headers should be namespace'd on the filename level and **not** on the directory level. Let's imagine we do not rename the files, and leave them as they are, then in future we will have our `#include`'s to be like
```
#include <misc/base.h>
```
`base.h` is way too generic a name to *not* cause a collision at some point down the road with some other package. Again, namespacing the headers is the GTK+ way of doing things: In `/usr/include/gtk-3.0/gtk/` all files are prepended with `gtk` such as `gtktypes.h` for instance (as `types.h` would be too generic). I believe this solution to be the best and most correct long-term solution, as it avoids header collisions.

## less painful, most compatible approach
If you consider this renaming too invasive, there is another less invasive solution to the problem. Instead of using `/usr/include/seqan/2.1.0/`, you can create `/usr/include/seqan-2.1/seqan` and then you pass the compiler the flag `-I/usr/include/seqan-2.1`, which will not require any renaming of files and not changing of any source code. This is close to the GTK+ model, without the namespacing but retaining most of the backward compatibility.

Tell me what you think of my ideas.

h-2:
> Given that you clearly stated that you guarantee API stability between patch versions (i.e. from 2.1.0 to 2.1.1), we have a case of "overversioning". [...]

I agree.

> Most correct, but painful solution

I don't think this would be the right way to go, but even if, I don't think we would do it, for the reasons you mentioned.

> less painful, most compatible approach

We could do this, but I still think it is a little overkill, for the following reasons:
  * there is 2.0 now, there will be 2.1 in January and there probably will be another 2.2; while they are not 100% compatible, they are very close
  * there will be seqan3 maybe in 2017h1, which is planned for co-existance with seqan2 and will likely have include/seqan3 as a path directly, so no problem there
  * we have no external apps in any free software package repository that depend on seqan right now, so we are not breaking anyone's compatibility; also we would not maintain different packages for different minor releases anyway
  * seqan is a header-only library so only build-time is important, not link-time, further reducing compatibility problems

I therefore propose:
  * for `seqan-2.*.*` we use `seqan-2.pc` and `${prefix}/include/seqan` as include path
  * for `seqan-3.*.*` we use `seqan-3.pc` and `${prefix}/include/seqan3` as include path

SoapZA:
@h-2 I agree with your solution, especially because you can *require* certain versions with pkg-config, so users can always say they want `>=2.2.0` but `<2.3.0` for instance. The only thing you will have to keep in mind once you start developing `seqan-3.0.0` is that you will have to rename **all** `#include <seqan/foo/bar>` to `#include <seqan3/foo/bar>`, including all consumers of seqan. If you are comfortable with such changes, by all means go for it. My second approach has the advantage that it only requires changing the includedir layout but no sourcecode changes for the library and the consumers.

h-2:
> @h-2 I agree with your solution, especially because you can require certain versions with pkg-config, so users can always say they want >=2.2.0 but <2.3.0 for instance. 

Good.

> The only thing you will have to keep in mind once you start developing seqan-3.0.0 is that you will have to rename all #include <seqan/foo/bar> to #include <seqan3/foo/bar>, including all consumers of seqan.

seqan3 will likely include major restructuring anyway, so this will be the smallest problem ;)

SoapZA:
```
diff --git a/util/cmake/SeqAnBuildSystem.cmake b/util/cmake/SeqAnBuildSystem.cmake
index 595890e..96a4d27 100644
--- a/util/cmake/SeqAnBuildSystem.cmake
+++ b/util/cmake/SeqAnBuildSystem.cmake
@@ -201,8 +201,8 @@ macro (seqan_build_system_init)
                  DESTINATION share/doc/seqan)
         # Install pkg-config file, except on Windows.
         if (NOT CMAKE_SYSTEM_NAME MATCHES Windows)
-            configure_file("util/pkgconfig/seqan.pc.in" "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION}.pc" @ONLY)
-            install(FILES "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION}.pc" DESTINATION share/pkgconfig)
+            configure_file("util/pkgconfig/seqan.pc.in" "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION_MAJOR}.pc" @ONLY)
+            install(FILES "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION_MAJOR}.pc" DESTINATION share/pkgconfig)
         endif (NOT CMAKE_SYSTEM_NAME MATCHES Windows)
     endif ()
```

esiragusa:
Hey I obviously meant to use a path like ```/usr/include/seqan/2.1.0/include/seqan```... no include has to be changed.

As for 2.1 vs 2.1.0, I was a bit lazy so I used ```SEQAN_VERSION``` instead of pasting major & minor variables. Nonetheless, in some weird situations one might want to have 2.1.0 and 2.1.1... you decide.

esiragusa:
> I therefore propose:
> * for seqan-2.*.* we use seqan-2.pc and ${prefix}/include/seqan as include path
> * for seqan-3.*.* we use seqan-3.pc and ${prefix}/include/seqan3 as include path

:-1: because we need at least minor versions (e.g. 2.1) and include/seqan3 is inconvenient and ugly too...

SoapZA:
You probably mean `/usr/include/seqan/2.1.0/seqan`; the second include is redundant. Again, you guys have to decide how you want to proceed, this is not my call. My personal preference ranks:
1. `${prefix}/include/seqan-${MAJOR}.${MINOR}/seqan` (my suggestion, identical to the GNOME/GTK guidelines https://developer.gnome.org/programming-guidelines/unstable/parallel-installability.html.en#solution)
2. `${prefix}/include/seqan/${MAJOR}.${MINOR}.${PATCH}/seqan` (Enrico)
3. `${prefix}/include/seqan` for seqan-2.x.y and `${prefix}/include/seqan3` for seqan-3.x.y (Hannes)

esiragusa:
Yep I meant ```/usr/include/seqan/2.1.0/seqan```. I am for major + minor too, but that might be too limiting in some cases, isn't it?

SoapZA:
This is like GCC before version 5. You could install GCC 4.7.4 and 4.8.5 in parallel, but installing GCC 4.8.4 and 4.8.5 in parallel requires a lot of tricks (exactly for the same reason, as GCC installs in a major + minor fashion, without patchlevel). I agree, there are borderline cases where installing two patchlevel versions is desirable (if testing for regressions etc.), but I think this is so specialised that no distribution/end-consumer would want to do it. Another point to consider: the .pc file should **never** contain the patchlevel version, as then maintaining external tools becomes a nightmare, as the build system requires constant updating purely for bugfixes. So if the .pc file doesn't include the patchlevel, then installing `/usr/include/seqan/2.1.0` and `/usr/include/seqan/2.1.1` will make one of them not accessible to the user, as seqan-2.1.pc will only yield `/usr/include/seqan/2.1.1`, such that using 2.1.0 requires setting compiler includes manually. If you have to set compiler includes manually, you might as well just install in a totally different prefix.

The reason I prefer `${prefix}/include/seqan-${MAJOR}.${MINOR}/seqan` is because it starts a new directory tree without mucking around with the other trees. In addition, it is not as deep as your suggestion. Starting a whole new tree outside of `/usr/include/seqan/` allows for partial coexistence with seqan versions before 2.1.

PR #1279, opened 2015-11-12
Gustaf can load FastA files with Iupac characters

ktrappe:


esiragusa:
Failure of test_test_bam_io is unrelated.

Issue #1272, opened 2015-11-05
bam io does not support custom tags with floats


h-2:
i.e. this does not work:
```c++
appendTagValue(bamR.tags, "ZE", mIt->eValue, 'f');
```
There are multiple overloads for integers:
https://github.com/seqan/seqan/blob/master/include/seqan/bam_io/bam_tags_dict.h#L675

and for sequences:
https://github.com/seqan/seqan/blob/master/include/seqan/bam_io/bam_tags_dict.h#L691

(same is true for `extractValue()`).

I am not completely sure how the internals work,  @esiragusa can you have a look?

esiragusa:
Is eValue a float? See https://github.com/seqan/seqan/blob/master/include/seqan/bam_io/bam_tags_dict.h#L637

esiragusa:
Or should https://github.com/seqan/seqan/blob/master/include/seqan/bam_io/bam_tags_dict.h#L675 be enabled for ```NumberConcept```?

h-2:
fixed by #1285

PR #1285, opened 2015-11-13
[FIXED-#1272] change integerconcept to numericalconcept

h-2:
This could be made a hotfix, but since bam_io has diverged between master and develop this would become unclean. Since it hasn't been used yet anyway, I propose to just merge into develop.

### DIFF
--- a/include/seqan/bam_io/bam_tags_dict.h
+++ b/include/seqan/bam_io/bam_tags_dict.h
@@ -506,7 +506,7 @@ struct ExtractTagValueHelper_
 };
 
 template <typename TResultValue, typename TId>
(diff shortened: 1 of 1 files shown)


esiragusa:
Do you know what diverged? I am not aware of any big changes in BAM I/O...

h-2:
there are some c++11 optimizations. I tried all ways, but I cannot create a PR than applies cleanly against both branches. Since 2.0.2 and 2.1.0 will be released at the same time and no one else ever noticed this bug, I don't think its a problem to just apply to develop.

h-2:
All is green. Can I merge?

PR #1286, opened 2015-11-13
Fix Intel C++ Compiler support

h-2:
This preliminary support for ` icc>=16`.

If this doesn't break anything else, please merge asap, as I want to activate nightly-builds, so that we know easily which things need fixing.

I will then start fixing things.

Most tests I tried actually work already with the exception of the blast module.


include/seqan/platform/platform_gcc.h
```diff
@@ -39,9 +39,20 @@
...
+// GNU COMPATIBLE
 #ifndef PLATFORM_GCC
 #define PLATFORM_GCC
 
+#if !defined(__clang__) && !defined(__ICC) && !defined(PLATFORM_GNU)
+// ACTUALLY GNU
+#define PLATFORM_GNU
+#endif
+
+#if defined(__clang__) && !defined(__ICC) && !defined(PLATFORM_CLANG)
```
esiragusa: ???

include/seqan/platform/platform_gcc.h
```diff
@@ -39,9 +39,20 @@
  * @signature #define PLATFORM_GCC
  */
 
+// GNU COMPATIBLE
 #ifndef PLATFORM_GCC
 #define PLATFORM_GCC
 
+#if !defined(__clang__) && !defined(__ICC) && !defined(PLATFORM_GNU)
```
esiragusa: ???

esiragusa:
This is a hotfix...

include/seqan/platform/platform_gcc.h
h-2: We have set up our system so that PLATFORM_GCC implies "GCC or compatible" which is confusing, but apps (and possibly external people) depend on it. This variable explicitly tells us if the compiler is GCC. This is the recommend way:
http://nadeausoftware.com/articles/2012/10/c_c_tip_how_detect_compiler_name_and_version_using_compiler_predefined_macros

h-2:
> This is a hotfix...

Well, no, because Intel is not officially supported for the 2.0.* series, so there is nothing requiring us to add support. There would be branch-spedific changes (BLAST module, already in this commit), possibly our stl-adaptations (not sure, but I have local changes)... so its not worth it from my POV.

include/seqan/platform/platform_gcc.h
esiragusa: ```#if defined(__clang__)``` suffices to detect clang

include/seqan/platform/platform_gcc.h
esiragusa: According to that article, this (PLATFORM_GNU) is ```defined(__GNUC__) || defined(__GNUG__)```

include/seqan/platform/platform_icc.h
```diff
@@ -59,19 +64,15 @@
 #define finline __inline__
 
 // default 64bit type
-#define __int64 int64_t    // nolint
```
esiragusa: Are all these signed types defined in icc?

esiragusa:
This PR contains only hotfixes... watch your diff line by line.

include/seqan/platform/platform_gcc.h
h-2: True, but I thought it would make more sense that we use our own DEFs throughtout. We also define PLATFORM_WINDOWS_VS, although detecting MSC_VER would suffice.

include/seqan/platform/platform_gcc.h
h-2: That check is already done before including the header platform_gcc.h 
So inside the header we just need to make sure its nothing else that real gcc.

include/seqan/platform/platform_icc.h
h-2: Yes, and not removing them breaks all SeqAn builds.

h-2:
> This PR contains only hotfixes... watch your diff line by line.

But it has changes to the blast module which is not present in master.

h-2:
> > This PR contains only hotfixes... watch your diff line by line.
> But it has changes to the blast module which is not present in master.

I have tried seperating this out, but there is other divergence between master and develop in the platform files which can't be resolved cleanly without other major changes, I therefore strongly suggest we only merge these changes into develop. Seeing that ICC was not supported before, this effects no existing users, I therefore call it a feature :)

Issue #973, opened 2015-04-21
StringSet is missing insert 


rrahn:
The insert interface according to the StringConcept (http://docs.seqan.de/seqan/develop/concept_StringConcept.html#StringConcept%23insert) is not implemented for StringSets.

h-2:
Also add removeValue and removeValueById

PR #1297, opened 2015-11-30
[Feature] StringSet: add insertValue, insert, replace 

smehringer:
Fixes: #973


h-2:
@smehringer what is the status here? Which tests still fail, what needs to be done?

PR #1313, opened 2016-01-11
[FIX] Addition to StringSet: add insertValue, insert, replace #1297

smehringer:
readded test of String<short, Packed<> > on a ConcatDirect-Stringset. Test does not fail anymore because the double conversion to TString was removed.


rrahn:
lgtm!

PR #1316, opened 2016-01-11
various fixes and improvements to the build system

h-2:
- [x] generate and install pkgconfig files

supersedes: #1075, #1278 

The decision on this happened today:
  * we begin with seqan-2.pc for the 2.1.* and 2.2.* series. No other changes are being made, because we expect no interface breakage for 2.1->2.2, i.e. after the next release we plan for semantic versioning

@SoapZA can you check whether this patch works properly for you? Thanks!

  - [x] install FindSeqAn.cmake to the correct location when doing a library install

  - [x] add "make dox" to the default make target in library release mode


SoapZA:
Hi @h-2,
thanks for the PR, I'll check it in the evening. Have you considered includedir versioning too? i.e. something along ``${prefix}/include/seqan-${SEQAN_MAJOR_VERSION}/seqan`` for the headers? Versioning the .pc file is good, but that only prevents interface breakage, it does not allow for coexistence.

In addition, while I personally do not use cmake, and versioning is handled for the .pc file, you might want to consider versioning **FindSeqAn.cmake** as well. Just a suggestion, and not as important, as pkg-config files are the canonical way on Unix-like systems.

h-2:
>  Have you considered includedir versioning too?

SeqAn2.* will use `include/seqan` and anything after that will use something different, e.g. `include/seqan3`

> you might want to consider versioning FindSeqAn.cmake as well

true, I will have a look at it.

h-2:
@rrahn this is good from my POV. If you agree, please merge.

rrahn:
Ok looks good.

PR #1323, opened 2016-01-12
Correction of CIGAR example

### LINKED ISSUE #1309: Documentation error with CIGAR format in Pairwise Sequence Alignment Tutorial
In this tutorial, and in particular in the Assignment 5, the author considers "S" operations in CIGAR means substitutions. This is not true.
In accordance to SAM format specifications (http://samtools.github.io/hts-specs/SAMv1.pdf), "S" operations should mean soft-clipping, which is widely used with local aligners like bowtie --local, or even BLAST. The "M" operation means both match and mismatch.
This is not the first time I see problems with soft-clipping in Seqan, which is not correctly implemented in the Fragment Store for example (see https://github.com/seqan/seqan/issues/535). I dont have the technical abilities to debug this pb, but as a user I can tell you that this is not an unfrequent pb. In local alignments you can have up to 80% of soft-clipped sequences in your sam file, which means that we cannot treat it with high level structures in seqan.

xenigmax:
According to the definition, S stands for "soft-clipping" and X stands for "mismatch".
So "S" in the CIGAR string in ParwiseSequenceAlignment should be changed to "X".

This will fix #1309 .

### DIFF
--- a/manual/source/Tutorial/PairwiseSequenceAlignment.rst
+++ b/manual/source/Tutorial/PairwiseSequenceAlignment.rst
@@ -593,7 +593,7 @@ Assignment 5
 
           ref: AC--GTCATTT
           r01: ACGTCTCA---
-          Cigar of r01: 2M2I1S3M3D
+          Cigar of r01: 2M2I1X3M3D
 
     Solution (Step 1)
       .. container:: foldable


rrahn:
:+1:

PR #1328, opened 2016-01-14
[FIX] detect c++11 for demos on master, too

h-2:
we only had this in develop. It could cause a situation where some demos failed on gcc-5, where the compilers default defines lead to SEQAN_CXX11_STANDARD being defined, but our custom CXX11_CXX_FLAGS where not set in time, so some demos *thought* they had c++11, while they did not.
I am not sure why we did not see this on cdash, maybe due to differences in the build config, or because something changed between gcc-5.2 and gcc-5.3

### DIFF
--- a/util/cmake/SeqAnBuildSystem.cmake
+++ b/util/cmake/SeqAnBuildSystem.cmake
@@ -604,6 +604,7 @@ macro (seqan_build_demos_develop PREFIX)
     # Find SeqAn with all dependencies.
     set (SEQAN_FIND_DEPENDENCIES ALL)
     find_package (SeqAn REQUIRED)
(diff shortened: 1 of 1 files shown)


rrahn:
So this is already changed in develop?
Could you please verify that develop can be merged into master without problems?
Last time we had a lot of troubles because of fixes added asynchronously to develop and master.

h-2:
> So this is already changed in develop?

yes

> Could you please verify that develop can be merged into master without problems?
> Last time we had a lot of troubles because of fixes added asynchronously to develop and master.

Indeed, this breaks it. I will try to resolve this nicely somehow.

Issue #706, opened 2014-11-23
SeqAn libraries installation


sagitter:
Hi all.

mason2 and seqan_flexbar compilation provides two private libraries named libseqan_flexlib.so and libmason_sim.so . I think they must be present in same directory of binary files in order to be used.

If I install both libraries in a private seqan sub-directory, mason and seqan_flexbar tests fail.
Can I link the binaries to the libraries if located in a different path?

esiragusa:
Does setting ```BUILD_SHARED_LIBS=OFF``` by default solve this issue?

esiragusa:
Or, stated differently, why do we need ```BUILD_SHARED_LIBS=ON```? We have only app-specific shared objects...

h-2:
Should be fixed by #1329 
If not, please reopen.

PR #1327, opened 2016-01-14
[FEATURE] more CI platforms and fix of current bugs

h-2:
there are currently issues in travis. Moving to the trusty images hopefully will fix it. Also I have added more compilers and osx.

PR #1329, opened 2016-01-14
[FIX] make static builds fully static

h-2:
This makes our binaries static, except on OS X where this is not possible in general. It also changes the behaviour to use libc++ on OSX as there is no reason to use libstdc++ anymore by default (libstdc++ is solder now).

Fixes: #706 and also things like 
https://lists.fu-berlin.de/pipermail/seqan-dev/2015-December/msg00000.html


h-2:
apparently clang in combination with libstdc++ and on linux cannot build static binaries "just like that". I will have a look at it tomorrow.

h-2:
the build issue is a bug in clang-3.4. Since we are only targetting clang 3.5, 3.6 and 3.7 with seqan 2.1.0 this can be ignored.
Waiting for #1327 to see if this is fixed with newer clang.

h-2:
@rrahn seems like our solution worked! It only has the queue_example warnings left. Can you tell once those are solved? Then I update this pull and trigger a rerun of the tests, just to make sure!

rrahn:
@h2 Nice. 
I tell you when I have the patch ready.

rrahn:
@h-2 please have a look at #1356

h-2:
@rrahn all green, please merge

rrahn:
:+1:

esiragusa:
**Could you please disable static linking by default?** Static linking introduces problems with std::thread. I had to spend quite some time debugging Yara to discover that this change triggered a segmentation fault in std::thread when compiled in Debug mode on Linux g++ 4.9... For a description and a possible solution see http://stackoverflow.com/questions/23499656/segmentation-fault-in-stdthreadids-stdoperator. Cheers :beers:

h-2:
@esiragusa can you tell me how to reproduce your problem? Is it covered by your app tests? 

If it only appears in debug mode, than I will deactivate static linking for debug mode. I can also look at the  solution described above, i.e. `-Wl,--whole-archive -lpthread -Wl,--no-whole-archive`.

However I am strongly against deactivating static linking for our shipped binaries. I know it is "not recommended", but the same way distributing stand-alone linux binaries is "not recommend". Especially since many of our apps now link against libgomp we need static linking, because we have no way of knowing if the installed lib provides the correct symbols. We have had reports about this. On Mac using GCC and static linking is the only way to get OpenMP at all...

h-2:
> @esiragusa can you tell me how to reproduce your problem? Is it covered by your app tests?

On both Linux and FreeBSD, each with g++-4.9 and Debug your app tests pass... so please provide instructions on how to reproduce the problem!

esiragusa:
I think dynamic linking should be used by default because that's what's everybody expects - it's the norm. Apart from the problem that I am encountering, binary sizes increase dramatically with static linking. For instance, the size of yara_mapper increases from 2.9 to 5.3 MB.

If you want to release statically-linked binaries for seqan apps so to avoid glibc versioning issues then you can enable static linking via a cmake flag like ```-DSEQAN_STATIC_BUILDS```.

Yara app tests pass because they run only with one thread as OpenMP is (was?) not available on clang. Anyway, the set of reads is so small that no prefetching would happen. To reproduce the issue, compile yara_mapper in Debug mode and invoke it as: ```yara_mapper -t 2 -vv <index> <single-end-reads>``` providing at least 200k reads.

I reproduced this issue on Ubuntu 16.04 Xenial when compiling in Debug mode with g++ 4.9.3. The user that wrote me might have compiled in Release mode as this is the default. You can try the posted solution for static linking but in any case I would like to have dynamic linking by default.

h-2:
> I think dynamic linking should be used by default because that's what's everybody expects - it's the norm. 

It is only the norm for binaries installed through package management systems, because these systems ensure that the right shared libraries are present. Whenever software is distributed independently, you should make it static unless you know precisely where you are deploying. You can have a look at bowtie, usearch, diamond... they are all built statically.

>  For instance, the size of yara_mapper increases from 2.9 to 5.3 MB

I am sorry, but I think if you target your software at managing GBs of Data on systems with 10s - 100s GBs of RAM then a 2MB increase in your binary size is not an issue from my point of view.

> you can enable static linking via a cmake flag like -DSEQAN_STATIC_BUILDS.

I will introduce this flag, but if its up to me than it will be on by default in our package builds. This PR was discussed and merged weeks ago, it fixes issues and we are in the middle of releasing right now, so I don't want to start a new policy discussion.

I will however try to get your fix in, but I need some more details, can you provide me with a link to example files (either locally on one of our servers or fetchable)?
Like I said, we are quite busy with last-minute issues, also a yara build breakage on 32bit btw :wink:

[on an unrelated sidenote: some guides recommend static linking for performance reasons and the intel compiler automatically does so at highest optimization level]

esiragusa:
I think that one wants to dynamically link when downloading sources and compiling binaries for himself. Obviously, none of the versioning problems you mentioned can occur when one compiles and runs on the same system. Conversely, static linking adds problems like the one I reported.

What I proposed above is that you enable ```-DSEQAN_STATIC_BUILDS``` when you package app builds to be released. This flag should be off by default.

As for performance reasons, you should benchmark that to see if there is any difference for our apps.

h-2:
> What I proposed above is that you enable `-DSEQAN_STATIC_BUILDS` when you package app builds to be released. This flag should be off by default.

That's ok for me, but then we should have some weekly jobs building packages (which would be nice anyway) to spot problems on a regular basis.

esiragusa:
I know... packaging of single apps is untested too... :disappointed:

h-2:
@esiragusa can you still provide me with some files so I can include the fixes for pthread?

esiragusa:
You can take any reference (even a fasta with 10bp sequence) and any fastq file containing at least 200k reads. The segfault happens immediately (as soon as the reads get loaded).

h-2:
@esiragusa I cannot reproduce your observations with this file ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data/HG00096/sequence_read/SRR062634.filt.fastq.gz
which has more that 300K sequences. If you would like me to investigate and have this fixed for the static yara packages, please provide me with a file.

esiragusa:
I used dm3 from ```ftp://igenome:G3nom3s4u@ussd-ftp.illumina.com/Drosophila_melanogaster/UCSC/dm3/Drosophila_melanogaster_UCSC_dm3.tar.gz``` as reference and SRR1658525_1.fastq.gz as single-end reads (converted from SRA). Anyway I would spare you the time - the problem is unrelated to the data...

PR #1333, opened 2016-01-15
fix detection of omp for clang-3.7.1

h-2:
see #1332 
[develop branch]


h-2:
please hold, it seems the upstream clang37 bug that is triggered by seqan is not fixed, yet, our test_parallel still crashes clang during build :(

h-2:
it seems like the clang omp fix won't be backported to 3.7.* so I will instead adapt the warning

PR #1356, opened 2016-01-21
[FIX] Fixes gcc-5 warning of uninitialised value.

rrahn:
Removed forced move of temporary object in return statement and let the compiler optimise.


h-2:
@rrahn can you make a PR from this branch into master also? Thanks!

Issue #1161, opened 2015-09-01
Alignment algorithms accept unsupported scoring schemes


esiragusa:
Right now it is possible to call a restricted alignment algorithm with an unsupported general scoring scheme. The result is then undefined.

For instance, one can do the following:
```globalAlignment(align, Score<int, Simple>(3, -2, 0, -5), Hirschberg());```

We should:
* If possible, restrict the scoring scheme type at compile time;
* Otherwise, assert the scoring scheme parameters with ```SEQAN_ASSERT```.

rrahn:
Fixed by #1361

PR #1361, opened 2016-01-23
[FIX] Asserts wrong use of scoring scheme for Hirschberg algorithm.

rrahn:
Requires scoring scheme Simple for Hirschberg algorithm and asserts if affine gap costs are used.


rrahn:
Solves issue #1161.

h-2:
Don't our other DP algorithms use the linear space feature of hirschberg anyway? Why do we need to specify "Hirschberg" explicitly?

rrahn:
No. Hirschberg is a special algorithm using linear space for the traceback as well. This is not the default case for the other dp algorithms. Because it works only with linear gaps for now.

h-2:
Ah, ok!

PR #1352, opened 2016-01-21
Compiler detection

marehr:
See ticket #1349

PR #1365, opened 2016-01-25
Use cmake3.2 in travis builds.

marehr:
This feature is needed for pull request #1352.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       addons:
         apt:
           sources: ['ubuntu-toolchain-r-test', 'george-edison55-precise-backports']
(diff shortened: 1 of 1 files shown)


h-2:
Does this install multiple cmake versions in travis, or does it update the installed one? Because if the former is the case, you might want to create some symbolic links in the script as well...
Otherwise I will merge it!

marehr:
According to the [build log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/104592687/log.txt) it replaces it.

```
[...]
Preparing to replace cmake-data 2.8.7-0ubuntu5 (using .../cmake-data_3.2.3-5ppa5~precise1_all.deb) ...
Unpacking replacement cmake-data ...
Preparing to replace cmake 2.8.7-0ubuntu5 (using .../cmake_3.2.3-5ppa5~precise1_amd64.deb) ...
Unpacking replacement cmake ...
[...]
```

PR #1366, opened 2016-01-25
fix URIEncoding problem for entities with ::

### LINKED ISSUE #831: The escaping of html file names in the dox system seems to be odd

xenigmax:
This will fix #831 .

Check how it works : 
http://docs.seqan.de/seqan/master_version_test_832/?p=FragmentStore::PredefinedAnnotationTypes

### DIFF
--- a/util/py_lib/seqan/dox/tpl/js/common.js
+++ b/util/py_lib/seqan/dox/tpl/js/common.js
@@ -35,7 +35,6 @@
                         hash = '::' + tmp.split('::')[1];
                         console.log('p == ' + p + ' -- hash = ' + hash);
                     }
-
                     if(window.lookup.hasOwnProperty(p+hash)) { // eg. "enum_*.html"
                         redirectTo = window.lookup[p+hash] + '.html'
                     }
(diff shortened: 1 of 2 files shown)


util/py_lib/seqan/dox/tpl/js/common.js
```diff
@@ -35,7 +35,11 @@
                         hash = '::' + tmp.split('::')[1];
                         console.log('p == ' + p + ' -- hash = ' + hash);
                     }
-                    if(window.lookup.hasOwnProperty(p)) {
+
```
rrahn: Please remove the empty line.

rrahn:
LGTM just the minor nail polishing thing and I'll merge.

PR #1376, opened 2016-01-26
[FIX] Replace spaces with underscores (ICC related)

temehi:
This accounts for the changes made by #1296. 
Teh new ${SEQAN_DATE} is delimited by underscore instead of space. The change is made to accommodate this.


h-2:
have you verified that this works? Do you not have to escape the literal underscore?

temehi:
Sure it works! I don't think underscores need escaping.

PR #1369, opened 2016-01-25
Require and fully use C++11

marehr:
Resolves #1320

PR #1378, opened 2016-01-27
Require C++11 and remove lots of dead code

h-2:
supersedes #1369 

@esiragusa we have noticed a bug in a test of the modifier_mod_pos (which we fixed). This hints though that there was some unexpected copying due to broken constructors before the full switch to c++11. I don't think anything in yara relies on modifying a mod_pos string and yara used c++11 anyway and the yara test passes, but just so that you know...

@rrahn if the tests pass this can be merged from my POV.


esiragusa:
@h-2 thanks for the info. yara uses c++11. did the bug show up in the c++03 test?

By the way, on develop yara doesn't compile anymore - I tried on larix. There is a linking problem with libpthreads. If I checkout yara-v0.9.3 it compiles. I am wondering why this problem doesn't show up in the nightlies...

h-2:
@esiragusa Yes, the problem was that the test expected wrong behaviour , because the C++03 behaviour was wrong.
For me yara builds fine on develop, both FreeBSD and Linux, please write me an email if it doenst work...

PR #1379, opened 2016-01-27
[FIXED-#903] Fixes compiler warnings for simd vector.

### LINKED ISSUE #903: Fix SIMD Vector
#892 introduced compile errors on platforms which are not under CI - see e.g. http://cdash.seqan.de/viewBuildError.php?buildid=123258 or http://cdash.seqan.de/viewBuildError.php?buildid=123260

@weese do you know what could it be?

rrahn:
Adds workaround for missing function ```_mm_cvtsi64_si128``` on 32 bit arch.
@esiragusa if you could have  a quick look, that would be gerat.

Fixes #903.

### DIFF
--- a/include/seqan/basic/basic_simd_vector.h
+++ b/include/seqan/basic/basic_simd_vector.h
@@ -409,11 +409,17 @@ _shuffleVector(TSimdVector1 const &vector, TSimdVector2 const &indices, SimdPara
         reinterpret_cast<const __m128i &>(vector),
         reinterpret_cast<const __m128i &>(indices)));
 }
+
 template <typename TSimdVector1, typename TSimdVector2>
 inline TSimdVector1
 _shuffleVector(TSimdVector1 const &vector, TSimdVector2 const &indices, SimdParams_<16, 8>, SimdParams_<8, 8>)
 {
+#if defined(SEQAN_IS_32_BIT)
+    __m128i idx = _mm_slli_epi16(_mm_unpacklo_epi32(_mm_cvtsi32_si128(reinterpret_cast<const __uint32 &>(indices)),
+                                                    _mm_cvtsi32_si128(reinterpret_cast<const __uint64 &>(indices) >> 32)), 1);
+#else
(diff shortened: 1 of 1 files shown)


esiragusa:
The old code loads a 64 bit word w into 128bit vector and shifts left:
```
_mm_cvtsi64_si128(w)
r0 := w
r1 := 0

_mm_slli_epi16(..., 1)
r0 << 1
```

Now:
```
_mm_cvtsi32_si128(w)
r0 := w
r1 := r2 := r3 := 0

_mm_cvtsi32_si128(w >> 32)
r4 := w >> 32 = 0
r5 := r6 := r7 := 0

_mm_unpacklo_epi32(..., ...)
r0 := w ; r1 := 0
r2 := 0 ; r3 := 0

_mm_slli_epi16(...,1)
r0 << 1
```

seems to do the same stuff... but the extra shift, ```_mm_cvtsi32_si128``` and ```_mm_unpacklo_epi32``` just moves zeros around, or? :scream:

rrahn:
Well, in fact ```w``` in our case is a SimdVector of size 8, right?
Since ```_mm_cvtsi64_si128``` seems to be missing on 32-bit arch, I used ```_mm_cvtsi32_si128``` to achieve the same the thing.

So your example would look like:
```
a = _mm_cvtsi32_si128(reinterpret_cast<const __int32 &>(vec))
r0 := vec[31:0]
r1 := r2:= r3 := 0

b = _mm_cvtsi32_si128(reinterpret_cast<const __int64 &>(vec) >> 32)
r0 := vec[63:32]
r1 := r2:= r3 := 0

_mm_unpacklo_epi32(a, b)
r0 := vec[31:0]
r1 := vec[63:32]
r2 := r3 := 0
```
which is the same as the old version:
```
_mm_cvtsi64_si128(vec)
r0 := vec[31:0]
r1 := vec[63:32]
r2 := r3 := 0
```

esiragusa:
Ah ok I assumed w to be a 32 bit word :+1:

Issue #565, opened 2014-10-04
Create test for reading both UCSC file for store annotations


holtgrewe:
Issue #277 was not detected since there is no test for loading both the known genes and isoforms to the FragmentStore. We should have such a test.

rrahn:
@temehi ping

temehi:
@rrahn you can close this issue now

PR #1383, opened 2016-01-28
A test for loading both Genes and Isoforms UCSC files in to a store

temehi:
A `UcscRecord` should be able to handle records from both Genes and Isoforms UCSC files. This added test makes sure that we can load correctly both types of files into a single store.

This fixes issue #565

### DIFF
--- a/tests/store/test_store_io.h
+++ b/tests/store/test_store_io.h
@@ -162,6 +162,167 @@ SEQAN_DEFINE_TEST(test_store_io_read_ucsc_known_genes)
     SEQAN_ASSERT(atEnd(it));
 }
 
+SEQAN_DEFINE_TEST(test_store_io_read_ucsc_known_genes_and_isoforms)
+{
+    // The file contains 13 annotations in total which will be checked line
+    // after line.
+    seqan::CharString ucscGenesPath = SEQAN_PATH_TO_ROOT();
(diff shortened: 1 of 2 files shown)


tests/store/test_store_io.h
```diff
@@ -162,6 +162,167 @@ SEQAN_DEFINE_TEST(test_store_io_read_ucsc_known_genes)
     SEQAN_ASSERT(atEnd(it));
 }
 
+SEQAN_DEFINE_TEST(test_store_io_read_ucsc_known_genes_and_isoforms)
+{
+    // The file contains 13 annotations in total which will be checked line
+    // after line.
+    seqan::CharString ucscGenesPath = SEQAN_PATH_TO_ROOT();
+    append(ucscGenesPath, "/tests/store/example_knownGene.txt");
```
rrahn: please use ```getAbsolutePath``` instead of SEQAN_PATH_TO_ROOT().
See here the documentation: http://docs.seqan.de/seqan/develop/

temehi:
@rrahn right now all the tests use SEQAN_PATH_TO_ROOT() instead of `getAbsolutePath`. We changed all the demos to  `getAbsolutePath` last time. I suggest to do the same for all the tests. For now It's better if you close this issue and we will deal with the `getAbsolutePath` in the future for consistency.

rrahn:
Ok maybe we do it like this.

PR #1386, opened 2016-01-29
[FIX] clang wrongly detected as GCC

h-2:


h-2:
fix the warnings we have on current clang

PR #1389, opened 2016-01-29
[FIX] fixes wrong file extensions eg .fasta.bam

temehi:
- Removed wrong extensions like .fasta.bam
- Added bgzf compression for all file types.
- Done by reusing the .bgzf extension for BAM files

This fixes issue #1037

### DIFF
--- a/include/seqan/bam_io/read_bam.h
+++ b/include/seqan/bam_io/read_bam.h
@@ -97,6 +97,7 @@ readHeader(BamHeader & header,
            Bam const & /*tag*/)
 {
     clear(header);
+    FileExtensions<BgzfFile>::VALUE[0] = ".bam";
     // Read BAM magic string.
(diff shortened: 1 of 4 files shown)


include/seqan/bam_io/read_bam.h
```diff
@@ -97,6 +97,7 @@ readHeader(BamHeader & header,
            Bam const & /*tag*/)
 {
     clear(header);
+    FileExtensions<BgzfFile>::VALUE[0] = ".bam";
     // Read BAM magic string.
```
h-2: Do you actually edit the static variable here? This is a hack at best. It will break other things, e.g. you won't be able to read a .bgzf file after that. It would also produce all sorts of problems in a multi-threaded context.

Please find a different solution :pray: 

Thanks!

temehi:
Okay, okay... I will try another way

PR #1393, opened 2016-01-29
[FIX] fixes wrong file extensions eg .fasta.bam …

temehi:
- Removed wrong extensions like .fasta.bam
 - Added bgzf compression for all file types.
 - Done by setting primaryExtensionOnly to true when looking for compression extensions
 - This fixes issue #1037


h-2:
Ah, that's a smart way of solving the problem :) It only works so long as our compression formats only have one extension each, but since that's the case, lets get for it :)

PR #1395, opened 2016-01-30
More compiler error/warning fixes for VS.

rrahn:
Fixes:
* warning about integrateAlign
* diable simd_vector tests on windows platforms and fixed #warning issue
* Fixes conversion error in file_mapping.h
* Fixes test for atomic primitives.


h-2:
LGTM (only bam_io errors <- see it affects linux, too!)

Issue #1372, opened 2016-01-25
blast module on windows


h-2:
The blast module triggers multiple bugs on windows:
http://blogs.msdn.com/b/vcblog/archive/2015/06/02/constexpr-complete-for-vs-2015-rtm-c-11-compiler-c-17-stl.aspx?CommentPosted=true&PageIndex=5#comments

  - [x] deactivate blast module for 2.1.0 on Windows [ in #1396 ]
  - [ ] reevaluate after next update to Visual Studio; or
  - [ ] reevaluate at 2.1.2 if to replace the blast code with C++98 code

h-2:
wait for #1385 and see if we need only block it on MSVC

h-2:
do first [ ]  and then move to 2.1.1

h-2:
@cpockrandt Could you check if this works with the new compiler version? I also changed something to make the blast module build with icc, maybe that made it work on VS, too.

h-2:
This works with ICC on Windows and Clang on Windows (due to me merged by @marehr) which is good enough for now.

PR #1396, opened 2016-01-31
[INTERNAL] deactivate blast module on windows again

h-2:
see #1372


h-2:
@rrahn if the tests pass, can you merge this before tonight, so we see it in the nightlies? Thanks!

PR #1414, opened 2016-02-03
Dox issues about page redirection

xenigmax:
This will fix below 3 situations.

1. "::" in p
Before: http://docs.seqan.de/seqan/master/?p=ArgParseArgument::ArgumentType
(not working in window system)
After: http://docs.seqan.de/seqan/master_test_r/?p=ArgParseArgument::ArgumentType

2. multiple "#"s
Before : http://docs.seqan.de/seqan/master/?p=BamTagsDict#BamTagsDict%3A%3ABamTagsDict
After : http://docs.seqan.de/seqan/master_test_r/?p=BamTagsDict#BamTagsDict%3A%3ABamTagsDict

3. "_" in the middle of the file name
Before : http://docs.seqan.de/seqan/master/?p=SEQAN_HOST
After : http://docs.seqan.de/seqan/master_test_r/?p=SEQAN_HOST

### DIFF
--- a/util/py_lib/seqan/dox/tpl/js/common.js
+++ b/util/py_lib/seqan/dox/tpl/js/common.js
@@ -25,31 +25,38 @@
         if(window != window.parent && window.name == 'list') {
             try {
                 var redirectTo = null;
-                var hash = $.urlHash(window.parent.location);
-                if($.urlParam('p', window.parent.location)) {
-                    var p = $.urlParam('p', window.parent.location).split('/')[0];
-                    if (p.indexOf('::') != -1)
-                    {
-                        var tmp = p;
-                        p = tmp.split('::')[0];
-                        hash = '::' + tmp.split('::')[1];
-                        console.log('p == ' + p + ' -- hash = ' + hash);
-                    }
-                    if(window.lookup.hasOwnProperty(p+hash)) { // eg. "enum_*.html"
-                        redirectTo = window.lookup[p+hash] + '.html';
-                    }
-                    else if(window.lookup.hasOwnProperty(p)) { // valid for entries in "class_*.html"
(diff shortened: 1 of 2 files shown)


util/py_lib/seqan/dox/tpl/js/common.js
```diff
@@ -25,31 +25,51 @@
...
-                    } else {
-                        $(window.parent['main'].document).find('#content').prepend('<div class="open-in-frame alert alert-danger">Could not find page for <strong>' + p + '</strong></div>'); 
-                        // TODO: start search using search form for p
-                    }
+                var p = null;
+                var hash = null;
+ 
+                // from parameter p
+                if($.urlParam('p',window.parent.location))
+                    p = decodeURIComponent($.urlParam('p',window.parent.location).split('/')[0]);
```
rrahn: Why first checking if the function does not return null.
Could we not directly call ```p = decodeURIComponent($.urlParam('p',window.parent.location).split('/')[0]);``` in line 28?
The same for hash?

rrahn:
Ok apart from the minor issue this lgtm. :+1:

util/py_lib/seqan/dox/tpl/js/common.js
xenigmax: Hmm.. I'm not an expert.
However, if `$.urlParam('p',window.parent.location)` returns null, `null.split('/')[0]` looks unsafe to me.
So I changed my code like below.
Just feel free to let me know your opinion!

```
28         var p = decodeURIComponent($.urlParam('p',window.parent.location)); // from parameter p
29         var hash = decodeURIComponent($.urlHash(window.parent.location)); // from hash (start with #)
30 
31         if(p != null) {
32             p = p.split('/')[0];
```

PR #1418, opened 2016-02-03
[FIX] flexbar: Workaround for MSVC bug.

rrahn:
There is a nasty bug in the Visual Studio c++ compiler, that causes the compiler to crash when using _pragma clause wrapped by macros in a nested way. 
This happens to be a problem with our ```SEQAN_OMP_PRAGMA``` statement.
This workaround, although not beautiful, fixes this. However, seqan_flexbar is deleted in develop, so we won't support it with 2.1 anyway. So merging this into develop is not necessary.


h-2:
Won't this cause warnings on all non-omp-platforms (clang)? I thought that was the reason for the construct...

rrahn:
There is a #if defined(_OPENMP) ...  #else ... #endif construct.
If it is not defined it falls back to the same version as if there was no OpenMP specified. So the code actually does this what our macros would do in a more easy fashion. 
But as described above, this doesn't work with MSVC.

h-2:
Ah, ok, I will merge it.

PR #1419, opened 2016-02-03
Fixes VS 14 builds.

rrahn:
Since we already fixed these errors in develop, I cherrypicked the commits and will create a separate PR into develop, which needs some manual conflict solving.


apps/ngs_roi/project_interval.cpp
```diff
@@ -237,7 +237,7 @@ void IntersectBed::finishContig()
     if (options.verbosity >= 2)
         std::cerr << "finishContig()\n";
 
-    while (!empty(bedRecords))
+    while (!bedRecords.empty())
     {
```
h-2: Why this change? Is the global function broken? If yes, we should fix the global function...

rrahn: Apparently the global empty function for std::list is not overloaded.
So one has to make another PR for the missing empty function for 2.0.2.
Note that this is gone in develop by the restructuring of the adaption of the STL containers.
So this issue goes not back into develop.
Anyway, since here always the member function is preferred over the global overload, I think it is just conventional style to also use the empty member function.

rrahn: And it solves the problem :smile:

PR #1422, opened 2016-02-04
Fix VS bug: missing '>' before identifier 'int64_t'

marehr:
Fixes a VS bug in bit_twiddling, see [build log](http://cdash.seqan.de/viewBuildError.php?buildid=163211):

> C:\seqan-nightly\co-git-develop\seqan-Nightly-64\include\seqan/misc/bit_twiddling.h(639): error C2146: syntax error: missing '>' before identifier 'int64_t'

-----
And removes _popCountImplGeneric, because it was only introduced for backward compatibility as stated in the following comment

>  The functions below are implemented as follows.  _popCountImplGeneric() is used if there are no intrinsics provided by the compiler (the case for Visual C++ 2008)

This [commit](https://github.com/seqan/seqan/pull/1369/files#diff-6274dcb7db964b2f07ae413d59f8bba3L339) removed the last use of the function.


h-2:
@rrahn can this be merged, mr visualstudio?

Issue #1074, opened 2015-08-06
PackedString relies on allocated memory to be zero-initialized


catkira:
I experienced strange behaviour and test_sequence unit test failing on visual studio 2015. I could isolate the problem to uninitialized memory that PackedString uses somehow.
The problems in visual studio 2015 disappeared when I added 
     memset((void*)data, 0, count * sizeof(TValue));
in include/seqan/basic/allocator_interface.h
Furthermore I could reproduce the failing unit test in GCC and Clang by doing
     memset((void*)data, 0xFFFFFFFF, count * sizeof(TValue));

I think actually PackedString is supposed not to rely on zero-initialized memory. However the details are quite complex, so that I think its best that somebody who knows this code better than me fixes this. For now I add this memset 0 workaround (could be a little performance penalty though,if the allocator function is called very often)

esiragusa:
Great job! :+1:

If I recall correctly, during the SeqAn 2.0.0 release, I saw something related in Valgrind reports...

esiragusa:
@catkira did you solve this issue? :sweat_smile:

catkira:
@SRR1175698.5 AdapterXX length=51
GGGGGCTGAAGATCGGAAGAGCACACGTCTGGATCCACGACGCTCTTCCGA
+SRR1175698.5 SEB9BZKS1:84:D1F78ACXX:4:1101:2363:2237 length=51
CCCFFFFFHHHHHHIIHIIIJHHIIIFHGGFEFF>FGGGGIEEIDHAHGHC
@SRR1175698.1 Adapter1 length=51
NCTGACTGAGATCGGAAGAGCACACGTCCGTCTGAACTCCAGTCACTTAAA
+SRR1175698.1 SEB9BZKS1:84:D1F78ACXX:4:1101:1969:2170 length=51
#1=DDFFFHHGHHJJHIIIIHIIJJJIJJIIJJIJJJJGHIJJJJJJJGHF
@SRR1175698.2 Adapter2 length=51
NGGGTCTGACATTTTGGAGGCACATCTGCGATCGGAAGAGCACACGTCTGAAA
+SRR1175698.2 SEB9BZKS1:84:D1F78ACXX:4:1101:2216:2128 length=51
#1=DADEFGHHHHJJJJHIIGIJJIJJJIJJJDIJJJJGHIIIJIIIFGHI
@SRR1175698.3 Adapter3 length=51
[...]

PR #1425, opened 2016-02-04
[INTERNAL] revert to old packed_string implementation

h-2:
Please see #1074 for the problem description.

I have tried to track down this issue, but I could not get satisfactory results and could also not provide a workaround. This problem goes back way before seqan-2.0 and the current implementation was apparently **always broken**.
This PR reverts our implementation to the one we still had in `packed_string_old.h`. This has a few features less (no binary operators for `String<bool, Packed<> >`...) but apparently nothing was using this anyway and it wasn't documented. I don't know if there are performance differences between the two implementations, but ultimately correctness is more important than speed.

We can reassess this for 2.2 or 3.0.

### DIFF
--- a/include/seqan/sequence.h
+++ b/include/seqan/sequence.h
@@ -107,7 +107,8 @@
 #include <seqan/sequence/string_alloc.h>
 #include <seqan/sequence/string_cstyle.h>
 #include <seqan/sequence/string_block.h>
-#include <seqan/sequence/string_packed.h>
+#include <seqan/sequence/string_packed_old.h>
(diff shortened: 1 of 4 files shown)


esiragusa:
@h-2 @rrahn Nooooo please don't do that! :scream:

rrahn:
@h-2 Sorry, but apparently the binary functions are used heavily. :confounded:  
So I close this PR. We should rather focus on solving the issue. 
Please see the discussion on #1074 for more details that I will post.

PR #1429, opened 2016-02-05
[WORKAROUND] fix insert() in packed_string

h-2:
* resize() without value is unchanged
  * more efficient host-based resize() with value added
  * insert() changed to where clearUnusedBits() is called before the real insert to fix the bugs we where seeing with #1074 

There is probably a larger problem hidden within ClearSpaceStringPacked_ or even the memcpy functions, but the problem only appeared with inserts and they have a workaround now, so everything is usable and regular resizes are not affected (initializing resizes should even be faster now).

PLEASE NOTE that some valgrind warnings for unintialized reads are unrelated to this and simply part of the implementation, i.e. seqan resizes() dont initialize and if only the first n bits are read and written, this implicitly read/writes the other bits although there value is irrelevant to the implementation.  This could be worked around (to silence the warnings) but this can be done for one of the next releases.

### DIFF
--- a/include/seqan/sequence/string_packed.h
+++ b/include/seqan/sequence/string_packed.h
@@ -37,6 +37,8 @@
 #ifndef SEQAN_SEQUENCE_STRING_PACKED_H_
 #define SEQAN_SEQUENCE_STRING_PACKED_H_
 
+#include <bitset>
+
 namespace seqan {
 
 // ============================================================================
@@ -583,14 +585,10 @@ resize(
         clear(host(me));
         return 0;
     }
-
-    TStringSize _new_length = new_length;
-    TStringSize max_length = (resize(host(me), TTraits::toHostLength(_new_length) + 1, tag) - 1) * TTraits::VALUES_PER_HOST_VALUE;
-    if (_new_length > max_length)
-        _new_length = max_length;
-    front(host(me)).i = _new_length;
-
-    return _new_length;
+    TStringSize max_length = (resize(host(me), TTraits::toHostLength(new_length) + 1, tag) - 1) * TTraits::VALUES_PER_HOST_VALUE;
+    if ((TStringSize)new_length > max_length)
+        new_length = max_length;
+    return front(host(me)).i = new_length;
 }
 
 template <typename TValue, typename THostspec, typename TSize, typename TExpand, typename TValue2>
@@ -603,36 +601,46 @@ resize(String<TValue, Packed<THostspec> > & me,
     typedef String<TValue, Packed<THostspec> > TString;
     typedef PackedTraits_<TString> TTraits;
     typedef typename Size<TString>::Type TStringSize;
+    typedef typename TTraits::THostValue THostValue;
 
-    TStringSize _new_length = new_length;
     TStringSize old_length = length(me);
+    TStringSize old_host_length = length(host(me));
 
-    if (_new_length < length(me))
-        return resize(me, _new_length, tag);
+    if (static_cast<TStringSize>(new_length) < length(me))
+        return resize(me, new_length, tag);
 
     // create WORD
-    String<TValue, Packed<THostspec> > tmp;
-    resize(tmp, static_cast<unsigned>(TTraits::VALUES_PER_HOST_VALUE), Exact());
-    for (unsigned i = 0; i < TTraits::VALUES_PER_HOST_VALUE; ++i)
-        tmp[i] = newValue;
+    THostValue tmp;
+    tmp.i = 0ul;
+    __uint64 ord = ordValue(TValue(newValue));
+    for (unsigned j = 0; j < TTraits::VALUES_PER_HOST_VALUE; ++j)
+        tmp.i |= ord << (j * TTraits::BITS_PER_VALUE);
 
     // fill up host string word-wise
     TStringSize max_length = (resize(host(me),
-                                     TTraits::toHostLength(_new_length) + 1,
-                                     getValue(host(tmp), 1),
+                                     TTraits::toHostLength(new_length) + 1,
+                                     tmp,
                                      tag) - 1)
                              * TTraits::VALUES_PER_HOST_VALUE;
 
     // set values on formerly last word, because not affected by above
-    TStringSize trailing_values = TTraits::VALUES_PER_HOST_VALUE - (old_length % TTraits::VALUES_PER_HOST_VALUE);
-    for (TStringSize i = 0; i < trailing_values; ++i)
-        me[old_length + i] = newValue;
+    if (old_host_length > 1)
+    {
+        TStringSize alreadySet = old_length % TTraits::VALUES_PER_HOST_VALUE;
 
-    if (_new_length > max_length)
-        _new_length = max_length;
-    front(host(me)).i = _new_length;
+        // clear non-set positions (which may be uninitialized ( != 0 )
+        tmp.i = (~0ul << (64 - alreadySet * TTraits::BITS_PER_VALUE)) >> TTraits::WASTED_BITS;
+        host(me)[old_host_length-1].i &= tmp.i;
 
-    return _new_length;
+        for (TStringSize i = alreadySet; i < TTraits::VALUES_PER_HOST_VALUE; ++i)
+            host(me)[old_host_length-1].i |=  ord << ((TTraits::VALUES_PER_HOST_VALUE - i - 1) * TTraits::BITS_PER_VALUE);
+    }
+
+    if (static_cast<TStringSize>(new_length) > max_length)
+        new_length = max_length;
+    front(host(me)).i = new_length;
+
+    return new_length;
 }
 
 // --------------------------------------------------------------------------
--- a/include/seqan/sequence/string_packed.h
+++ b/include/seqan/sequence/string_packed.h
@@ -582,7 +582,33 @@ resize(
     TStringSize max_length = (resize(host(me), TTraits::toHostLength(new_length) + 1, tag) - 1) * TTraits::VALUES_PER_HOST_VALUE;
     if ((TStringSize)new_length > max_length)
         new_length = max_length;
-    return front(host(me)).i = new_length;
+    front(host(me)).i = new_length;
+
+    // NOTE(h-2): this fixes valgrind issues and insertion errors, otherwise the string
+    //            is dependent on getting 0 initialized memory; however it is unclear
+    //            whether this fixes all the problems and it does incur a performance
+    //            penalty when explicitly resizing the string
+    _clearUnusedBits(me);
+    return new_length;
+}
+
+template <typename TValue, typename THostspec, typename TSize, typename TExpand, typename TValue2>
+inline typename Size< String<TValue, Packed<THostspec> > >::Type
+resize(String<TValue, Packed<THostspec> > & me,
+       TSize new_length,
+       TValue2 const & newValue,
+       Tag<TExpand> tag)
+{
+//     typedef typename Size< String<TValue, Packed<THostspec> > >::Type             TSize;
+    typedef typename Iterator<String<TValue, Packed<THostspec> >, Standard>::Type TIter;
+
+    TSize oldLen = length(me);
+    TSize ret = resize(me, new_length, tag);
+
+    if (ret >= oldLen)
+        for (TIter it = iter(me, oldLen, Standard()), itEnd = end(me, Standard()); it != itEnd; ++it)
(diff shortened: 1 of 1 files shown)


include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
         clear(host(me));
         return 0;
     }
-    TStringSize max_length = (resize(host(me), TTraits::toHostLength(new_length) + 1, tag) - 1) * TTraits::VALUES_PER_HOST_VALUE;
-    if ((TStringSize)new_length > max_length)
-        new_length = max_length;
-    return front(host(me)).i = new_length;
+
+    TStringSize _new_length = new_length;
```
rrahn: Why the extra copy?

include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
...
+    TStringSize max_length = (resize(host(me), TTraits::toHostLength(_new_length) + 1, tag) - 1) * TTraits::VALUES_PER_HOST_VALUE;
+    if (_new_length > max_length)
```
rrahn: When is this true? max_length is the maximum number of values the packed string can hold which should always be greater or equal than the actual length of the packed string, shouldn't it?

include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
...
+    // create WORD
+    String<TValue, Packed<THostspec> > tmp;
+    resize(tmp, static_cast<unsigned>(TTraits::VALUES_PER_HOST_VALUE), Exact());
+    for (unsigned i = 0; i < TTraits::VALUES_PER_HOST_VALUE; ++i)
+        tmp[i] = newValue;
+
+    // fill up host string word-wise
+    TStringSize max_length = (resize(host(me),
+                                     TTraits::toHostLength(_new_length) + 1,
+                                     getValue(host(tmp), 1),
```
rrahn: getValue is deprecated!
And I would prefer not using a tmp packed string, but a tmp host value which can be filled with bit shifting.
The values have to be shifted from left.
As far as I understand, the packed tuple does not yet overload the shift operators. So just use the .i member instead.
As far as I can tell, the class is not designed to easily overload the HostValueType from the PackedString, right?
So that shouldn't cause any trouble for now.

include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
...
+resize(String<TValue, Packed<THostspec> > & me,
+       TSize new_length,
+       TValue2 const & newValue,
+       Tag<TExpand> tag)
+{
+    typedef String<TValue, Packed<THostspec> > TString;
+    typedef PackedTraits_<TString> TTraits;
+    typedef typename Size<TString>::Type TStringSize;
+
+    TStringSize _new_length = new_length;
```
rrahn: Why copying the new_length?

include/seqan/sequence/string_packed.h
h-2: There was a problem with using the enum value in the function below, so the cast did not work (but assignment did). With the last change this could be removed again...

include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
...
+                                     getValue(host(tmp), 1),
+                                     tag) - 1)
+                             * TTraits::VALUES_PER_HOST_VALUE;
+
+    // set values on formerly last word, because not affected by above
+    TStringSize trailing_values = TTraits::VALUES_PER_HOST_VALUE - (old_length % TTraits::VALUES_PER_HOST_VALUE);
+    for (TStringSize i = 0; i < trailing_values; ++i)
+        me[old_length + i] = newValue;
+
+    if (_new_length > max_length)
```
rrahn: Again why this check?

include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
...
+    TStringSize max_length = (resize(host(me),
+                                     TTraits::toHostLength(_new_length) + 1,
+                                     getValue(host(tmp), 1),
+                                     tag) - 1)
+                             * TTraits::VALUES_PER_HOST_VALUE;
+
+    // set values on formerly last word, because not affected by above
+    TStringSize trailing_values = TTraits::VALUES_PER_HOST_VALUE - (old_length % TTraits::VALUES_PER_HOST_VALUE);
+    for (TStringSize i = 0; i < trailing_values; ++i)
+        me[old_length + i] = newValue;
```
rrahn: Again just mask the word at the corresponding positions in the host value. 
Using subscript operator creates a certain overhead.

include/seqan/sequence/string_packed.h
h-2: I don't know why it was done. I thought maybe we have packed fixed size arrays as well that can't be resized past their max size? All of the functions in this file always check this. That's the whole point of resize() returning a value or not?

include/seqan/sequence/string_packed.h
```diff
@@ -579,10 +583,74 @@ resize(
...
+inline void
+insert(String<TValue, Packed<THostSpec> > & me,
+       TPosition pos,
+       TSeq const & insertSeq,
+       Tag<TExpand>)
+{
+    // NOTE(h-2): something is wrong (probably in ClearSpaceStringPacked_ below)
+    //            which requires us to _clearUnusedBits on inserts()
+    //            [this has a moderate performance penalty]
+    _clearUnusedBits(me);
```
rrahn: Ok, so we defer the linear setting of unused bits until an insert occurs. 
I think this will be ok for now as a workaround, because one should avoid calling insert on packed strings very often anyway.

include/seqan/sequence/string_packed.h
h-2: > getValue is deprecated!

I thought only value() was deprecated. But it doesn't matter for this place.

> And I would prefer not using a tmp packed string, but a tmp host value which can be filled with bit shifting.

I wanted to that, too, but couldn't figure out an easy way to do it. If I shift in the ordValue() it will shift more than the required bits, won't it?

include/seqan/sequence/string_packed.h
h-2: see above [can be removed]

include/seqan/sequence/string_packed.h
h-2: again: see above :)

include/seqan/sequence/string_packed.h
rrahn: Ah ok, so if the host is not resizable. Then this makes sense.

include/seqan/sequence/string_packed.h
rrahn: Why not using ```word = (word >> BitsPerValue<TValue>::VALUE) | MASK```, where ``` MASK = static_cast<TWord>(ordValue(value)) << (BitsPerValue<TWord>::VALUE - 1 - BitsPerValue<TValue>::VALUE)```
I just saw, that the bit shift operators are implemented for bit compressed tuples, so you can use directly the tuple itself.

h-2:
I think I addressed all your issues, please review again.

include/seqan/sequence/string_packed.h
```diff
@@ -585,6 +589,76 @@ resize(
...
+
+    TStringSize old_length = length(me);
+    TStringSize old_host_length = length(host(me));
+
+    if (static_cast<TStringSize>(new_length) < length(me))
+        return resize(me, new_length, tag);
+
+    // create WORD
+    THostValue tmp;
+    tmp.i = 0ul;
```
rrahn: 0ul could cause problems on win64 platforms, couldn't it?
I mean unsigned long is only 32 bit on 64bit windows machines?

include/seqan/sequence/string_packed.h
```diff
@@ -585,6 +589,76 @@ resize(
...
+    __uint64 ord = ordValue(TValue(newValue));
```
rrahn: I thought we made quite some effort to replace this with uint64_t?

include/seqan/sequence/string_packed.h
h-2: true, I will just go with 0

include/seqan/sequence/string_packed.h
h-2: this is a PR against master and develop, so I have to go with something that both understand. I can make an extra PR after that if you want to :smile:

include/seqan/sequence/string_packed.h
```diff
@@ -585,6 +589,76 @@ resize(
...
+                                     tag) - 1)
+                             * TTraits::VALUES_PER_HOST_VALUE;
+
+    // set values on formerly last word, because not affected by above
+    if (old_host_length > 1)
+    {
+        TStringSize alreadySet = old_length % TTraits::VALUES_PER_HOST_VALUE;
+
+        // clear non-set positions (which may be uninitialized ( != 0 )
+        tmp.i = (~0ul << (64 - alreadySet * TTraits::BITS_PER_VALUE)) >> TTraits::WASTED_BITS;
```
rrahn: here 0ul should be rather explicitly casted into the word type.

include/seqan/sequence/string_packed.h
```diff
@@ -585,6 +589,76 @@ resize(
...
+
+    if (static_cast<TStringSize>(new_length) < length(me))
+        return resize(me, new_length, tag);
+
+    // create WORD
+    THostValue tmp;
+    tmp.i = 0ul;
+    __uint64 ord = ordValue(TValue(newValue));
+    for (unsigned j = 0; j < TTraits::VALUES_PER_HOST_VALUE; ++j)
+        tmp.i |= ord << (j * TTraits::BITS_PER_VALUE);
```
rrahn: I think you need to shift the bits from left to right, don't you?
AFAIK, the order is from MSB to LSB.

include/seqan/sequence/string_packed.h
```diff
@@ -585,6 +589,76 @@ resize(
...
+    if (old_host_length > 1)
+    {
+        TStringSize alreadySet = old_length % TTraits::VALUES_PER_HOST_VALUE;
+
+        // clear non-set positions (which may be uninitialized ( != 0 )
+        tmp.i = (~0ul << (64 - alreadySet * TTraits::BITS_PER_VALUE)) >> TTraits::WASTED_BITS;
+        host(me)[old_host_length-1].i &= tmp.i;
+
+        for (TStringSize i = alreadySet; i < TTraits::VALUES_PER_HOST_VALUE; ++i)
+            host(me)[old_host_length-1].i |=  ord << ((TTraits::VALUES_PER_HOST_VALUE - i - 1) * TTraits::BITS_PER_VALUE);
```
rrahn: Again, please check if the order is not the other way around.
Might as well that I am wrong, but I think I remember it to be from MSB to LSB.

rrahn: You could as well use setValue/assignValue of the the bit compressed tuple if you know the position within the word :wink: See here: https://github.com/seqan/seqan/blob/master/include/seqan/basic/tuple_bit_compressed.h#L246
Then you don't have to play around with the correct bit shifts.
That would be easier and safer.

include/seqan/sequence/string_packed.h
h-2: believe me, I have sat a long time on this until it worked, if only a little thing is wrong all oft the tests fail and I have manually compared the entire bit-registers with the version from before that goes through the iterator. This is the right way to do it :sunglasses:

include/seqan/sequence/string_packed.h
h-2: assignValue only works when setting individual bits, but the whole point of these optimizations was to save operations, so it is better to do it this way.
[ord is already a pattern that is shifted in]

include/seqan/sequence/string_packed.h
h-2: yes

include/seqan/sequence/string_packed.h
h-2: > I think I remember it to be from MSB to LSB.

It is `[ WASTED_BITS | FIRST_VALUE | SECOND_VALUE | ... ]` and inside the values it is MSB to LSB that is correct. Bus since I am shifting in an entire pattern in this is the way to go.

h-2:
@rrahn the last commit should address all changes.

PR #1431, opened 2016-02-06
tree-view addition in dox system

### LINKED ISSUE #496: Browse documentation by module
I think the documentation would profit greatly from being browsable by module, e.g. when nothing is entered in the search bar, show a tree-like view on the left that has all the modules; upen clicking on a module the list should be expanded to show the contained dox items, sorted by type.

reasons:
our code and our doc is spread between many files, and while the new interface stuff draws some connections between items, it is still not clear what is where.

people include seqan functionality by including a header file, so this is the most intuitive way to structure functionality from my pov.

xenigmax:
http://docs.seqan.de/seqan/master_test_496/

It was not an easy task.. We should learn HTML, CSS, Javascript(jQuery), Bootstrap, and Python to maintain this system (at least). I have spent about 4 full days for this. :sob:  However, I'm still not confident about this documentation system. **So you should review cafefully before merging this PR.**

If you think we need more work, I suggest you to postpone this feature to the next release.
I'm not sure whether I can do more in the next week. :sob:

this will fix #496


xenigmax:
All right. Test was failed. I didn't know that there is a test for dox. I will check it!

h-2:
Wow, this is really cool! Thanks first of all first of all! 

There are some minor displaying issues on Firefox here and altogether the interface is now very "full". I see two ways one could improve this:
  1. when one clicks on a module, hide all other modules in the tree and instead display a "← back to Module list"; or
  1. move the module list into a extra drop-down menu, either on the bottom beside the version selector, make a drop-down menu below the search bar that reads "Browse by module".

Just my thoughts! @seqan/all what to you think?

h-2:
@xenigmax can you fix the test? what do you think about my comments?

xenigmax:
@h-2 
1) about the test
I will try virtualenv as you said.  The failure will be reproduced if the problem is about Python.

2) about the suggestions.
I think your ideas are good :) If there is no more opinion about this, I will do this as you suggested.

xenigmax:
This should not be merged yet.

xenigmax:
I'm not sure about the UI / UX.  
But I made a PR according to your suggestions on last Monday.

1) Smaller font-size 
2) The module tree will not be shown at the first time. 
3) Hide the tree after searching 
4) The scrolling issue around version selector is resolved.

Demo page:
http://docs.seqan.de/seqan/master_test_496/

xenigmax:
@rrahn @h-2 
I have applied your suggestions so far. 
http://docs.seqan.de/seqan/master_test_496/

Maybe we can talk about other issues(if any) on next Monday.

h-2:
@seqan/seqan-rw I like it, what does everyone else think?

temehi:
looks great. One small thing, it would be awesome if the main window changes automatically when you click on a leaf node eg.  align/align_base. Right now one has to click again on the left side pane to see the contents

jtroehr:
hey Hannes, looks also great to me.

rrahn:
@xenigmax nice! thanks!

xenigmax:
1) @temehi  's suggestion was resolved in the discussion.
2) A Firefox specific layout & javascript issues were fixed.
3) Hide module-tree on keyword input.

PR #1432, opened 2016-02-06
tree-view addition in dox system

### LINKED ISSUE #496: Browse documentation by module

xenigmax:
http://docs.seqan.de/seqan/master_test_496/

It was not an easy task.. We should learn HTML, CSS, Javascript(jQuery), Bootstrap, and Python to maintain this system (at least). I have spent about 4 full days for this. However, I'm still not confident about this documentation system. So you should review carefully before merging this PR.

If you think we need more work, I suggest you postpone this feature to the next release.
I'm not sure whether I can do more in the next week. 

this will close #496


xenigmax:
@rrahn @h-2 
I did the test on my PC and it works fine. 
In other words, I can't reproduce this failure with my PC.

What I did was below.

From /manual/CMakeLists.txt
```
 23 add_test (build_manual
 24           ${SPHINX_EXECUTABLE} -W -n
 25           -b html -d ${CMAKE_BINARY_DIR}/manual/doctrees
 26           ${CMAKE_CURRENT_SOURCE_DIR}/source
 27           ${CMAKE_BINARY_DIR}/manual/html)
```

How can I reproduce it?

h-2:
@xenigmax can you try with a fresh virtualenv? Maybe you are also using a different python version?

xenigmax:
@cpockrandt 
Below was the suggestion from @h-2  in #1431 .

> when one clicks on a module, hide all other modules in the tree and instead display a "← back to Module list"; or

To do this exactly, I have to remove unselected nodes in the tree. However,  the current version of bootstrap-treeview does not support this. It's on the list of the next release.
https://github.com/jonmiles/bootstrap-treeview/issues/50

So, in this PR, I hide the entire tree right after selection of an item and show "back to the module list".
UI-related enhancement can be discussed in the near future again.

cpockrandt:
Are you still working on it? According to the last travis build, ``build_manual`` is still failing.

xenigmax:
It's moved to 3.0 by @rrahn  and I will check the broken test later.

PR #1446, opened 2016-02-10
Added Changelog for Release 2.0.2

cpockrandt:


h-2:
There is now also official FreeBSD support. Can you add that to the Platforms stuff?

Also for the docs, I would not write `moved cpp code blocks into seperate files`, but instead `all code snippets in the documentation system now undergo build tests and continuous integration to ensure that they do not get outdated.` or something like that → the effect for the user should be clear.

At `fixed compiler-warnings in Visual Studio` I would add `( /W2 produces no warnings with SeqAn, now)`

Issue #1172, opened 2015-09-15
KNIME YaraIndexer node fails


esiragusa:
Description
------------------------------
KNIME YaraIndexer node produces a "Configure failed (ArrayIndexOutOfBoundsException): 0" error when connected to an input. The problem seems related to the IPrefixURIPortObject output.

How to reproduce the issue
------------------------------------
* Make the prepare_workflow_plugins target in SeqAn and use the output "workflow_plugin_dir" with the GenericKnimeNodes tools to produce the SeqAn KNIME nodes and input them in KNIME.
* Then connect YaraIndexer (Community Nodes->SeqAn->Read Mapping->YaraIndexer) to an input node (Community Nodes->GenericKnimeNodes->IO->Input File) with any fasta sequence.
* The error "Configure failed (ArrayIndexOutOfBoundsException): 0".

Explanation 
---------------

The latest Yara versions are tagged from develop, while KNIME builds use old master version. This fact should explain the issue.

rrahn:
@temehi is this still a problem?

esiragusa:
@rrahn this issue will be solved when our KNIME nodes will be built for 2.1.0 (see the explanation above). You can close the issue now or perhaps wait until it's solved...

rrahn:
@esiragusa Ah I see. Thanks for the update.

temehi:
@esiragusa @rrahn I was looking more into it. Sorry to say this but I don't think the YaraIndexer KNIME node of the new release  2.1.0 will work! The problem is related to the fact that the indexer doesn't output a single known file_type. The GenericKnimeNodes has a way to deal output-prefixes but yet it expects a common ending/extension for all the resulting files. Which is not the case for YaraIndexer.

rrahn:
@temehi don't be sorry. Fix it! :stuck_out_tongue_winking_eye:
What do you mean with common? Can we not just put some ending to it like: file.index something?
I mean I don't know which files are output by the indexer at the end, but it would be ok to just give it some extension, or am I overseeing something here?

h-2:
Our indexes consist of a bunch of files with all sorts of endings. This a bad design anyway, and in the future we should find some way to stuff the files into one, like `.tar`. Then we could arbitrarily compress this, as well. But for the short term we need a different soluation.
@temehi can we not hack around this some way? This effects all SeqAn tools that save indexes to disk, also Lambda (which I haven't had time to put into KNIME, but plan to do so in the future).

esiragusa:
[...]

PR #1451, opened 2016-02-10
[FIX] Fixes the CTD support for the prefix file types #1172

temehi:
- Added new mimetypes for yara_indexer
- Adds the "any" extension (*) in case of prefix types (input/output)
- This fixes the issue #1172


temehi:
@rrahn @h-2 @esiragusa can you guys merge this ASAP I need it for my next work.

h-2:
@temehi I think it would be better to have a generic SA-Index or FM-Index type, not a "Yara"-type, as the endings are not specific to yara.
Altogether I still do not know why KNIME needs to know about this, though. None of the calling apps need to see these extensions, you always just use the base filename as argument, so I think it is completely unrelated to CTD, or not?

temehi:
@h-2 do other indexers produce this extensions as well? If yes, I will make them an index type. that is not a problem. 

> so I think it is completely unrelated to CTD, or not? 

in order to connect two nodes you need to make sure that the output of one node is compatible to the input type of the second. in the case of input/output prefixes we need to make evrything compatable with everything... So NO it is related to CTD.

esiragusa:
Looks good to me. However Yara's index adds two files .rid.concat and .rid.limits to the FMIndex files. These files store reference ids (sequence names).

temehi:
@esiragusa you mean these two extensions are specific to yara? @h-2 do you have extensions specific to Lambda as well?

esiragusa:
@temehi yes

h-2:
>  @h-2 do you have extensions specific to Lambda as well?

Yes, I do. They depend on selected options, though. Is that a problem?

temehi:
Not at all! all we have to do is register them. But we can do this latter when LAMBDA is KNIME ready. lol.. you know it's a big commitment?. .lol

temehi:
@h-2 ping!  Is there still a problem not to merge this PR?

Issue #912, opened 2015-03-19
Replace "SEQAN_NAMESPACE_MAIN" with "seqan"


esiragusa:
**To avoid future merge conflicts, do it just before a major release.**

h-2:
fixed by #1454

PR #1454, opened 2016-02-12
[INTERNAL] replaced SEQAN_NAMESPACE_MAIN with seqan #912

cpockrandt:
solves #912


h-2:
@cpockrandt Could you also check for files that now have
```c++
namespace seqan
{
...
}

namespace seqan
{
...
}
```
And join these into one? Should be a nice regex ;-) At least you could scan for files that contain two invocations...

PS: You do not have your git-email setup correctly...

PR #1444, opened 2016-02-10
[FIX] build yara build warnings

h-2:
@esiragusa your PR fixed the fails but introduced warnings:
http://cdash.seqan.de/viewBuildError.php?type=1&buildid=164104
so I have renewed my PR. Can you greenlight this?

PR #1455, opened 2016-02-13
Yara: fix static_cast regression

esiragusa:
@h-2 fyi #1444 broke yara... please merge this one.

### DIFF
--- a/apps/yara/bits_matches.h
+++ b/apps/yara/bits_matches.h
@@ -821,10 +821,10 @@ findProperMates(TMatches const & mates, TMatch const & match,
     typedef typename Size<TReadSeqs>::Type          TReadId;
     typedef typename Value<TReadSeqs const>::Type   TReadSeq;
     typedef typename Size<TReadSeq>::Type           TReadSeqSize;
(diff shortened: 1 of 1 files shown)


esiragusa:
@h-2 @rrahn please merge asap

h-2:
@esiragusa Have you verified that this produces no warnings with GCC and Clang, on 32bit and 64bit, each?

esiragusa:
It's correct now, have a look at the types...

h-2:
You are not casting on the mateLength, are you sure this doesn't cause problems on 32bit, i.e. have you *tested* on clang and gcc with -m32 ? Because we have already "fixed" these two lines three times now, I would like this to be the final fix :wink:

esiragusa:
Why don't you add 32 bit tests on travis? :wink:

h-2:
> Why don't you add 32 bit tests on travis? :wink:

Thats actually a good idea for the future. For now: have you done the tests?

esiragusa:
Sure it works on my mac... can you please merge it? Thanks.

PR #1458, opened 2016-02-13

temehi:
- Added new mimetypes for index files and yara_indexer
 - Adds the "any" extension (*) in case of prefix types (input/output) if not provided
 - This fixes the issue #1172
 - Missing newline added from previous commit


h-2:
This is a different commit than 6f4f253741c56f5dfc5ccc1a199cc0dd690723c2

Please use git cherry pick

h-2:
Please re-open with cherry-picked commit.

temehi:
The correct commit is https://github.com/seqan/seqan/commit/09b5b75. But when I cherry-pick that commit on top of the current master I will get a different commit number with the same changes. Won't that be a problem? I mean master will be 1 commit ahead of develop or not?

PR #1459, opened 2016-02-13

temehi:
- Added new mimetypes for index files and yara_indexer
 - Adds the "any" extension (*) in case of prefix types (input/output) if not provided
 - This fixes the issue #1172
 - Missing newline added from previous commit


h-2:
The changes already are in develop, they only need to be pulled into master again.

PR #1474, opened 2016-02-16
[FIX] library package creation

h-2:


marehr:
Why is it assigned to me? What should I do? The fix itself seems okay to me, but the milestone seems to be wrong.

PR #1483, opened 2016-02-20
Yara v0.9.6

esiragusa:
* Fixed handling of reference metagenomes (references larger than 16k sequences)
* Enabled support for reference metagenomes by default (-DYARA_LARGE_CONTIGS=ON)
* Added option --sensitivity (low, high, full)
* Replaced option --output-secondary with --secondary-alignments (tag, record, omit)
* Renamed option --output-rabema as --rabema-alignment
* Renamed option --vverbose as --very-verbose
* Set default value of option --threads using detected number of cores
* Improved various options descriptions
* Removed useless BAM tag XT

PR #1489, opened 2016-02-23
Yara v0.9.6

esiragusa:
See #1483.


esiragusa:
Random fail of BAM I/O test...

PR #1493, opened 2016-02-24
remove side-effect of NDEBUG on SEQAN_ENABLE_DEBUG

marehr:
* Make ```SEQAN_ENABLE_DEBUG = 0``` default behavior
* Force-Set ```NDEBUG = !SEQAN_ENABLE_DEBUG``` (when false undefine NDEBUG)
* Update Documentation
* Update make script scaffolds (Do we even use them?)
* Remove ```SEQAN_ENABLE_DEBUG = 0``` in SeqAnBuildSystem.cmake, since it is now the default behavior

PR #1502, opened 2016-02-25
 remove side-effect of NDEBUG on SEQAN_ENABLE_DEBUG

marehr:
The same PR like the one for the master #1493


marehr:
Superseeded by #1504 (contains the changes)

PR #1504, opened 2016-02-25
update docs regarding ndebug

marehr:
See master PR #1503


h-2:
in the future: please do not mix different commits into different PRs to master/develop. It makes it hard to follow and increases the chance of introducing breakage between the branches.

marehr:
When reopen #1502 and merge it, the build passed.

The diff of this pr should be then smaller...

PR #1505, opened 2016-02-25
changes to cmake -DSEQAN_RELEASE_LIBRARY

marehr:
* don't ship apps and the manual on releases
* Introduce cmake -DSEQAN_OFFICIAL_PKGS=1 to build upstream releases with static binaries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,12 +132,17 @@ if (("${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE") OR
     add_subdirectory (demos)
 endif ()
 
-message (STATUS "Configuring apps")
-add_subdirectory (apps)
+if (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE_LIBRARY")
+    message (STATUS "Configuring apps")
+    add_subdirectory (apps)
+endif ()
 message (STATUS "Configuring dox")
 add_subdirectory (dox)
-message (STATUS "Configuring manual")
-add_subdirectory (manual)
+if ((NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE_LIBRARY") AND
+    (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE_APPS"))
+    message (STATUS "Configuring manual")
+    add_subdirectory (manual)
+endif ()
 message (STATUS "Configuring util/py_lib")
 add_subdirectory (util/py_lib)
 
--- a/util/cmake/SeqAnBuildSystem.cmake
+++ b/util/cmake/SeqAnBuildSystem.cmake
@@ -202,9 +202,31 @@ macro (seqan_build_system_init)
         endif (NOT CMAKE_SYSTEM_NAME MATCHES Windows)
     endif ()
 
-    if (("${SEQAN_BUILD_SYSTEM}" MATCHES "SEQAN_RELEASE") OR
-        ("${SEQAN_BUILD_SYSTEM}" MATCHES "APP"))
+    if ((CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64") OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64"))
+        set (SEQAN_64BIT_PROCESSOR 1)
+    endif()
+
+    if ((SEQAN_OFFICIAL_PKGS) AND
+        (NOT MSVC) AND
+        ((SEQAN_BUILD_SYSTEM MATCHES "SEQAN_RELEASE") OR
+        (SEQAN_BUILD_SYSTEM MATCHES "APP")))
+
         set (SEQAN_STATIC_APPS TRUE CACHE INTERNAL "Create static app binaries")
+        message (STATUS "Building static binaries")
+
+        if (SEQAN_64BIT_PROCESSOR)
+            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mmmx -msse -msse2 -msse3 -mssse3")
+            message (STATUS "Static binaries require SSE3.")
+        endif ()
+
+    endif ()
+
+    if ((SEQAN_BUILD_SYSTEM STREQUAL "DEVELOP") AND
+        (NOT MSVC) AND
+        (SEQAN_64BIT_PROCESSOR))
+
+        set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -march=native")
+        message (WARNING "Building CPU-optimized binaries that may not work elsewhere.")
     endif ()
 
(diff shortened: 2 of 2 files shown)


util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -202,9 +202,31 @@ macro (seqan_build_system_init)
...
+        (NOT MSVC) AND
+        ((SEQAN_BUILD_SYSTEM MATCHES "SEQAN_RELEASE") OR
+        (SEQAN_BUILD_SYSTEM MATCHES "APP")))
+
         set (SEQAN_STATIC_APPS TRUE CACHE INTERNAL "Create static app binaries")
+        message (STATUS "Building static binaries")
+
+        if (SEQAN_64BIT_PROCESSOR)
+            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mmmx -msse -msse2 -msse3 -mssse3")
+            message (STATUS "Static binaries require SSE3.")
```
h-2: Change to "Including optimizations for SSE3 in binaries."

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -202,9 +202,31 @@ macro (seqan_build_system_init)
...
+        set (SEQAN_64BIT_PROCESSOR 1)
+    endif()
+
+    if ((SEQAN_OFFICIAL_PKGS) AND
+        (NOT MSVC) AND
+        ((SEQAN_BUILD_SYSTEM MATCHES "SEQAN_RELEASE") OR
+        (SEQAN_BUILD_SYSTEM MATCHES "APP")))
+
         set (SEQAN_STATIC_APPS TRUE CACHE INTERNAL "Create static app binaries")
+        message (STATUS "Building static binaries")
```
h-2: include "." at the end :nail_care:

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -202,9 +202,31 @@ macro (seqan_build_system_init)
...
+        if (SEQAN_64BIT_PROCESSOR)
+            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mmmx -msse -msse2 -msse3 -mssse3")
+            message (STATUS "Static binaries require SSE3.")
+        endif ()
+
+    endif ()
+
+    if ((SEQAN_BUILD_SYSTEM STREQUAL "DEVELOP") AND
+        (NOT MSVC) AND
+        (SEQAN_64BIT_PROCESSOR))
```
h-2: You were right, we can remove the 64bit_processor here.

Instead add `(NOT SEQAN_NONATIVE)` so people at least have a way to deactivate this. [we might also want to on travis, if it increases build times too much]

h-2:
Thanks for the PR.

Can you please also verify that setting the SEQAN_CXX_FLAGS in the above location actually has the desired effect? Just to make sure that they are not overwritten at a later point, or that there are scope problems or sth like that (I still haven't fully understood how this works in cmake).

util/cmake/SeqAnBuildSystem.cmake
h-2: Should we also check CMAKE_BUILD_TYPE==Release? → do we want march-=native only in debug as well or is that bad for debug?

esiragusa: I would define ```SEQAN_NATIVE=OFF``` instead (like in [Yara](https://github.com/seqan/seqan/blob/master/apps/yara/CMakeLists.txt#L65)). That's more intuitive than double negation.

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -202,9 +202,31 @@ macro (seqan_build_system_init)
...
+
+        set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -march=native")
+        message (WARNING "Building CPU-optimized binaries that may not work elsewhere.")
```
esiragusa: Wow in this way we have a cmake warning by default...

esiragusa: I would write something like: "CPU-optimized binaries that may not work on other computers. If you plan to distribute binaries, call cmake with -DSEQAN_RELEASE... or with -DSEQAN_NATIVE=OFF."

CMakeLists.txt
```diff
@@ -132,12 +132,17 @@ if (("${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE") OR
...
+    message (STATUS "Configuring apps")
+    add_subdirectory (apps)
+endif ()
 message (STATUS "Configuring dox")
 add_subdirectory (dox)
-message (STATUS "Configuring manual")
-add_subdirectory (manual)
+if ((NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE_LIBRARY") AND
+    (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE_APPS"))
+    message (STATUS "Configuring manual")
```
h-2: This is not yet fully correct, as we don't want the manual either when we have APP:*

I propose
```cmake
if ("${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
    message (STATUS "Configuring manual")
    add_subdirectory (manual)
endif ()
```
and also for dox:
```cmake
if (("${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP") OR
    ("${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE_LIBRARY"))
    message (STATUS "Configuring dox")
    add_subdirectory (dox)
endif ()
```

CMakeLists.txt
```diff
@@ -132,12 +132,17 @@ if (("${SEQAN_BUILD_SYSTEM}" STREQUAL "SEQAN_RELEASE") OR
     add_subdirectory (demos)
```
h-2: remove first condition for demos. There is no SEQAN_RELEASE (anymore), only SEQAN_RELEASE_APPS and SEQAN_RELEASE_LIBRARY. The latter automatically gets the demos it rquires for dox.

esiragusa:
@h-2 ?

util/cmake/SeqAnBuildSystem.cmake
h-2: I agree, use STATUS instead of warning. I also agree with the more verbose message.

util/cmake/SeqAnBuildSystem.cmake
h-2: I don't agree. SEQAN_NATIVE is is not set by default anywhere, it is the equivalent of not being defined in C.
Actually changing something by defining it it off seems very unintuitive, to me at least.

`SEQAN_NO_NATIVE` very clearly implies that native is the default, and that defining this variable will change the default behaviour, so it is much better :crown:

PR #1520, opened 2016-03-02
[TUTORIAL: Metafunctions] reworked metafunction tutorial.

### LINKED ISSUE #1307: Documentation error in the Metafuction tutorial
In the Hint of [assignment 1](http://seqan.readthedocs.org/en/develop/Tutorial/Metafunctions.html#assignment-1):

> Use the Metafunctions Value to access the type of the elements in the container. Use the function value to assign the values.``

Firstly, [Value](http://docs.seqan.de/seqan/develop/?p=ContainerConcept#Value) is mentioned, but not used in the solution.

Secondly, [value](docs.seqan.de/seqan/develop/?p=RandomAccessContainerConcept#value) is marked as deprecated. Thus, the solution would be written as 

```c++
template<typename TCollection>
void swap(TCollection & items, int i, int j, int k) {
    TCollection tmps;
    resize(tmps, k);

    for(int x = 0; x < k; ++x) {
        tmps[x] = items[i + x];
    }

    for (int x = 0; x < k; ++x) {
        items[j + x] = items[i + x];
    }

    for (int x = 0; x < k; ++x) {
        items[i + x] = tmps[x];
    }
}
```

which uses no visible metafunctions and therefore the tutorial loses its purpose.

----------------

> If you want to search for metafunctions only you can do so by only selecting the metafunction category to the left of the search window at the online documentation.

This sentence makes no sense to me.

----------------

> The function is now quite generic allowing any container of type T. In addition we specify two positions that should be swapped (as integers which is not really generic, but it suffices for the demo) an the length of the swapped region. Now we can define a helper variable help, which can be of type T.

should be:
> demo) _and_ the length

smehringer:
Fixes #1307 

please review for spelling and grammar mistakes. I do not entirely trust my english skills.


manual/source/Tutorial/GettingStarted/Metafunctions.rst
```diff
@@ -97,10 +120,13 @@ For example the following function prints the length of a fixed sized string usi
 .. includefrags:: demos/tutorial/metafunctions/base.cpp
     :fragment: length
 
-SeqAn Metafunctions
--------------------
+.. includefrags:: demos/tutorial/metafunctions/base.cpp.stdout
+    :fragment: length
+
+.. important::
+      Redundant use of "*Value*":
```
h-2: s/Redundant use/Different uses/
;-)

manual/source/Tutorial/GettingStarted/Metafunctions.rst
```diff
@@ -97,10 +120,13 @@ For example the following function prints the length of a fixed sized string usi
...
 
-If you want to search for metafunctions only you can do so by only selecting the metafunction category to the left of the search window at the online documentation.
+      There is a **Type Metafunction** called :dox:`ContainerConcept#Value` (``Value<TSomeType>::Type``), which has nothing to do with the **Value Metafunctions** described above.
```
h-2: A little "colloquial". Maybe:
> Please note that :dox:`ContainerConcept#Value` (``Value<TSomeType>::Type``) is a **Type Metafunction**, because it returns a Type (e.g. of Values in a container) and not a value.

smehringer: first you use capital latters for Type and Value in
> .. returns a Type 

and 
> .. e.g. of Values

but than you switch to lowercase
> .. and not a value.

so just to be picky about consistency i wanted to ask if i might change the latter also to 'Value' ?

h-2: > first you use capital latters for Type and Value in

Well types typically begin with a capital letter in our c++ dialect, so I stook with ``Type``

> but than you switch to lowercase

Well value should be lowercase really I just thought of it as a type there, so I upperased it. Anyway, just make value lowercase in both places.

xenigmax:
It seems every issue has resolved. LGTM.

PR #1521, opened 2016-03-02
platform support: OpenBSD

h-2:
This PR adds basic support for OpenBSD as a platform. 6 out of  377 test still fail, but that will be tackled later.


xenigmax:
The changes are mainly about using of safer functions (eg. sprintf -> snprintf).
This is a mandatory thing in OpenBSD.

I have checked all the boundary conditions. LGTM :)

PR #1522, opened 2016-03-02
Platform support: OpenBSD

h-2:
see #1521


xenigmax:
The changes are mainly about using of safer functions (eg. sprintf -> snprintf).
This is a mandatory thing in OpenBSD.

I have checked all the boundary conditions. LGTM :)

PR #1529, opened 2016-03-07
rework of seeds

xenigmax:
The previous "seed and extend" section was divided into "seed" and "seed and extend" sections.
This PR will separate the seed section from the ancestor more clearly.

### DIFF
--- a/manual/source/Tutorial/DataStructures/Seeds.rst
+++ b/manual/source/Tutorial/DataStructures/Seeds.rst
@@ -8,21 +8,18 @@ Seeds
 =====
 
 Learning Objective
-  In this tutorial, you will learn about the seeds-related SeqAn functionality.
-  You will learn how to do seed-and-extend with SeqAn, how to do local and global chaining of seeds.
-  Finally, we will show how to create a banded alignment around a seed chain.
+  In this tutorial, you will learn about the seeds-related SeqAn class.
 
 Difficulty
-  Average
+  Basic 
 
 Duration
-  2 h
+  15 min 
 
 Prerequisites
   :ref:`tutorial-datastructures-sequences`
 
-Many efficient heuristics to find high scoring, but inexact, local alignments between two sequences start with small exact (or at least highly similar) segments, so called **seeds**, and extend or combine them to get larger highly similar regions.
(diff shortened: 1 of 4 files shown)


manual/source/Tutorial/DataStructures/Seeds.rst
```diff
@@ -44,16 +40,16 @@ In most cases, the :dox:`SimpleSeed Simple Seed` class is sufficient since the b
 
 You can get/set the begin and end position in the horizontal/vertical sequences using the functions :dox:`Seed#beginPositionH`, :dox:`Seed#beginPositionV`, :dox:`Seed#setBeginPositionH`, and :dox:`Seed#setBeginPositionV`.
 The band information can be queried and updated using :dox:`Seed#lowerDiagonal`, :dox:`Seed#upperDiagonal`, :dox:`Seed#setLowerDiagonal`, and :dox:`Seed#setUpperDiagonal`.
-Note, we use the capital letters 'H' and 'V' to indicate horizontal direction and vertical direction, respectively, while the database is always considered as the horizontal sequence and the query as the vertical sequence in the context of sequence alignments.
+Note, we use the capital letters 'H' and 'V' to indicate horizontal direction and vertical direction, respectively, while the **database** is always considered as the **horizontal sequence** and the **query** as the **vertical sequence** in the context of sequence alignments.
```
h-2: I think we should use the word **subject sequence** instead of **database**, because it is more general. So always:

  * **query** instead of *read*
  * **subject* instead of *reference* or *database* or *contig*

xenigmax: I agree that "subject sequence" is a more general term.
However, I can find this much of "databases" in the current tutorial.
http://seqan.readthedocs.org/en/feature-manual-restructure/search.html?q=database&check_keywords=yes&area=default#

Should we consider them as well?

h-2: > Should we consider them as well?

you have just earned a new task :)

esiragusa:
@h-2 ?

h-2:
@xenigmax can you adjust this PR already so that we can merge it?

xenigmax:
@h-2  
I *accidently* merged this to the manual branch by myself.
However, I will change "database" to "subject sequence" for resolving #1532 .
Let me know if you find another problem so that I will revert or patch it.

Sorry for the inconvenience.

PR #1540, opened 2016-03-08
reworked all the infrastructure, cmake and getting started stuff

h-2:
@cpockrandt can you merge if evreything is green?
@cpockrandt @temehi can you both check everything in "infrastructure" *after* it is merged? Especially the stuff regarding windows and mac, but maybe also in general and also language? Because there is a lot of new documentation...

Thank you!


cpockrandt:
Not green, but since only test_bam_io failed, I'll merge it anyway :wink:

Issue #1530, opened 2016-03-07
better default of  SEQAN_HAS_EXECINFO


h-2:
Right now it is 0 by default and header is detected by cmake. 

This should be changed and set in platform.h:
  * if compiler can do header detection, detect and set
  * if not and MSVC set to 0
  * otherwise set to 1

h-2:
fixed by #1543

PR #1543, opened 2016-03-09
[INTERNAL] more sensible execinfo detection, also outside cmake

h-2:
before this variable was only set by cmake. Now we do some detection in the code, too, in case people use other build systems.


h-2:
fixes #1530

cpockrandt:
lgtm

But is there a reason why it is ``_SEQAN_HAVE_EXECINFO`` and not ``_SEQAN_HAS_EXECINFO`` to be consistent?

PR #1535, opened 2016-03-07
[FIX] delete temporary files when running unit tests

smehringer:
Fixes #1387.

PR #1549, opened 2016-03-10
[FIX] Fixes windows warnings when deleting temporary files.

smehringer:
Addition to #1535


cpockrandt:
Much better code now :+1: :grin:

PR #1550, opened 2016-03-10
[FIX] Fixes windows warnings when deleting temporary files.

smehringer:
Corresponds to master PR #1549.


cpockrandt:
Much better code now :+1: :grin:

PR #1525, opened 2016-03-03
[NOP, DOC] changed INPUTPREFIX to INPUT_PREFIX (develop)

temehi:
- changed INPUTPREFIX to INPUT_PREFIX to be consistent with other ArgumentTypes
- added missing documentation in the dox for INPUT_PREFIX and OUTPUT_PREFIX ArgumentTypes.

@esiragusa Yara_mapper (mapper.cpp, line 116)  is affected by this could you please adjust to this change?

PR #1524, opened 2016-03-03
[NOP, DOC] changed INPUTPREFIX to INPUT_PREFIX

temehi:
- changed INPUTPREFIX to INPUT_PREFIX to be consistent with other ArgumentTypes
- added missing documentation in the dox for INPUT_PREFIX and OUTPUT_PREFIX ArgumentTypes.

@esiragusa Yara_mapper (mapper.cpp, line 116)  is affected by this could you please adjust to this change?

PR #1551, opened 2016-03-10
[API] fixes the broken API by #1524

temehi:
@h-2 as we have discussed I created this PR to maintain the API backward compatibility which was broken by #1524 and #1525. @esiragusa I can revert the commit to yara but that will cause a "deprecated" warning...

I do not see a reason to revert the previous  PRS and it will be doing the same thing twice.


esiragusa:
@temehi Thank you for this PR! Yara is good now, please leave it as it is.

temehi:
ah, unfortunately __attribute__((deprecated))  wont work for enum values in gcc. Removed it. This means we wont have compiler warnings for using the deprecated value :cry:  but we still have API backward compatibility.

PR #1553, opened 2016-03-10
Review of Infrastructure manual

cpockrandt:


manual/source/Infrastructure/Contribute/index.rst
```diff
@@ -8,7 +8,7 @@ SeqAn is on GitHub: http://github.com/seqan/seqan
 Use the GitHub page to fork SeqAn, create tickets and/or pull requests.
 You can also follow and like the project there!
 
-Contributing Code or documentation
+Contributing Code or Documentation
 ----------------------------------
 
 If you are unfamiliar with git, you need to learn about it first. TODO link
```
h-2: do you know about a general git tutorial we could reference?

manual/source/Infrastructure/Use/CMakeBuildDirs.rst
```diff
@@ -110,7 +110,7 @@ Different versions (please note that versions older than 2015 are not supported
 
     **64Bit builds on Windows**
 
-    You almost always want 64Bit builds when using SeqAn so don't forget to specify a generator that ends in "64Bit", because it is not the default, even on 64Bit Windows installs.
+    You almost always want 64Bit builds when using SeqAn, so don't forget to specify a generator that ends in "64Bit". It is not the default, even on 64Bit Windows installations.
```
h-2: I think it should end in "Win64" actually, or not?

h-2:
Thanks for the review, did you also try the things, e.g. cmake example on windows?

manual/source/Infrastructure/Contribute/index.rst
cpockrandt: How about https://www.atlassian.com/git/tutorials/ ?

manual/source/Infrastructure/Use/CMakeBuildDirs.rst
cpockrandt: Yes, you're right, it should be ``Win64``.

manual/source/Infrastructure/Contribute/index.rst
h-2: hm, they are ok, but something more like a "git in 30min" page would be better I think... but it's not so important.

manual/source/Infrastructure/Use/CMakeBuildDirs.rst
h-2: Ok, then please change it your PR!

cpockrandt:
The cmake part looks good to me.

There are two sentences I struggled with rewriting:

In ./Infrastructure/Use/CustomBuildSystem.rst (I don't understand the sentence)
> Note that we added the include path to the directory ``include`` that contains the directory ``seqan``.

In ./Infrastructure/Manage/WriteUnitTests.rst (I don't think this is grammatically correct)
> Each test should isolate target an as small as possible and/or feasible unit of your code.

h-2:
> In ./Infrastructure/Use/CustomBuildSystem.rst (I don't understand the sentence)

Replace it with
```
Above we added the include path to SeqAn's include directory manually.
```

In the same file there is also the Makefile.rules and it include this line twice:
```
CXXFLAGS+=-I../../../include
```
Can you remove one of the lines?

> In ./Infrastructure/Manage/WriteUnitTests.rst (I don't think this is grammatically correct)

Can you replace it with:

```
Each test should verify a part of the library as small as possible while still being meaningful.
```

xenigmax:
All the issues raised by others have resolved. I don't find anything to add.
LGTM.

manual/source/Infrastructure/Manage/RepositoryStructure.rst
```diff
@@ -26,7 +26,7 @@ SeqAn supports the usual CMake build types and we recommend :ref:`using multiple
 In addition to ``CMAKE_BUILD_TYPE`` there is also the ``SEQAN_BUILD_SYSTEM`` parameter which can be one of
 
 #. ``DEVELOP`` -- all build targets (apps, demos, tests) and documentation (dox, manual) are created (the default).
-#. ``SEQAN_RELEASE_LIBRARY`` -- only dox and libraray target created.
+#. ``SEQAN_RELEASE_LIBRARY`` -- only dox and library targets are created.
```
xenigmax: Oops. I found one thing.
According to #1560 "SEQAN_RELEASE_LIBRARY" will not contain dox anymore.
I didn't check that PR in detail but it describes like that.

Don't you think that we have to change this line?

xenigmax: OK. It's an option. That PR add "SEQAN_NO_DOX". So the default mode contains dox like before.

xenigmax:
LGTM!

h-2:
> But if you think it's better to add that mode to the manual, please just do it.

By default the library install contains the dox. I don't think we need to document this flag right now, but you can if you want to!

Issue #1532, opened 2016-03-07
Change "database" to "subject sequence" in the tutorial


xenigmax:
@h-2  suggested me to change "database" to "subject sequence" in #1529 
And there are many "database" in the tutorial.
http://seqan.readthedocs.org/en/master/search.html?q=database&check_keywords=yes&area=default

If there is no objection about this, I will change all of these.
@rrahn @esiragusa  What do you think?

xenigmax:
The suggestion was the substituon of words like :

1) database/reference/contig => subject sequence
2) read => query

However, there are many occurences of "database", "reference", "contigs" and "read".
And some of them should not be changed to deliver clear explanation.

eg.
"FragmentStore is a *database*"
"Do not use pointers. Use *references*"
"..against a *reference* genome"
".. a genome consist of *contigs*"

I made a quick fix for changing obvious cases including the one in #1529 and move this issue to the backlog.
We may need a discussion before chainging entire manual.

PR #1556, opened 2016-03-14
database/reference -> subject sequence

xenigmax:
Based on @h-2 's suggestion  from #1529 
This PR is related to #1532 

Obviously, we have to use "database" and "reference" for some cases.
For example, when we say "FragmentStore is a database",  we should not change the word "database" to "subject sequence".
That's same in "reference", "contig" and "read".

So this PR changes only a part of them, that means only changes obvious cases.
For other parts, I think we need a discussion.

### DIFF
--- a/manual/source/Tutorial/DataStructures/Sequence/Sequences.rst
+++ b/manual/source/Tutorial/DataStructures/Sequence/Sequences.rst
@@ -309,8 +309,8 @@ Assignment 4
      In this task you will compare whole sequences.
      Reuse the code from above. Instead of a ``String<Dna5>`` we will now deal with a ``String<Dna5String>``.
      Build a string which contains the Dna5Strings "ATATANGCGT", "AAGCATGANT" and "TGAAANTGAC".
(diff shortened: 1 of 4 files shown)


manual/source/Tutorial/DataStructures/Sequence/Sequences.rst
```diff
@@ -310,7 +310,7 @@ Assignment 4
      Reuse the code from above. Instead of a ``String<Dna5>`` we will now deal with a ``String<Dna5String>``.
      Build a string which contains the Dna5Strings "ATATANGCGT", "AAGCATGANT" and "TGAAANTGAC".
      Now check for all elements of the container, if they are lexicographically smaller or bigger than the  given subject sequence "GATGCATGAT" and append them to a appropriate list.
-     Print out the final list.
+     Print out the final lists.
```
smehringer: when I look into the current file Sequence.rst on feature/manual/restructure it already says "lists"

https://github.com/seqan/seqan/blob/feature/manual/restructure/manual/source/Tutorial/DataStructures/Sequence/Sequences.rst

does this perhaps cause a conflict? It's correct though of course

smehringer:
This all looks fine to me. I think it can be merged.

Issue #1481, opened 2016-02-19
CMake caching of OpenMP flags is broken


esiragusa:
CMake now tries to detect OpenMP flags for each app, test folder and demo that needs them. 

```
-- Try OpenMP CXX flag = [ ]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [-fopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [/openmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [-Qopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [-openmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [-xopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [+Oopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [-qsmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Try OpenMP CXX flag = [-mp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Failed
-- Could NOT find OpenMP (missing:  OpenMP_CXX_FLAGS) 
```

Note that this thing is particularly slow for Xcode builds. Before this lookup was done only once and cached.

h-2:
We have removed a bunch of cmake code before the release, because it was duplicated system code that caused problems in different places.
Do you have an idea for caching that does not involve maintain modified copies of system code?

esiragusa:
Can't you have a look at the code that you removed?

h-2:
[...]

Issue #1478, opened 2016-02-17
v2.1 SEQAN_BUILD_SYSTEM=SEQAN_RELEASE_LIBRARY make test fails


mr-c:
```
mcrusoe@mrcdev:~/src/seqan-build$ cmake ../seqan-seqan-v2.1.0/ -DSEQAN_BUILD_SYSTEM=SEQAN_RELEASE_LIBRARY
-- The CXX compiler identification is GNU 5.3.1
-- Check for working CXX compiler: /usr/lib/ccache/c++
-- Check for working CXX compiler: /usr/lib/ccache/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Initializing SeqAn Build System...
-- Performing Test CXX11_DETECTED
-- Performing Test CXX11_DETECTED - Success
-- Performing Test CXX14_DETECTED
-- Performing Test CXX14_DETECTED - Success
--   Selected repository dir: /home/mcrusoe/src/seqan-seqan-v2.1.0
-- No revision system found.
--   Repository date not determined.
--   Repository revision not determined.
-- Configuring apps
-- Configuring apps
-- Configuring dox
-- Found PythonInterp: /usr/bin/python (found version "2.7.11") 
-- Configuring manual
-- Found Sphinx: /usr/bin/sphinx-build  
-- Configuring util/py_lib
--   adding nosetests for seqan.dox
-- Configuring done
-- Generating done
-- Build files have been written to: /home/mcrusoe/src/seqan-build
```
```
mcrusoe@mrcdev:~/src/seqan-build$ make test
Running tests...
Test project /home/mcrusoe/src/seqan-build
    Start 1: build_dox
1/3 Test #1: build_dox ........................   Passed   25.00 sec
    Start 2: build_manual
2/3 Test #2: build_manual .....................***Failed    0.37 sec
    Start 3: test_py_lib_dox
3/3 Test #3: test_py_lib_dox ..................   Passed    0.59 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =  25.97 sec

The following tests FAILED:
          2 - build_manual (Failed)
Errors while running CTest
Makefile:61: recipe for target 'test' failed
make: *** [test] Error 8
```
```
mcrusoe@mrcdev:~/src/seqan-build$ uname -a
[...]

PR #1560, opened 2016-03-14
build system changes

h-2:
this implements:
  * don't ship apps and the manual on library releases
  * Introduce cmake -DSEQAN_OFFICIAL_PKGS=1 to build upstream releases with static binaries
  * cache dependencydetection on DEVELOP
  * make it possible to do RELEASE_LIBRARY without dox
  * lots of cleanup, make yara behave like the rest

it solves/supersedes:
#1505 #1481 #1478 

**please also test creating release packages before merging this**


h-2:
@esiragusa can you greenlight the changes to yara? Thanks!

apps/yara/CMakeLists.txt
```diff
@@ -14,22 +14,14 @@ message (STATUS "Configuring apps/yara")
...
 # ----------------------------------------------------------------------------
 
-# Disable older MSVC.
-if (MSVC AND MSVC_VERSION LESS 1800)
-    message (STATUS "  Visual Studio 2012 or newer required to build yara.")
-    return ()
-endif ()
-
-# Disable parallelism in MSVC as it supports only OpenMP 2.0.
-if (NOT MSVC)
```
esiragusa: @h-2 I had disabled OpenMP under MSVC. Please re-apply this check.

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -180,41 +173,78 @@ macro (seqan_build_system_init)
...
+        (SEQAN_BUILD_SYSTEM MATCHES "APP")) # either APP:$app or SEQAN_RELEASE_APPS
+
+        # static linkage
         set (SEQAN_STATIC_APPS TRUE CACHE INTERNAL "Create static app binaries")
+        message (STATUS "Building static binaries.")
+
+        # machine specific optimizations
+        if ((CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64") OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64"))
+            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mmmx -msse -msse2 -msse3 -mssse3 -mpopcnt")
+            message (STATUS "Release binaries built with optimizations for SSE3 and POPCNT.")
```
esiragusa: You can add -msse4.2 as well because popcnt in practice is available from sse4.2 cpus upwards.

apps/yara/CMakeLists.txt
h-2: I had thought that vs2015 had openmp3, but apparently its not the case :facepunch: 

Well, since we have global detection now (for caching), it cannot be disabled locally, without doing major regex voodoo to all the build vars.
Can I just deactivate builds entirely for msvc? Nobody would seriously use it without threading anyway, would they?

util/cmake/SeqAnBuildSystem.cmake
h-2: I think we would loose AMD barcelona which was quite popular...

esiragusa: Then -msse4...

apps/yara/CMakeLists.txt
esiragusa: If we can distribute windows/icc binaries then yes.

h-2: We don't currently support windows/ICC at all... We will support windows/c2 (clang) as soon as cmake supports it, but it will likely take another few months.

util/cmake/SeqAnBuildSystem.cmake
h-2: barcelona has SSE4a which is a different instruction set than SSE4.1/SSE4.2 on Intel. GCC also has a `-msse4` but documentation implies this might be just ==`-msse4.1 -msse4.2` so it is also not compatible.

apps/yara/CMakeLists.txt
h-2: I am on a conference, so I cannot test anything on windows (and on vacation immediately after that). If you want to still have single-threaded visual builds for yara, can you send me diff (or make a PR) that does
  * remove `/openmp` from `SEQAN_CXX_FLAGS`
  * replace `-DSEQAN_HAS_OPENMP=1` with =0 in SEQAN_DEFINITIONS
?

Thanks!

esiragusa: Excellent! Then you have so much free time to change it during the trip back :trollface: 

Okay seriously, I don't like this solution either, so you can disable MSVC for now. But my questions are:
* Where and how do you do ```find_package``` now? Are OpenMP flags added to all targets?
* Why don't you make Windows/ICC builds a priority? Windows/MSVC builds suck because of OpenMP 2.0...

h-2: > Okay seriously, I don't like this solution either, so you can disable MSVC for now.

Will do.

> Where and how do you do find_package now? Are OpenMP flags added to all targets?

Its in this diff, so feel free to do some easter egg searching ;-) Actually for BUILD_SYSTEM==DEVELOP we detect the dependencies once in `build_system_init()` and then add them to every target. For release builds everything is still done locally.

> Why don't you make Windows/ICC builds a priority? Windows/MSVC builds suck because of OpenMP 2.0...

Let me think... I think it's because I am neither responsible for Intel stuff nor for Windows stuff :bulb:  (although I have done 90% of making the library work with ICC).

I think @rrahn, @marehr and @cpockrandt  would all be more suited for the task. @cpockrandt and I tried to make "Clang with Microsoft Codegen"/C2 work to improve the situation on Windows, but cmake on Windows doesn't work with it, yet. Also we haven't even been given access to the Windows version of ICC, just Linux right now (I will discuss that with Intel folks here at IXPUG).

But I agree, we really should get an alternative to MSVC!

h-2:
@cpockrandt and @xenigmax can you verify that the package builds work and that they are static/not static depending on what you want? If yes, it is good to merge from my POV.

apps/yara/CMakeLists.txt
esiragusa: There's nothing to feel (un)responsible and whine. I only suggested to you all (not just to you you) to support Windows/ICC in order to distribute good Windows binaries. This seems to me more urgent than supporting platforms like Windows/c2 or OpenBSD/whatever. Also, please note that I am not suggesting to remove support for Windows/MSVC in favor of Windows/ICC. Thanks.

cpockrandt:
@h-2 I reviewed your PR and tested ``DSEQAN_NO_DOX``as well as whether the app builds are (non-)static depending on ``DSEQAN_OFFICIAL_PKGS``. I think we can merge if @xenigmax agrees.

xenigmax:
@cpockrandt  Please do it if you think it's OK. :) I don't think that we have to do verifications two times. Thank you.

PR #1567, opened 2016-03-16
an essential rework of seed extensions

xenigmax:
This PR reorganize seed extension part of the tutorial.

Previously, the _seed_ and _seed extension_ part were not divided. So example/solution.cpp initially started with number 2 because example1/solution1.cpp were used in the _new seed section_.
I have corrected and moved all of them. That's why there are so many changes(in diff.) in this PR.


include/seqan/seeds/seeds_extension.h
```diff
@@ -796,7 +796,7 @@ extendSeed(Seed<Simple, TConfig> & seed,
     // SEQAN_ASSERT_LT(scoreMismatch(scoringScheme), 0);
     SEQAN_ASSERT_LT(scoreGapOpen(scoringScheme), 0);
     SEQAN_ASSERT_LT(scoreGapExtend(scoringScheme), 0);
-    SEQAN_ASSERT_EQ(scoreGapExtend(scoringScheme), scoreGapOpen(scoringScheme));
+    //SEQAN_ASSERT_EQ(scoreGapExtend(scoringScheme), scoreGapOpen(scoringScheme));
```
xenigmax: @h-2  @esiragusa 
I would like to ask you check this.

Because of this assertion, we cannot use different scores for gapExtend and gapOpen. 
This is nonsense. Isn't it?
So I would like to comment this out.

If we have to maintain this we also have to remove assignment#4 which uses Gotoh algorithm.
http://seqan.readthedocs.org/en/master/Tutorial/SeedAndExtend.html

Let me know whether it's OK or not :)

xenigmax:
@smehringer  Please review this PR when you have a time. However, you don't need to check the one that  I mentioned above(about assertion). Thank you :)

h-2:
IIRC the implementation just doesnt support affine gap costs, so please double-check that.
I think it was one of the reasons I wrote extendalignment() (which is based on our regular DP.

manual/source/Tutorial/Algorithms/SeedExtension.rst
```diff
@@ -57,25 +56,25 @@ Assignment 2
...
      .. container:: foldable
 
-        .. includefrags:: demos/tutorial/seed_and_extend/solution2.cpp
+        .. includefrags:: demos/tutorial/seed_and_extend/solution1.cpp
 
-A more complex case is the standard bioinformatics approach of **x-drop extension**:
+A more complex case is the standard bioinformatics approach of **x-drop extension**.
 
-In the ungapped case, we extend the seed by comparing the *i*-th character to the left/right of the seed of the horizontal sequence with the *j*-th character to the left/right of the seed in the vertical sequence.
+In the ungapped case, we extend the seed by comparing the *i*-th character to the left/right of the seed of the horizontal sequence(subject sequence) with the *j*-th character to the left/right of the seed in the vertical sequence(query sequence).
```
marehr: maybe a space after sequence

e.g.  ``horizontal sequence (subject sequence)``

marehr:
lgtm :+1:

include/seqan/seeds/seeds_extension.h
xenigmax: OK. I found that we need an investigation about this.
So I do not change the source code and make another issue for this.

xenigmax:
I excluded changes in source code from this PR.
Check #1596 for seed_extension with different gapOpen/Extension code.

I will merge this if continous-integration works since Marcel has checked the content.

Issue #1573, opened 2016-03-17
Strange behaviour of indexed fasta I/O when building an index twice


smehringer:
This issue occurs in the following demo file:
> demos/tutorial/indexed_fasta_io/base.cpp

The function `getIdByName()` raises the error, that it cannot find "chr1". The _example.fasta_ file does contain a sequence named "chr1" though.

If you try out `std::cout << empty(faiIndex) << std::endl;`, it returns `1`. 

Strangely, after `getIdByName()` fails, `sequenceLength()` successfully fetches the length of the sequence. But `empty(faiIndex)` still returns `1`...

This is not a problem for the tutorials, since the file is not included in any assignment.
@temehi can you take a look at it? Maybe some time after the release.

smehringer:
@cpockrandt can you please add labels, Milestone and Assignee.

temehi:
@cpockrandt you may close this issue now

cpockrandt:
Oops, forgot it. Thanks!

PR #1574, opened 2016-03-17
[FIX] closes the fasta file after fai_index is built

temehi:
This solves issue #1573.

### DIFF
--- a/demos/tutorial/indexed_fasta_io/base.cpp.stdout
+++ b/demos/tutorial/indexed_fasta_io/base.cpp.stdout
@@ -1,1 +1,0 @@
-ERROR: FAI index has no entry for chr1.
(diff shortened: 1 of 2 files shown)


cpockrandt:
Thanks for the fast fix! Can you adjust the output file ``demos/tutorial/indexed_fasta_io/base.cpp.stdout``? There shouldn't be any output anymore.

temehi:
@cpockrandt I was wrong to close the fast a file. It actually needs to stay open so that it can be accessed with readregion..and so on. Instead a new solution would be to put the head of the ifstream to the beginning of the file when clear is called. I was also wrong about the tests. They are fine as they are.

cpockrandt:
LGTM.

Thanks @temehi !

PR #1577, opened 2016-03-17
Fix #1572: disable building with -march=native for 32 bit target platforms

### LINKED ISSUE #1572: 32bit fails...
... don't worry, we are working on it.

marehr:
Fixes #1572

### DIFF
--- a/util/cmake/SeqAnBuildSystem.cmake
+++ b/util/cmake/SeqAnBuildSystem.cmake
@@ -73,6 +73,19 @@ if (MINGW)
 endif (MINGW)
 
 # ---------------------------------------------------------------------------
+# Is it a 32 bit platform?
+# ---------------------------------------------------------------------------
+if (CMAKE_SIZEOF_VOID_P EQUAL 4)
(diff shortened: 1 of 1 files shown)


marehr:
Maybe to give a little bit more information:
This will fix the many building errors on http://cdash.seqan.de/index.php?project=SeqAn for 32 bit platforms.

xenigmax:
LGTM. Thanks

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -234,8 +247,11 @@ macro (seqan_build_system_init)
     # settings for development mode
     if ((SEQAN_BUILD_SYSTEM STREQUAL "DEVELOP") AND
         (NOT MSVC) AND
-        (NOT SEQAN_NO_NATIVE))
+        (NOT SEQAN_NO_NATIVE) AND
+        (NOT SEQAN_32BIT_TARGET_PLATFORM))
```
xenigmax: It seems too stringent. Some people may want to make CPU-optimized binaries in 32bit machine.
However, I think it's just enough for this time.

PR #1579, opened 2016-03-17
[FIX, TEST] fixes demo_checker.py

### LINKED ISSUE #1576: test of demo file passes even when the stdout was not as expected.
I found the following two tests which passed, even if the stdout was not as expected.

http://cdash.seqan.de/testDetails.php?test=27779920&build=169398
http://cdash.seqan.de/testDetails.php?test=27779935&build=169398

They both have in common that the stored .stdout file was empty. 

This shoud not happen, should it?
Or is there some special case where a test is allowed to have different standard output?

@xenigmax can you assign someone who might know this?

temehi:
fixes #1576


temehi:
Well, As expected there were 19 tests which were passing because of this "loophole" :smile: 

> The following tests FAILED:
	100 - test_demo_dox_align_gaps_example (Failed)
	107 - test_demo_dox_basic_zip_iterator (Failed)
	116 - test_demo_dox_graph_algorithms_dijkstra (Failed)
	138 - test_demo_dox_modifier_cyclic_shape (Failed)
	148 - test_demo_dox_sequence_container_view_zip (Failed)
	151 - test_demo_dox_stream_lexical_cast (Failed)
	153 - test_demo_howto_scores_load_score (Failed)
	174 - test_demo_tutorial_basics_allocator (Failed)
	179 - test_demo_tutorial_bed_io_example1 (Failed)
	180 - test_demo_tutorial_bed_io_solution1 (Failed)
	190 - test_demo_tutorial_fragment_store_access_aligned_reads2 (Failed)
	282 - test_demo_tutorial_q_gram_index_index_assignment6 (Failed)
	326 - test_demo_tutorial_sequences_in_depth_assignment_exact_generous_solution (Failed)
	331 - test_demo_tutorial_simple_rna_seq_genequant_assignment2 (Failed)
	332 - test_demo_tutorial_simple_rna_seq_genequant_assignment3 (Failed)
	333 - test_demo_tutorial_simple_rna_seq_genequant_assignment4 (Failed)
	334 - test_demo_tutorial_simple_rna_seq_genequant_assignment5 (Failed)
	350 - test_demo_tutorial_vcf_io_example1 (Failed)
	351 - test_demo_tutorial_vcf_io_solution1 (Failed)

Most of them have an empty .stdout as it was mentioned by @smehringer in #1576. 

But some of them like  `test_demo_tutorial_basics_allocator` has an stdout with the time_elapsed to compute some staff...and this will never be similar for different runs. We need to find a solution for that. May be make demo-checker.py to do some regex or ...

Others like `test_demo_dox_modifier_cyclic_shape` are failing because one is trying to `cout <<` an `unsigned char` type. For the time being I will fix only the demos. But for the future, we could do something like overloading the operator `<<` for `unsigned char` 

```c++
std::ostream& operator<<(std::ostream& oStream, const unsigned char & s)
{
    oStream << (int)(s);
    return oStream;
}
```

I am going to fix as many of these tests as I can. If I don't make it then at the release sprint ....

temehi:
Uninitialised member variable isDefaultPairwiseAlignment was used [here](https://github.com/temehi/seqan/blob/hotfix/%231576/include/seqan/graph_msa/graph_align_tcoffee_msa.h#L394) and [here](https://github.com/temehi/seqan/blob/hotfix/%231576/include/seqan/graph_msa/graph_align_tcoffee_msa.h#L394). This was causing a subtle error in MSA. The last commit will fix the failing tests of this PR and closes issues #1576 and #1585

PR #1580, opened 2016-03-17
[FIX, TEST] fixes demo_checker.py (develop)

### LINKED ISSUE #1576: test of demo file passes even when the stdout was not as expected.

temehi:
fixes #1576


temehi:
It looks like `globalMsaAlignment()` produces different outputs for b/n clang and gcc.
The test is failing because of that. I will create an issue for this.

cpockrandt:
Lgtm, ty! :+1:

PR #1587, opened 2016-03-21
[FIX] deletes all temp files that are generated by the app bs_tools.

smehringer:
@cpockrandt can you have a look ? :smiley:


cpockrandt:
lgtm!

PR #1594, opened 2016-03-23
performance degrade on windows

marehr:
Attempt to fix #1586.

The windows binaries are currently taking much more time than some days ago. After doing a bisect on windows, it seems to be related to the reverted commits in this pr. I'm not completely sure what is going on, but it seems that the generated build files for windows doesn't set the optimization flags as they should/as they did before.

PR #1597, opened 2016-03-23
performance degrade on windows (develop)

### LINKED ISSUE #1586: app_test_razers/app_test_stellar timeouts after 600s (performance degrade)
http://cdash.seqan.de/testDetails.php?test=31266148&build=169958

marehr:
Attempt to fix #1586.

The windows binaries are currently taking much more time than some days ago. After doing a bisect on windows, it seems to be related to the reverted commits in this pr. I'm not completely sure what is going on, but it seems that the generated build files for windows doesn't set the optimization flags as they should/as they did before.


xenigmax:
Check #1594

PR #1613, opened 2016-03-24
[FIX] Temporarily changing stderr on Windows because of VS14 bug

cpockrandt:
As requested, I changed the ``cerr`` for the release.


xenigmax:
This should be changed since our demo_checker cannot check platform-dependent output/errors.
LGTM.

PR #1615, opened 2016-03-24
Yara: verify seeds

esiragusa:


esiragusa:
@h-2 @rrahn @xenigmax FYI I am rolling back Yara's version to 0.9.6 (35962be). I will rebuild packages and tag yara-v0.9.6 afterwards. Sorry for messing up.

It would be nice if you let me know when you touch Yara in the future. It would be nicer if apps could stay out of the library repo... :disappointed:

PR #1617, opened 2016-03-26
Build System Cleanup

esiragusa:
@h-2 @rrahn I am proposing some modification to the build system, following recent changes in #1560:
* Split cmake flag SEQAN_OFFICIAL_PKG as SEQAN_STATIC_APPS and SEQAN_OPTIMIZED_BUILDS
* Removed optimizations in DEVELOP builds

The reasons for these modifications should be pretty obvious. As soon as you agree, I will update the documentation and submit a PR for develop.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,9 @@ option (SEQAN_ENABLE_CUDA
         "Enable CUDA, if the toolkit is available."
 	OFF)
 
+option (SEQAN_STATIC_APPS "Build static apps." OFF)
+option (SEQAN_OPTIMIZED_BUILDS "Build optimized binaries." OFF)
+
 # ===========================================================================
 # Setup Modules and Find Packages.
 # ===========================================================================
--- a/util/cmake/SeqAnBuildSystem.cmake
+++ b/util/cmake/SeqAnBuildSystem.cmake
@@ -213,51 +213,31 @@ macro (seqan_build_system_init)
         set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} /D_SCL_SECURE_NO_WARNINGS")
     endif ()
 
-    # DEFFAULT BUILD SYSTEM
     if (NOT SEQAN_BUILD_SYSTEM)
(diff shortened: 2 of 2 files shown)


h-2:
> Split cmake flag SEQAN_OFFICIAL_PKG as SEQAN_STATIC_APPS and SEQAN_OPTIMIZED_BUILDS

ok...

> Removed optimizations in DEVELOP builds

Why? You yourself requested this :question:

esiragusa:
No, I didn't request that! I think there has been a misunderstanding. For me it's enough to have a flag SEQAN_OPTIMIZED_BUILDS and distribute optimized builds, so people don't have to compile the app on their own. Default optimizations in develop mode are useless and require a warning message.

h-2:
I have just discussed this with @rrahn and we agree in most points. However, we would propose changing SEQAN_OPTIMIZED_BUILDS to SEQAN_CPUOPT_SSE4 because "OPTIMIZED" is too general (could apply to -O for example). Also native should optionally be available (but off by default). So we have:

SEQAN_STATIC_APPS (same as now)
SEQAN_CPUOPT_SSE4
SEQAN_CPUOPT_NATIVE

Check if _NATIVE is defined first, if not check if _SSE4 is set, else do nothing.

Do you agree?

Also, could you please adapt the infra/manage/deploy tutorial to recommend setting SEQAN_STATIC_APPS and SEQAN_CPUOPT_SSE4 instead of SEQAN_OFFICIAL_PKG?

h-2:
Actually, if that is ok, I would take over the PR, because I have to do other changes to the build system anyway... ok?

esiragusa:
Sure, you can take it over. I would propose SEQAN_ARCH_NATIVE and SEQAN_ARCH_SSE4. These names correspond a bit better to flags -march=native and -march=sse4. Would be good to change cpack name suffixes from x86_64 to sse4 and native along with SEQAN_ARCH_NATIVE and SEQAN_ARCH_SSE4.

h-2:
superseded by #1654 

@esiragusa saw your response too late, but I agree with your terms. I will change it. regarding cpack that will create lots of problems with our pipelines for generating packages.seqan.de, so I would rather not do it. Instead I will put a static warning on the page that newer binaries require certain cpu extensions.

PR #1632, opened 2016-03-31
[DOC] Changes to IndexedFastaIO tutorial (develop)

marehr:
* Fix that the demos/tutorial/indexed_fasta_io/example.fasta.fai file is empty
* When reading the tutorial the variable `pathToFaiFile` is undefined, therefore remove any reference to it.


cpockrandt:
lgtm :smiley:

Issue #1526, opened 2016-03-04
PizzaChili Index Adapter not working


cpockrandt:
I just noticed, that the PizzaChili index is not working (use of undeclared identifier `SEQAN_ABORT` and `SEQAN_REPORT`). It should be easy to fix the bug but there are some open questions/issues:
- the adapter is not tested anywhere
- is anyone using this adapter?
- there are many indices in the Pizza & Chili API (see http://pizzachili.dcc.uchile.cl/indexes.html) that SeqAn does not support. 

@esiragusa Could you comment on this issue? Do you think we should still support the API or just remove it from SeqAn? It seems to me that noone uses it and we don't even nearly support all indices from the API.

esiragusa:
I would say to remove it as we did for other broken and unused modules. The code is released in v2.1.0, in case anyone wants to work on it (but I would suggest to integrate sdsl-lite rather than pizza&chili).

cpockrandt:
@h-2 If you agree, I will remove it and open a PR into develop for Milestone 2.1.1

h-2:
Hm, I don't think we should remove any code anything for a 0.0.+1 release.  Or are you sure that absolutely all of the functions and datastructures are broken?
For the manual and dox we can definitely remove it...

xenigmax:
It seems we all want to remove this part from the manual at least.
I made a new issue #1568 for this and reassign this issue to backlog.

h-2:
Closed because inactivity.

PR #1657, opened 2016-04-04
[FIX, DOC] Mark Pizza&Chili as deprecated

cpockrandt:
Referring to issue #1526.

I couldn't easily fix the Pizza&Chili-Index and thus just marked it as deprecated.


rrahn:
lgtm!

PR #1665, opened 2016-04-06
[FEATURE] support intel compiler 

marehr:
This fixes some bugs which allow building seqan with the intel compiler 16.0.2.

See ticket #1274

### DIFF
--- a/apps/fiona/fiona.cpp
+++ b/apps/fiona/fiona.cpp
@@ -83,7 +83,6 @@
     #endif  // #if !defined(__MINGW32__) || defined(__amd64__) || defined(__x86_64__) || defined(__ia64__)
 #else
     #warning "Please enable OpenMP."
-    #define omp_get_wtime() 0
 #endif  // #ifdef _OPENMP
 
 // The q-gram length used for the q-gram index.  This has to be hard-coded as a precompiler definition since it is part
--- a/apps/fiona/fiona.cpp
+++ b/apps/fiona/fiona.cpp
@@ -81,6 +81,13 @@
     // Thus, we do not enable it in this case.
         #define _GLIBCXX_PARALLEL
     #endif  // #if !defined(__MINGW32__) || defined(__amd64__) || defined(__x86_64__) || defined(__ia64__)
+
+    #if defined(__INTEL_COMPILER)
+        #pragma message("The Intel Compiler crashes with _GLIBCXX_PARALLEL defined (at least until v16.0.2). Therefore, the feature will be disabled.")
+        #if __INTEL_COMPILER < 1600 || (__INTEL_COMPILER == 1600 && __INTEL_COMPILER_UPDATE <= 2)
+            #undef _GLIBCXX_PARALLEL
+        #endif
+    #endif // #if defined(__INTEL_COMPILER)
 #else
     #warning "Please enable OpenMP."
 #endif  // #ifdef _OPENMP
--- a/apps/searchjoin/join.cpp
+++ b/apps/searchjoin/join.cpp
@@ -42,6 +42,14 @@
 #ifdef _OPENMP
     #define _GLIBCXX_PARALLEL
     #include <omp.h>
+
+    #if defined(__INTEL_COMPILER)
+        #pragma message("The Intel Compiler crashes with _GLIBCXX_PARALLEL defined (at least until v16.0.2). Therefore, the feature will be disabled.")
+        #if __INTEL_COMPILER < 1600 || (__INTEL_COMPILER == 1600 && __INTEL_COMPILER_UPDATE <= 2)
+            #undef _GLIBCXX_PARALLEL
+        #endif
+    #endif // #if defined(__INTEL_COMPILER)
+
 #else
     #if !defined(SEQAN_IGNORE_MISSING_OPENMP) || (SEQAN_IGNORE_MISSING_OPENMP == 0)
         #pragma message("OpenMP not found! Shared-memory parallelization will be disabled in join tool.")
--- a/tests/journaled_string_tree/test_journaled_string_tree_traverser.h
+++ b/tests/journaled_string_tree/test_journaled_string_tree_traverser.h
@@ -338,12 +338,15 @@ SEQAN_DEFINE_TEST(test_journaled_string_tree_traverser_basic_traversal)
 
     while (!atEnd(sub))
     {
-        auto pos = position(sub);
-        for (auto p : pos)
+        auto const & pos = position(sub);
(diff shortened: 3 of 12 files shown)


apps/fiona/fiona.cpp
```diff
@@ -81,9 +81,15 @@
     // Thus, we do not enable it in this case.
         #define _GLIBCXX_PARALLEL
     #endif  // #if !defined(__MINGW32__) || defined(__amd64__) || defined(__x86_64__) || defined(__ia64__)
+
+    #if defined(__INTEL_COMPILER)
+        #warning "The Intel Compiler crashes with _GLIBCXX_PARALLEL defined (at least until v16.0.2). Therefore, the feature will be disabled."
+        #if __INTEL_COMPILER_BUILD_DATE <= 20160204
```
h-2: Is COMPILER_BUILD_DATE the right macro to check version? I know that for GCC it is not, because 5.3.2 might be released after 6.0.0...

apps/searchjoin/join.cpp
```diff
@@ -42,6 +42,14 @@
 #ifdef _OPENMP
     #define _GLIBCXX_PARALLEL
     #include <omp.h>
+
+    #if defined(__INTEL_COMPILER)
+        #warning "The Intel Compiler crashes with _GLIBCXX_PARALLEL defined (at least until v16.0.2). Therefore, the feature will be disabled."
```
h-2: also warnings are actually critical in our nightly-build system so tests for these will still fail..

tests/journaled_string_tree/test_journaled_string_tree_traverser.h
```diff
@@ -339,11 +339,14 @@ SEQAN_DEFINE_TEST(test_journaled_string_tree_traverser_basic_traversal)
     while (!atEnd(sub))
     {
         auto pos = position(sub);
-        for (auto p : pos)
+        for (auto it = begin(pos), itEnd = end(pos); it != itEnd; ++it)
+        {
+            auto p = *it;
```
h-2: Do you know that p is an int or similar? Otherwise please use `auto &`

util/cmake/FindSeqAn.cmake
```diff
@@ -152,6 +152,13 @@ elseif (COMPILER_IS_CLANG)
         message(AUTHOR_WARNING "Clang version (${CMAKE_CXX_COMPILER_VERSION}) should be at least 3.5! Anything below is untested.")
     endif ()
 
+elseif (COMPILER_IS_INTEL)
+
+    # require at least clang 3.5
```
h-2: wrong comment ;-)

h-2:
Thanks for the PR, looking forward to trying out Intel.

apps/searchjoin/join.cpp
marehr: will changing that to `#pragma message` be okay?

apps/fiona/fiona.cpp
marehr: I looked at `icpc -dM -E - < /dev/null` and that was the only reasonable way to determine the version.

Maybe you find a better macro :+1:

tests/journaled_string_tree/test_journaled_string_tree_traverser.h
marehr: I'll change that

util/cmake/FindSeqAn.cmake
marehr: good catch ;)

apps/searchjoin/join.cpp
marehr: No, this warning would only be issued for future versions of icc and is a clear way to say "Hello future programmer please reevaluate this issue if its still there."

apps/fiona/fiona.cpp
```diff
@@ -81,9 +81,15 @@
     // Thus, we do not enable it in this case.
         #define _GLIBCXX_PARALLEL
     #endif  // #if !defined(__MINGW32__) || defined(__amd64__) || defined(__x86_64__) || defined(__ia64__)
+
+    #if defined(__INTEL_COMPILER)
+        #pragma message("The Intel Compiler crashes with _GLIBCXX_PARALLEL defined (at least until v16.0.2). Therefore, the feature will be disabled.")
```
h-2: the pragma should be inside the conditional or not?

apps/fiona/fiona.cpp
```diff
@@ -81,9 +81,15 @@
...
+        #if __INTEL_COMPILER < 1600 || (__INTEL_COMPILER == 1600 && __INTEL_COMPILER_UPDATE <= 2)
+            #undef _GLIBCXX_PARALLEL
+        #endif
+    #endif // #if defined(__INTEL_COMPILER)
 #else
     #warning "Please enable OpenMP."
-    #define omp_get_wtime() 0
```
h-2: this is not needed anymore?

tests/platform/test_platform_range_based_for_loops.h
```diff
@@ -0,0 +1,113 @@
...
+    SEQAN_ASSERT_EQ(hits[2], 'r');
+    SEQAN_ASSERT_EQ(hits[3], 'l');
+    SEQAN_ASSERT_EQ(hits[4], 'd');
+}
+
+SEQAN_DEFINE_TEST(test_platform_range_based_for_loops1) {
+    foobar1();
+}
+
+
```
h-2: remove double newline

tests/platform/test_platform_range_based_for_loops.h
```diff
@@ -0,0 +1,113 @@
...
+            SEQAN_SKIP_TEST;
+        #else
+            #warning "The Intel Compiler has a bug for this type of range-based for loop (at least until v16.0.2). Please reevaluate this issue for newer versions."
+        #endif
+    #endif
+
+    foobar2<true>();
+}
+
+
```
h-2: remove double newline

apps/fiona/fiona.cpp
marehr: This macro and the stub method in /include/seqan/parallel/parallel_macros.h
for omp_get_wtime have a name conflict.

So there is a stub method for the case that omp is not available

apps/fiona/fiona.cpp
marehr: No, because on update 16.0.3 I wouldn't be notified.

If they fix it in 16.0.3 then yes. Because then it is only a hint for older compiler versions.

For the meantime it should always notifiy me.

Issue #1678, opened 2016-04-21
Infixes of Containers of Containers are not Containers of Containers (but should be)


h-2:
pretty much as the title states:
if you have a `StringSet<CharString> s;` then the `ValueType<Infix<StringSet<CharString>>>` should be `CharString`, but instead it is `char`.

rrahn:
Ok, if I am not mistaken only the index module is using infix on StringSets so far.
In this context, however, not an infix of the StringSet, but of some string within the set is requested. This causes major confusion about what an infix of a StringSet should be, because the other way around is a valid assumption.

So here is my proposal:

- [ ] Make ```infix(set, beginPos, endPos)``` and ```infixWithLength(set, beginPos, length)``` behave as standard infix for strings. So it returns a subset of the StringSet specified by the begin and end position (the same for Prefix and Suffix).
- [ ] Add a new convenience function ```stringInfix``` for the index module that models the current ```infix``` and ```infixWithLength``` implementation for StringSets. So it explicitly returns a string infix not a string set infix.

@weese  @esiragusa Or am I missing something with the above mentioned design decisions?

esiragusa:
Do ```beginPos, endPos``` in ```infix(set, beginPos, endPos)``` refer to ```concat(set)```? This disambiguation seems reasonable, although it's not clear to me how StringSet fulfills the container concept.

rrahn:
Hmm, if I get you right, you mean if the given positions are global positions?
As far as I could see (from debugging some index tests) it is a local position (pair of seq id and offset).
Clearly, an infix spanning over multiple sequences does not make sense. So the infix implementations of StringSets merely assemble some sort of convenience functions to access the string content of some string in the set and is only used during index lookups. 
I guess whether StringSets fulfil the container concept is a philosophical question, but not the real problem here. I guess the problem is simply, that the current infix implementation is rather more a implementation detail, than a public interface. In fact, this leads to the current confusion with the usage of Infix for StringSets. Hence, I propose a different interface, which explicitly asks for a string infix and in the case of StringSets requires the StringSetPosition type. However, it would break API 😨 . That's why I am so precautious about it.

h-2:
> I guess whether StringSets fulfil the container concept is a philosophical question

Not really: https://github.com/seqan/seqan/blob/master/include/seqan/sequence/string_set_base.h#L373

> However, it would break API :fearful: . That's why I am so precautious about it.

[...]

PR #1684, opened 2016-04-28
[FIXED-#1678] Fixes issue with infix of string sets.

rrahn:
### Changes:
* Removed overloaded infix functions for StringSets.
* Replaced old behaviour with stringInfix | stringPrefix | stringSuffix interfaces.
* Introduced StringInfix | StringPrefix | StringSuffix Metafunction to obtain correct segment type.
* Replaced all instances of segment functions in apps and the library with the new interface.
* Added tests for standard segment and string segment functions.

Compiled and run all tests and app tests with g++-5 successfully.


rrahn:
I assigned it to @esiragusa because I touched a many index files and yara and fiona sources, but @h-2 you can also have a look.
This is a first design implementation, that fixes the Infix issue with StringSets.
Question remains, if we should make the new interface private and whether there is a better function name.

esiragusa:
Hmm I think that the current interface is designed to be generic for Indexes over Strings or StringSets, as described [here](http://seqan.readthedocs.io/en/master/Tutorial/DataStructures/Indices/IndexIterators.html) for example. Is this change really necessary?

rrahn:
Well, yes and no.
At the moment their is no documentation, that describes the "real" functionality of infix over StringSets, which looks like this: `infix(set, <seqId, seqOffB>, <seqId, seqOffE>)`.
So given the API doc, the following code does not behave as expected:
```cpp
StringSet set;
appendValue(set, "string1");
appendValue(set, "abcdefg");

auto inf = infix(set, 0, 1);
std::cout << inf[0] << std::endl;  // Should be "string1" but is "s".
```
However, only the indices are using the current version of infix, which is just a wrapper for `infix(set[seqId], seqOffB, seqOffE); `. 
Hence, the idea to put the convenience function into a private function and let infix work as expected and documented.
However, another idea would be to overload the infix function to get either a local StringSetPosition (Pair) or a default Position type. With C++11, we could also drop the Infix meta-function, because we could deduce the correct type from the return value.
What do you think?

h-2:
@esiragusa but the current infix interface does not behave as expected for containers of containers. So it might have been designed in a way that makes it convenient for index creation, but it fulfills no other purpose and is plain confusing (it took me very long to find the issue with the following:
`lengthSum(infix(myStringSet, 4, 7))` It will always be 3, but it should be the sum of the lengths of strings 4 to 6.
If I trip over things like this, I am sure many of our users will as well.

rrahn:
@h2 @esiragusa What is the current state about this?
My opinion is clearly this:
```
auto inf = infix(set, posB, posE)  // Returns infix over set.
```
and
```
auto inf(set, Pair<seqId, offB>, Pair<seqId, offE>)  // Returns the desired infix of seqId from the set
```
However, this makes the Infix Metafunction over StringSets ambiguous and only auto could be used in this context.
We could also have another Metafunction ```StringInfix```, which will be the correct infix over the containers in the sets, and for strings it defaults to ```Infix```.

But I really need this now.

rrahn:
@h-2 ping

apps/yara/find_extender.h
```diff
@@ -58,9 +58,9 @@ struct Extender
 template <typename THaystack, typename TNeedle, typename TSpec>
 struct Extender<THaystack, TNeedle, HammingDistance, TSpec>
 {
-    typedef typename Infix<THaystack const>::Type       THaystackInfix;
+    typedef typename InfixOnValue<THaystack const>::Type THaystackInfix;
     typedef ModifiedString<THaystackInfix, ModReverse>  THaystackInfixRev;
-    typedef typename Infix<TNeedle const>::Type         TNeedleInfix;
+    typedef typename InfixOnValue<TNeedle const>::Type   TNeedleInfix;
     typedef ModifiedString<TNeedleInfix, ModReverse>    TNeedleInfixRev;
```
h-2: types not aligned correctly

apps/yara/find_verifier.h
```diff
@@ -48,15 +48,15 @@ using namespace seqan;
...
-    typedef String<GapAnchor<int> >                 TGapAnchors;
-    typedef AnchorGaps<TGapAnchors>                 TAnchorGaps;
-    typedef typename Size<THaystackInfix>::Type     TSize;
-    typedef typename Position<THaystackInfix>::Type TPosition;
-    typedef TraceSegment_<TPosition, TSize>         TTraceSegment;
-    typedef String<TTraceSegment>                   TTrace;
-    typedef DPScoutState_<Default>                  TDPState;
-    typedef DPContext<int, AffineGaps>              TDPContext;
+    typedef typename InfixOnValue<THaystack const>::Type THaystackInfix;
+    typedef String<GapAnchor<int> >                     TGapAnchors;
```
h-2: same here

include/seqan/index/index_esa_stree.h
```diff
@@ -1911,7 +1911,7 @@ SEQAN_CONCEPT_IMPL((Index<TText, IndexEsa<TSpec> > const), (StringTreeConcept));
         TSize &lcp)
     {
         typedef typename Fibre<TIndex, FibreText>::Type const        TText;
-        typedef typename Infix<TText>::Type                            TInfix;
+        typedef typename InfixOnValue<TText>::Type                    TInfix;
         typedef typename Iterator<TInfix, Standard>::Type            IText;
```
h-2: strange alignment again

include/seqan/sequence/string_set_base.h
```diff
@@ -300,41 +300,39 @@ struct Size< StringSet< TString, TSpec > >
...
-// TODO(holtgrew): However, if this works nicely then it shows that implementing segments as Strings would not be advantageous since they now work for arbitrary sequential-access containers.
 
 template <typename TString, typename TSpec>
-struct Prefix< StringSet< TString, TSpec > >
+struct PrefixOnValue< StringSet< TString, TSpec > >
     : Prefix<TString > {};
 
 template <typename TString, typename TSpec>
-struct Prefix<StringSet< TString, TSpec > const>
+struct PrefixOnValue<StringSet< TString, TSpec > const>
```
h-2: This still look strange to me, but I don't remember if a const StringSet implies const values...

include/seqan/sequence/string_set_base.h
```diff
@@ -598,11 +596,12 @@ inline void posLocalize(TResult & result, Pair<T1, T2, TPack> const & pos, Strin
...
 // --------------------------------------------------------------------------
-// Function prefix()
+// Function prefix(); For local string set position.
 // --------------------------------------------------------------------------
 
 template < typename TString, typename TSpec, typename TPosition >
-inline typename Prefix<TString>::Type
+inline SEQAN_FUNC_DISABLE_IF(Is<IntegerConcept<TPosition> >,
+                             typename PrefixOnValue<StringSet< TString, TSpec > >::Type)
 prefix(StringSet< TString, TSpec > & me, TPosition const & pos)
```
h-2: can you not use a Pair<> here instead of the SFINAE?

include/seqan/sequence/string_set_base.h
```diff
@@ -694,35 +698,37 @@ infixWithLength(StringSet< TString, TSpec > const & me, TPosition const & pos, T
...
+infix(StringSet<TString, TSpec > & me, TPosBeg const & posBeg, TPosEnd const & posEnd)
 {
     typedef StringSet<TString, TSpec>               TStringSet;
     typedef typename Size<TStringSet>::Type         TSetSize;
     typedef typename Size<TString>::Type            TStringSize;
     typedef Pair<TSetSize, TStringSize, Pack> TPair;
 
     TPair localPosBegin, localPosEnd;
-    posLocalize(localPosBegin, posBegin, stringSetLimits(me));
+    posLocalize(localPosBegin, posBeg, stringSetLimits(me));
```
h-2: not necessary

tests/modifier/test_modifier_view.h
```diff
@@ -385,15 +385,15 @@ SEQAN_DEFINE_TEST(test_modifier_convert_in_place)
     // Non-const variant on segment.
     {
         CharString strCopy = originalStr;
-        Segment<CharString, InfixSegment> stringInfix(strCopy);
+        Segment<CharString, InfixSegment> infix(strCopy);
```
h-2: where do these come from? I thought you had invented the `stringInfex()`

rrahn: Seems like this was a name of the variable 😆

h-2:
LGTM [build failure unrelated]

esiragusa:
Hey, how is the new API then? I don't see any updates to the doc nor to the manual...

h-2:
the new API is basically as the old and current doc is:
```infix(container, beg, end)``` gives an infix on the container from [beg, end)
Additionally there is the (undocumented):
```infix(container, pair<i,beg>, pair<i,end>)``` which forwards the infix function to the i-th member of container IFF container is more than one-dimensional, e.g. a StringSet
The second interface does not return the Infix-type, but the new InfixOnValue-Type.

PR #1688, opened 2016-05-03
[FIX] Visual Studio 2015 Update 2 Bugfix

### LINKED ISSUE #1662: VS2015u2 introduced lots of errors
Apparently related to our STL adaptation. It looks like the windows stl is already adopting the global function interface and is now conflicting with our adaptation.

The annoying thing is that this happens inbetween minor updates so we need to figure out the right `#ifdef` to only (but then always) define our custom overloads.
Maybe the best way would be to do some (additional) SFINAE check that just looks for `std::insert` and doesn't define our function if that exists.

cpockrandt:
Fixes #1662 

Tested on klette (on develop only) and it worked :smiley: (see cdash)

### DIFF
--- a/include/seqan/basic/iterator_adapt_std.h
+++ b/include/seqan/basic/iterator_adapt_std.h
@@ -59,7 +59,8 @@ namespace std
 	// there is a bug in vc2015 stl, it doesnt check the iterator_traits correctly
 	// I have reported this bug to microsoft already (filed as DevDiv#1208117). 
 	// For now, this is a workaround.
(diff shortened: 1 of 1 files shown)


h-2:
I will merge this, but please also create a PR against master.

Please also look at these other errors on windows:
http://cdash.seqan.de/viewTest.php?onlyfailed&buildid=176035

cpockrandt:
According to your last mail we shell only create PR against develop (even for bug fixes), or did I misunderstand it?

PR #1693, opened 2016-05-06
[FEATURE] add dynamically choosable amino acid score matrix

h-2:
adds a matrix type that can by set to e.g. Blosum62 or Pam250 **at runtime**


include/seqan/score/score_matrix_dyn.h
```diff
@@ -0,0 +1,239 @@
...
+                   TagList<ScoreSpecBlosum45,
+                   TagList<ScoreSpecBlosum62,
+                   TagList<ScoreSpecBlosum80,
+                   TagList<ScoreSpecPam40,
+                   TagList<ScoreSpecPam120,
+                   TagList<ScoreSpecPam200,
+                   TagList<ScoreSpecPam250,
+                   TagList<ScoreSpecVtml200
+                   > > > > > > > > >;
+}
```
rrahn: Please add a comment to indicate which namespace was closed here.

include/seqan/score/score_matrix_dyn.h
```diff
@@ -0,0 +1,239 @@
...
+template <typename TCurTag,
+          typename TRestList,
+          typename TRunnable>
+inline void
+matrixTagDispatch(TagList<TCurTag, TRestList> const &,
+                  AminoAcidScoreMatrixID const m,
+                  TRunnable const & runnable)
+{
+    using TCurList = TagList<TCurTag, TRestList>;
+    if (LENGTH<MatrixTags>::VALUE  - LENGTH<TCurList>::VALUE == static_cast<uint8_t>(m))
```
rrahn: Please, don't use a new enum class.
You can use the Find metafunction and the TagSelector class in fundamental_tags.h:  https://github.com/seqan/seqan/blob/develop/include/seqan/basic/fundamental_tags.h#L481
In this case you don't need to worry about the correct order of the tags in the TagList and the enum, which is very evil.
```cpp
if (value(selector) == Find<TagList, ScoreSpecPam120>::VALUE)
```

Then you just pass the ScoreSpec tag instead of a new id, which is not necessary.

h-2:
@rrahn finally :heavy_check_mark:  can you review / merge?

include/seqan/blast/blast_statistics.h
```diff
@@ -555,6 +573,35 @@ _selectSet(BlastScoringScheme<TScore> & scheme)
...
+_selectSet(BlastScoringScheme<Score<TValue, ScoreMatrix<AminoAcid, TSpec>>> & scheme)
+{
+    using TScore = Score<TValue, ScoreMatrix<AminoAcid, TSpec>>;
+    using TKAValues = KarlinAltschulValues<TScore>;
+    return _selectSet<TScore, TKAValues>(scheme);
+}
+
+template <typename TValue>
+inline bool
+_selectSet(BlastScoringScheme<Score<TValue, ScoreMatrix<AminoAcid, ScoreSpecSelectable>>> & scheme)
```
rrahn: I know that with C++11 the spaces between closing angle brackets are optional, but yet our coding guidelines suggests the whitespaces in-between. I also do not prefer the one or the other scheme, but it would be better to make it consistent with the current code. We can adapt that for seqan3.

include/seqan/blast/blast_statistics.h
```diff
@@ -555,6 +573,35 @@ _selectSet(BlastScoringScheme<TScore> & scheme)
...
+{
+    using TScoreOrig = Score<TValue, ScoreMatrix<AminoAcid, ScoreSpecSelectable>>;
+    bool ret = false;
+    impl::score::matrixTagDispatch(impl::score::MatrixTags(),
+                                   getScoreMatrixId(seqanScheme(scheme)),
```
rrahn: Where does the function seqanScheme come from? Sounds strange?

include/seqan/blast/blast_statistics.h
```diff
@@ -706,6 +760,14 @@ getLambda(BlastScoringScheme<Score<int, Simple>> const & scoringScheme)
...
  */
 
+inline double
+getKappa(BlastScoringScheme<Score<int, ScoreMatrix<AminoAcid, ScoreSpecSelectable>>> const & scoringScheme)
+{
+    SEQAN_ASSERT(isValid(scoringScheme));
+    return *(scoringScheme._m + scoringScheme.parameterIndex * scoringScheme._nParams + 4);
+}
+
+
```
rrahn: Double newline.

include/seqan/blast/blast_report_out.h
```diff
@@ -247,6 +247,36 @@ _matrixName(Pam250 const & /**/)
...
+          typename TRestList>
+constexpr const char *
+_matrixNameTagDispatch(TagList<TCurTag, TRestList> const &,
+                       AminoAcidScoreMatrixID const m)
+{
+    using TUndType = std::underlying_type_t<AminoAcidScoreMatrixID>;
+    return (Find<impl::score::MatrixTags, TCurTag>::VALUE == static_cast<TUndType>(m))
+            ? _matrixName(Score<int, ScoreMatrix<AminoAcid, TCurTag>>())
+            : _matrixNameTagDispatch(TRestList(), m);
+}
```
rrahn: Why are you reimplementing this matrixTagDispatching, instead of passing a lambda function to the this function: https://github.com/seqan/seqan/pull/1693/files#diff-c5e6467d748904987ef176a8acd97606R161?

include/seqan/score/score_matrix_dyn.h
```diff
@@ -0,0 +1,240 @@
...
+ * @typedef SelectableAminoAcidMatrix
+ * @headerfile <seqan/score.h>
+ * @brief An AminoAcid score matrix that can be "specialized" at run-time
+ *
+ * @signature using SelectableAminoAcidMatrix =  Score<int, ScoreMatrix<AminoAcid, ScoreSpecSelectable> >;
+ */
+
+template <typename TValue>
+class Score<TValue, ScoreMatrix<AminoAcid, ScoreSpecSelectable> >
+ : public Score<TValue, ScoreMatrix<AminoAcid, Default> >
```
rrahn: Indentation should be 4 spaces. The colon should be one line up.

include/seqan/blast/blast_statistics.h
h-2: ok

include/seqan/blast/blast_statistics.h
h-2: same file. it is for converting a BlastScoringScheme to a SeqAn scoring scheme

include/seqan/blast/blast_statistics.h
h-2: ok

include/seqan/blast/blast_report_out.h
h-2: Because of the `constexpr const char * ` return value. This cannot be achieved with the lambda unfortunately. constexpr lambdas only work in c++17.
[Yes, I know, the constexpr won't work for the dynamic matrix anyway, however the current interface is like this now; I want to change it for SeqAn 3 and have already added a TODO]

include/seqan/score/score_matrix_dyn.h
h-2: yep

include/seqan/blast/blast_statistics.h
rrahn: If it did exist before, just leave it, but the name is kind of confusing.
Maybe something like ```toSeqanScoringScheme``` would be more telling. 😏

include/seqan/blast/blast_report_out.h
rrahn: Ok

include/seqan/blast/blast_statistics.h
h-2: *ping* [i changed the comment, maybe you didn't get notified]

rrahn: Ok, so if in the complete file everything is in this style, than leave it as is.
I also agree, that if we start a new module, within that module the scheme can be modernised, but when adding something to existing modules, those should stay consistent.

rrahn:
lgtm!

PR #1698, opened 2016-05-10
[DOC] update install instructions in manual and use shiny table

h-2:
* use much nicer table
* add gentoo and openbsd


rrahn:
lgtm

PR #1740, opened 2016-06-20
use c++14 by default and complain if it isn't set

h-2:
* do source-coded checks for c++14 and minimum compiler versions
* improve documentation for non-cmake consumers of SeqAn
* minor design fixes ans improvements to manual
* remove references to c++-standard from the cmake-manual

The last point depends on @marehr changing our cmake-module so that it automatically set the c++-standard to 14 if required. I am convinced this is the only right thing to do.

### DIFF
--- a/include/seqan/platform.h
+++ b/include/seqan/platform.h
@@ -51,6 +51,9 @@
  */
 #ifdef _MSC_VER
 #define STDLIB_VS
+#if _MSC_VER < 1900
+#error Visual Studio versions older than version 14 / "2015" are not supported. Please upgrade.
+#endif
 #endif
 
 /*!
@@ -95,6 +98,11 @@
  */
 #if defined(__GNUC__) && !defined(__ICC) && !defined(__clang__)
 #define COMPILER_GCC
+#define COMPILER_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
+#if COMPILER_VERSION < 40901
+    #warning GCC versions older than 4.9.1 are not supported, many modules will not work.
+#endif
+#undef COMPILER_VERSION
 #endif
 
 /*!
@@ -105,6 +113,9 @@
  */
 #if defined(__ICC)
 #define COMPILER_INTEL
+#if _ICC < 1600
+    #warning ICC versions older than 16 are not supported, many modules will not work.
+#endif
 #endif
 
 /*!
@@ -115,6 +126,21 @@
  */
 #if defined(__clang__)
 #define COMPILER_CLANG
+#define COMPILER_VERSION (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)
+#if COMPILER_VERSION < 30500
+    #warning Clang versions older than 3.5.0 are not supported, many modules will not work.
+#endif
+#undef COMPILER_VERSION
+#endif
+
+// ==========================================================================
+// C++ standard macros
+// ==========================================================================
+
+#ifndef STDLIB_VS // all Visual Studio >= 2015 compilers are c++14 by default
+#if __cplusplus < 201300
+    #error SeqAn requires C++14! You must compile your application with -std=c++14, -std=gnu++14 or -std=c++1y.
+#endif
 #endif
 
 // ==========================================================================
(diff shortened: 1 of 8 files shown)


h-2:
@rrahn I think the failure is unrelated, but can you double-check?

include/seqan/platform.h
```diff
@@ -51,6 +51,9 @@
  */
 #ifdef _MSC_VER
 #define STDLIB_VS
+#if _MSC_VER < 1900
+#error Visual Studio versions older than version 14 / "2015" are not supported. Please upgrade.
```
rrahn: Remove "Please upgrade."

include/seqan/platform.h
```diff
@@ -95,6 +98,11 @@
  */
 #if defined(__GNUC__) && !defined(__ICC) && !defined(__clang__)
 #define COMPILER_GCC
+#define COMPILER_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
+#if COMPILER_VERSION < 40901
+    #warning GCC versions older than 4.9.1 are not supported, many modules will not work.
```
rrahn: Remove "..., many modules..."

include/seqan/platform.h
```diff
@@ -105,6 +113,9 @@
  */
 #if defined(__ICC)
 #define COMPILER_INTEL
+#if _ICC < 1600
+    #warning ICC versions older than 16 are not supported, many modules will not work.
```
rrahn: See above

include/seqan/platform.h
```diff
@@ -115,6 +126,21 @@
  */
 #if defined(__clang__)
 #define COMPILER_CLANG
+#define COMPILER_VERSION (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)
+#if COMPILER_VERSION < 30500
+    #warning Clang versions older than 3.5.0 are not supported, many modules will not work.
```
rrahn: see above

manual/source/Infrastructure/Use/CustomBuildSystem.rst
```diff
@@ -17,6 +17,27 @@ You should be able to adapt the descriptions to configure your build system and/
...
+C++14 Standard
+--------------
+
+On GNU/Linux, BSD and macOS, always add ``-std=c++14`` (or a newer standard) when building on the command line.
+
+For XCode on macOS you need to set this option in the project settings.
+
+As of Visual Studio 2015 our subset of C++14 is already available in all supported compilers.
+
+
```
rrahn: empty line

manual/source/Infrastructure/Use/CustomBuildSystem.rst
```diff
@@ -17,6 +17,27 @@ You should be able to adapt the descriptions to configure your build system and/
...
+OpenMP
+------
+
+On GNU/Linux, BSD and macOS, add ``-fopenmp`` unless you are using Clang versions older than 3.8.0.
+
+For XCode on macOS OpenMP is not yet available.
+
+With Visual Studio OpenMP is switched on by default.
+
+
```
rrahn: empty line

util/makefile_project/Makefile.rules
```diff
@@ -1,5 +1,5 @@
 SRC=../src
-CXXFLAGS+=-I../../../include
+CXXFLAGS+=-I../include -std=c++14
```
rrahn: Is the include path correct?

h-2: Well it's apparently set relative to the repository which we don't want, because we don't want people building things in some sub-directory of our git-checkout. So it makes more sense to be relative to a possibly created new directory. This is also in line with the documentation which says
> we could create a directory include parallel to src, copy the release version of SeqAn into it and then change the include path of the compiler to point to this directory (value ../include).

h-2:
please hold, some more changes are coming

rrahn:
ok

h-2:
Now this does not depend on changes by @marehr anymore.

h-2:
only clang fails so its good from my pov

rrahn:
lgtm!

PR #1742, opened 2016-06-21
[INTERNAL] Fixes llvm ci builds.

rrahn:
Uses downloaded binaries of llvm toolchain to temporarily work around the unavailable llvm package provider.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,28 +32,42 @@ matrix:
       #install: export CXX="g++-6" CC="gcc-6"
 
     - os: linux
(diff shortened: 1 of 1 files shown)


h-2:
Great! But doesnt yet work for 3.6 and 3.7

Also, can you please still set the "compiler" field? It is not critical if it doesn't have a compiler and it makes reading the build status on travis easier.

rrahn:
Yes, I will check it, but close his PR for now and reopen when it's hopefully fixed 😄

PR #1741, opened 2016-06-21
[FIX,INTERNAL] split align breakpoint calculation

rrahn:
Before, the optimal trace was computed from the maximum in the last row/column.
But the obtained trace might not lead to the same trace when starting from the
best split position. This update fixes this, to only consider the optimal
trace starting from the computed split position.

See #1679 for discussion.

PR #1743, opened 2016-06-21
Feature/split align fix

rrahn:
Supersedes #1741.


h-2:
4 - test_test_align_split (Failed)
	372 - app_test_gustaf (Failed)

rrahn:
Ok, only the clang issues. Otherwise it is good from my pov.

PR #1744, opened 2016-06-22
Hot fix for travis llvm builds

rrahn:
Downloads the llvm binaries and uses them instead of using apt package manager, since llvm-toolchain-precise repo is temporarily switched off.
See here for more details: http://lists.llvm.org/pipermail/llvm-dev/2016-May/100303.html


rrahn:
Green builds 😂

h-2:
:+1:

PR #1746, opened 2016-06-22
[FIX] cmake macros broken on osx

h-2:
this should fix the broken nightlies on clang/mac.


util/cmake/FindSeqAn.cmake
```diff
@@ -174,10 +174,7 @@ endif ()
 # Require C++11
 # ----------------------------------------------------------------------------
 
-if (NOT (CMAKE_VERSION VERSION_LESS "3.1"))
-    set(CMAKE_CXX_STANDARD 14)
-    set(CMAKE_CXX_STANDARD_REQUIRED on)
-elseif (NOT MSVC) # this implies all compilers within visual studio
+if (NOT MSVC) # this implies all compilers within visual studio
```
rrahn: And on windows platforms we have no c++14 checks?
Can you update the comment above that says **Require C++11** to **C++14**.

h-2: > And on windows platforms we have no c++14 checks?

MSVC still doesn't define the macro correctly, but supports a large subset of c++14 (we will only using this subset).

PR #1747, opened 2016-06-22
[FIX] fix char 2 amino acid conversion

h-2:
[other commit is NOP because if forked from master originally]


demos/tutorial/alphabets/assignment_1_solution.cpp
```diff
@@ -16,7 +16,7 @@ void showAllLettersOfMyAlphabet(TAlphabet const &)
     // We iterate over all characters of the alphabet
     // and output them
     for (TSize i = 0; i < alphSize; ++i)
-        std::cout << i << ',' << TAlphabet(i) << "  ";
+        std::cout << static_cast<unsigned>(i) << ',' << TAlphabet(i) << "  ";
```
rrahn: Why cast?

h-2: Because the type is unsigned char which is transformed to an unprintable char upon cout. The .stdout is labeled as binary by github for exactly this reason (it had not printable characters before).

PR #1749, opened 2016-06-23
[FIX] defaulted to Exact()  CharString

### LINKED ISSUE #1653: Memory consumption explodes using Bam I/O with millions of reference sequences
It seems that our bam I/O consumes a lot of memory when dealing with Bam files containing large headers with millions of reference sequences.

temehi:
The references in bam files are now stored in  Exact()  CharString. This brings about 5X lower memory footprint for the case specified in #1653 

credit to @rrahn 

This fixes #1653


rrahn:
Should be merged into develop only.

PR #1750, opened 2016-06-23
[FIX] skip white spaces after reading a recored from Genbank file

temehi:


h-2:
we are only merging hotfixes (for release problems, regressions, security problems) to master.

temehi:
aha, I didn't know about that

PR #1752, opened 2016-06-23
[FIX] defaulted to Exact()  CharString

temehi:


rrahn:
This should not be a legal matter anyway, so I merge it without license agreement.

Issue #1686, opened 2016-05-02
index is wrong for methylation fasta


jdidion:
I am simulating bisulfite reads. To do this I use the following commands:

mason_variator -ir $REF_GENOME -ov sim.vcf -of sim.hap.fa \
  --meth-fasta-out sim.meth.fa --snp-rate 0.01 --small-indel-rate 0.001 --methylation-levels \
  --meth-cg-mu 0.8
mason_simulator --num-threads 8 -n 1000000 -o sim1.fq -or sim2.fq -oa sim.sam \
  -ir $REF_GENOME -iv sim.vcf --meth-fasta-in sim.meth.fa \
  --fragment-max-size 600 --fragment-mean-size 400 \
  --enable-bs-seq --bs-seq-conversion-rate 0.98 --illumina-read-length 150

The second command fails with error "Could not find top levels in methylation FASTA." Looking at the sim.meth.fa.fai file that mason_simulator constructed, it contains only one row for each sequence, not two rows with the expected /TOP and /BOT endings. If I use samtools faidx the correct index is built and the program runs fine.

h-2:
@holtgrewe can you have a look?

holtgrewe:
@h-2 This appears to be a bug with building the FAI index. As this worked for the initial version of Mason, I assume this broke on some I/O rewrite. Could someone of the experts with the new I/O system look at this?

h-2:
@holtgrewe Can you add an app test that you know worked before and now does not?
@temehi Can you have a look at this?

holtgrewe:
Well, the following FASTA file should be indexable by FAI and the ids should be everything before the first space (`id/with/suffix` or `another/id/with/suffix`).

```
>id/with/suffix this is a comment
CGAT
>another/id/with/suffix this is another comment
ACGT
```

temehi:
@holtgrewe  Indexing the suggested files works fine. I get the following result:
```
id/with/suffix  4       34      4       5
another/id/with/suffix  4       87      4       4

```
I have also accessed the sequences using the index without a problem.

PR #1753, opened 2016-06-24
[FIX] bug in mason2

temehi:
This fixes issue #1686. 
In mason the wrong index was being built.


h-2:
I just explained we don't pull to master anymore except in emergencies :open_mouth:

Issue #1641, opened 2016-03-31
Bidirectional FM-index search in constant time


cpockrandt:
using prefix sums in bit vectors

PR #1757, opened 2016-06-24
Feature/const dict levels ultrablocks

cpockrandt:
Still some TODOs in there but feedback is very much appreciated :smiley: 

#1641


include/seqan/basic/tuple_bit_compressed.h
```diff
@@ -84,17 +84,27 @@ template <> struct BitVector_<32> { typedef unsigned int Type; };
 template <> struct BitVector_<64> { typedef uint64_t Type; };
 template <> struct BitVector_<255>;
 
+template <typename TValue, unsigned SIZE, unsigned BITSIZE1, unsigned BITSIZE2, typename TSpec>
+struct Size<Tuple<TValue, SIZE, BitPacked<BITSIZE1, BITSIZE2, TSpec> > >
```
h-2: This didn't exist before somewhere?

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -109,42 +124,69 @@ struct Levels {};
...
-// Metafunction RankDictionaryBitMask_
+// Metafunction RankDictionaryWordSize_
 // ----------------------------------------------------------------------------
 
-template <typename TSpec>
-struct RankDictionaryBitMask_<uint32_t, TSpec>
-{
-    static const uint32_t VALUE = 0x55555555;
+template <typename TValue, typename TSpec, typename TConfig>
+struct RankDictionaryWordSize_<TValue, Levels<TSpec, TConfig> > : BitsPerValue<uint64_t> {
```
h-2: code style

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -271,21 +414,108 @@ struct RankDictionary<TValue, Levels<TSpec, TConfig> >
...
     RankDictionary() :
         _length(0)
-    {}
+    {
+        // TODO: no code duplication. make it a const-expr?
+        static const unsigned maxValue = (1 << _BITS_PER_VALUE) - 1;
+        for (unsigned i = 0; i < ValueSize<TValue>::VALUE; ++i)
+            _CHAR_BITMASKS[i] = _bitmaskWrapper<TWordType>(*this, _BITS_PER_WORD, _VALUES_PER_WORD, _BITS_PER_VALUE, maxValue-i, i + (1 << (_BITS_PER_VALUE-1)));
+        for (unsigned i = 0; i < _VALUES_PER_WORD; ++i)
+            _TRUNC_BITMASKS[i] = _bitmaskWrapper<TWordType>(*this, _BITS_PER_WORD, i+1, _BITS_PER_VALUE, 1, 1 << (_BITS_PER_VALUE-1));
```
h-2: use c++14 constexpr function with exception for COMPILER_MSVC

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -330,6 +582,28 @@ _toBlockPos(RankDictionary<TValue, Levels<TSpec, TConfig> > const & /* dict */,
 }
 
 // ----------------------------------------------------------------------------
+// Function _toSuperBlockPos()
+// ----------------------------------------------------------------------------
+
+template <typename TValue, typename TSpec, typename TConfig, typename TPos>
+inline typename Size<RankDictionary<TValue, Levels<TSpec, TConfig> > >::Type
+_toSuperBlockPos(RankDictionary<TValue, Levels<TSpec, TConfig> > const & /* dict */, TPos pos)
```
h-2: TPos const

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -580,10 +991,36 @@ _getValueRank(RankDictionary<TValue, Levels<TSpec, TConfig> > const & dict,
...
+
+    TSize valueRank = 0;
+
+    // TODO: I guess we can use the original loop since I couldn't measure any performance difference and it is easier to read
+    // NOTE(esiragusa): writing the loop in this form prevents the compiler from unrolling it.
+    //    for (TSize wordPrevPos = 0; wordPrevPos < wordPos; ++wordPrevPos)
+    //      valueRank += _getWordRank(dict, values[wordPrevPos].i, c);
+
+    for (TSize wordPrevPos = 0; wordPrevPos < TRankDictionary::_WORDS_PER_BLOCK; ++wordPrevPos)
+      if (wordPrevPos < wordPos) valueRank += _getWordRank(dict, values[wordPrevPos].i, TRankDictionary::_VALUES_PER_WORD - 1, c, smaller);
```
h-2: statement to next line

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -702,7 +1282,7 @@ inline void setValue(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TPo
 // Function appendValue()
 // ----------------------------------------------------------------------------
 // NOTE(esiragusa): Better not to have appendValue() - it is not efficient - and thus neither length().
-
+// TODO: try to remove this
```
h-2: TODO(cprockrandt):

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -714,33 +1294,127 @@ inline void appendValue(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict,
 // Function updateRanks()
 // ----------------------------------------------------------------------------
```
h-2: only specialize the loop (with an extra function)

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -775,9 +1456,50 @@ template <typename TValue, typename TSpec, typename TConfig, typename TSize, typ
...
+        result &= resize(dict.ultrablocks, (newLength + TRankDict_::_VALUES_PER_ULTRABLOCK - 1) / TRankDict_::_VALUES_PER_ULTRABLOCK, tag);
+    return result;
+}
+
+template <typename TValue, typename TSpec, typename TConfig>
+inline bool save(RankDictionary<TValue, Levels<TSpec, TConfig> > const & dict, const char * fileName, int openMode)
+{
+    String<char> name;
+    bool result = true;
+    name = fileName;        append(name, "");       result &= save(getFibre(dict, FibreRanks()), toCString(name), openMode);
```
h-2: insert linebreaks

h-2:
This looks mostly good from my point of view. @esiragusa do you have any comments? [we do need to merge this in the next days so if you do, please respond quickly, thanks :smile: ]

esiragusa:
I had only a quick look. Here some remarks:
* The code now is hardly readable. Please abstract constructs like [this](https://github.com/seqan/seqan/pull/1757/files#diff-2877502bc3ec648b114af645ee21988cR386), e.g., using metafunctions: 
* Make LevelConfig take only an unsigned parameter (being only 1, 2 or 3). Blocks size types are constrained by number of blocks and size type (see below).
* Make an array *blocks* whose size if controlled by a template integer so to have *blocks[0]* instead of *blocks*, blocks[1] instead of *superblocks* and blocks[2] for *ultrablocks*. Same for Fibre metafunction.
* What is PlusOne Tuple?
* What is dynamic mode? Why is Yara using that?

I think these are the only configurations that make sense:

|          | 0      | 1      | 2    |
| -------- | ------ | ------ | ---- |
| 1 Level  | SIZE   | -      | -    |
| 2 Levels | SIZE/2 | SIZE   | -    |
| 3 Levels | SIZE/4 | SIZE/2 | SIZE |

SIZE (32 or 64) is the size type of the underlying text.

apps/yara/misc_types.h
```diff
@@ -170,8 +170,13 @@ struct YaraFMConfig
     typedef TAlloc                                      Fibre;
     typedef TSum                                        Size;
 
+    typedef LevelConfig<Size>                           LvlConfig;
```
esiragusa: This will go away.

apps/yara/misc_types.h
```diff
@@ -170,8 +170,13 @@ struct YaraFMConfig
...
+
     // Sparse SA sampling rate.
     static const unsigned SAMPLING =                    10;
+    static const bool WORDS_PER_BLOCK_DYNAMIC_MODE =    1;
```
esiragusa: This too.

apps/yara/misc_types.h
```diff
@@ -170,8 +170,13 @@ struct YaraFMConfig
...
+    static const unsigned WORDS_PER_BLOCK =             4; // not used since WORDS_PER_BLOCK_DYNAMIC_MODE = 1
```
esiragusa: And this too.

apps/yara/misc_types.h
```diff
@@ -170,8 +170,13 @@ struct YaraFMConfig
...
+    static const unsigned LEVELS =                      1;
```
esiragusa: That is sufficient for RD configuration.

h-2:
@esiragusa  we have discussed these things here and came to the conclusion that its the best thing to do, but please sum up our points again @cpockrandt and address @esiragusa 's concerns.

On a more general note: **please adhere to our policies for labeling commits**

cpockrandt:
Let me explain the interface of the RankDictionary to make design decisions a bit clearer @esiragusa:

    RDConfig<TFibre, LevelConfig<uint16_t, uint32_t, uint64_t>, WORDS_PER_BLOCK_DYNAMIC_MODE_, WORDS_PER_BLOCK_>

``LevelConfig`` determines the data types for the max. 3 levels. A 2-level dictionary could look like ``LevelConfig<uint16_t, uint32_t>`` or ``LevelConfig<uint32_t, uint64_t>``. We don't want to use metafunctions instead because it doesn't allow different specializations in one project. Since we do not know the length of the text at compile-time, we cannot determine the data-type for the top-most level.

``DynamicMode``: We want the developer to choose how many words fit into one block (i.e. the number of popcounts per rank-query) and again want to avoid metafunctions. To keep it compatible with the previous seqan releases, the default value would have to be Sigma (alphabet size). Since ``RDConfig`` does not know the alphabet size, we introduced a tag ``DynamicWPB`` for the default behavior and ``FixedWPB`` to set it to a fixed value.

``PlusOne`` extends the Tuple to use ``log_2 sigma + 1`` bits per character (where the leftmost bit is always set to 0) allowing us to implement a unidirectional and bidirectional FM index in constant-time. I omit the details here but if you want I can send you the preliminary version of our paper.

Is there a benefit for using a single array _blocks_ or is it just a matter of taste? It might reduce the disk format to 1 instead of up to 3 files, but this way one cannot see what the type of a rank dictionary on disk is.

@h-2: I wanted to squash my commits before the PR but I forgot :no_mouth: I didn't mean my commit messages to go into the seqan repo :smile:

esiragusa:
* Just use a template unsigned for levels and a template unsigned for words per block, where 0 is dynamic. Using an array of levels makes your code cleaner (lookup can be defined recursively or iteratively).
* Of course you know the size type of the text at compile time. The suffix array needs that too...
* Plus one doesn't need to be stored in the dictionary. You can expand the words on lookup. Otherwise this will waste a lot of space.
* I don't understand your problem with metafunctions.

cpockrandt:
* I like your idea about removing the Dynamic/Fixed-Tag and using the value 0 instead.
* How can you know the size of the text at compile-time when you read it from a file at run-time?
* I agree, you could could extend PlusOne on lookup but it won't be constant-time in theory and we will most probably loose the speed-up for bidirectional indices in practice. The space paid off for the running time we measured in our tests (at least for smaller alphabets).
* The line you highlighted could be indeed rewritten as a metafunction (sorry, I thought you were referring to something else), I don't mind, but I don't see how this would make the code more readable.

h-2:
We are just discussing these changes here and have some questions.

> Just use a template unsigned for levels and a template unsigned for words per block, where 0 is dynamic

done

> Using an array of levels makes your code cleaner (lookup can be defined recursively or iteratively).

But they are of different types!? We discused putting it all in one vector, so that it would be

```
|   64bit        |      64bit       |
|----------------|--------|--------|
| superblock | block| block|
```
But it would not be very clean and would require major changes. Also it would break on-disk compatibility, because it would also affect the structure of 1-level.

esiragusa:
Sorry I meant a tuple.

esiragusa:
If Dna Rank Dictionary needs an extra carry bit (if it cannot be "unpacked" in constant time), it wastes 50% more space and looses word alignment. Why would this be appealing then? It's only saving a couple of bit operations... perhaps I am missing something.

esiragusa:
@cpockrandt For an example of how to configure the text **size type** you can look at [yara](https://github.com/seqan/seqan/blob/master/apps/yara/mapper.cpp#L345) and ask @h-2.

Please note that this strategy is common to all indices. A RD with dynamic text size type is pointless as long as other full-text indices (or other components of the FM-Index) need static text size type.

cpockrandt:
The extra carry bit is mainly for bidirectional indices. The running time for DNA is about 25% faster, for larger alphabets 2-3x faster (than WaveletTrees).

For our applications such as lncRNA search in the human genome the speed is more important than space ("wasted" by an additional bit)

esiragusa:
Thanks for explaining, but I still don't get it... I thought that cumulative rank for bidirectionality is orthogonal to rank computation. Maybe I will get a chance to discuss this live sometimes 😄

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -761,10 +1453,17 @@ length(RankDictionary<TValue, Levels<TSpec, TConfig> > const & dict)
...
 template <typename TValue, typename TSpec, typename TConfig, typename TSize, typename TExpand>
 inline typename Size<RankDictionary<TValue, Levels<TSpec, TConfig> > >::Type
-reserve(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize newCapacity, Tag<TExpand> const tag)
+reserve(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize const newCapacity, Tag<TExpand> const tag)
 {
-    return reserve(dict.ranks, (newCapacity + RankDictionary<TValue, Levels<TSpec, TConfig> >::_VALUES_PER_BLOCK - 1) /
-                               RankDictionary<TValue, Levels<TSpec, TConfig> >::_VALUES_PER_BLOCK, tag);
+    typedef RankDictionary<TValue, Levels<TSpec, TConfig> > TRankDict_;
+
+    bool res = true;
```
esiragusa: resize returns the new container size, not a bool!

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -761,10 +1453,17 @@ length(RankDictionary<TValue, Levels<TSpec, TConfig> > const & dict)
...
+    res &= reserve(dict.blocks, (newCapacity + TRankDict_::_VALUES_PER_BLOCK - 1) / TRankDict_::_VALUES_PER_BLOCK, tag);
+    if (TConfig::LEVELS > 1)
+        res &= resize(dict.superblocks, (newCapacity + TRankDict_::_VALUES_PER_SUPERBLOCK - 1) / TRankDict_::_VALUES_PER_SUPERBLOCK, tag);
```
esiragusa: Reserve or resize??

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -761,10 +1453,17 @@ length(RankDictionary<TValue, Levels<TSpec, TConfig> > const & dict)
 
 template <typename TValue, typename TSpec, typename TConfig, typename TSize, typename TExpand>
 inline typename Size<RankDictionary<TValue, Levels<TSpec, TConfig> > >::Type
-reserve(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize newCapacity, Tag<TExpand> const tag)
+reserve(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize const newCapacity, Tag<TExpand> const tag)
```
esiragusa: Note that const for arguments passed by value is superfluous according to our conventions.

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -773,11 +1472,72 @@ reserve(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize newCapacit
 
 template <typename TValue, typename TSpec, typename TConfig, typename TSize, typename TExpand>
 inline typename Size<RankDictionary<TValue, Levels<TSpec, TConfig> > >::Type
-resize(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize newLength, Tag<TExpand> const tag)
+resize(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize const newLength, Tag<TExpand> const tag)
 {
+    typedef RankDictionary<TValue, Levels<TSpec, TConfig> > TRankDict_;
+
+    bool result = true;
```
esiragusa: Again, return type is not a bool.

include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -773,11 +1472,72 @@ reserve(RankDictionary<TValue, Levels<TSpec, TConfig> > & dict, TSize newCapacit
...
+inline bool save(RankDictionary<TValue, Levels<TSpec, TConfig> > const & dict, const char * fileName, int const openMode)
+{
+    String<char> name;
+    bool result = true;
+
+    name = fileName;
+    append(name, "");
+    result &= save(getFibre(dict, FibreRanks()), toCString(name), openMode);
+
+    if (TConfig::LEVELS > 1)
```
esiragusa: Are blocks going to be refactored with a tuple? If so, this goes away.

apps/yara/misc_types.h
```diff
@@ -172,6 +172,8 @@ struct YaraFMConfig
 
     // Sparse SA sampling rate.
     static const unsigned SAMPLING =                    10;
+    static const unsigned WORDS_PER_BLOCK =             0;
```
esiragusa: Yara (and the old FM-index) used 2 words per block. Please keep binary compatibility. Are you planning to remove (or at least disable) dynamic size?

apps/yara/misc_types.h
```diff
@@ -172,6 +172,8 @@ struct YaraFMConfig
...
+    static const unsigned LEVELS =                      1;
```
esiragusa: Just curious, do you know if 2-Levels FM-index is equal/faster/slower than 1-Level?

include/seqan/index/index_fm_rank_dictionary_levels.h
h-2: We are moving more to the convention of "use const whenever possible" without exceptions. It may be pedantic, but it increases readability and consistency.

h-2:
@cpockrandt can you address the remaining questions of @esiragusa?

apps/yara/misc_types.h
cpockrandt: Yes, 2-Levels are a bit faster than 1-Levels and 3-Levels

apps/yara/misc_types.h
cpockrandt: I think for the next major release we will remove dynamic size.

h-2:
@cpockrandt @esiragusa is this mergeable from your POV? Then I will do a final review and merge if I find no issues.

esiragusa:
@h-2 If Yara's index config uses static size RD with 2 words per block, then it is binary compatible so this PR is mergeable for me. Then you decide on my design suggestions.

esiragusa:
For dynamic size RDs, I think that you should keep the feature (unless there's a disadvantage). But I would make indexes use static size config by default at least because this keeps binary compatibility.

cpockrandt:
@esiragusa I wanted to make the _words_per_block_ static for Yara but since it differs for Sentinels (bool -> 2) and BWT (Dna -> 4) and YaraFMConfig is recursive, I didn't see an easy way to do that (without messing up YaraFMConfig  :laughing: )

h-2:
@cpockrandt will you create a new PR for the changes regarding the "orientation" of the sequences? Then I will merge this now so that we get nightly results on it...

cpockrandt:
@h-2 Yes, you can merge. Regarding the orientation of sequences: Constructing an FMIndex on a ModString when a normal string is given, caused problems again with the Holder-thingy :grimacing: I didn't manage to make it work yet :D

I came up with some thoughts about this issue:
* we shouldn't tell the index whether we want a prefix or suffix trie, but rather the iterator. Since the index can be created without the iterator, there is no simple solution, maybe we should discuss this regarding Seqan 3
* To your original concern: it is clearly documented that an FMIndex can only be traversed as a prefix trie and not as a suffix trie (in readthedocs and thanks to @smehringer in the api dox soon as well). Developers who don't want to be bothered with these details, will/should use the Finder class anyway, which works with suffix and prefix tries without any further configuration.

esiragusa:
@h-2 @cpockrandt can you please give me an update on binary compatibility?

h-2:
@esiragusa Both the current default FM-Index and Yara's config should be entirely unchanged. Should you notice something different, please notify us!

PR #1760, opened 2016-06-27
[FEATURE] Version Check for SeqAn.

smehringer:
Implements the version check functionality inside the seqan argument parser.


h-2:
@smehringer you need to sign the clahub!

demos/my_demo_clp.cpp
```diff
@@ -0,0 +1,55 @@
+#include <iostream>
+
```
rrahn: Remove empty line.

demos/my_demo_clp.cpp
```diff
@@ -0,0 +1,55 @@
...
+    if (res != seqan::ArgumentParser::PARSE_OK)
+        return res;
+
+    // Extract option values.    
+    getOptionValue(options.period, parser, "period");
+
+    return seqan::ArgumentParser::PARSE_OK;
+}
+
+int main(int argc, char const ** argv)
```
rrahn: Should this entire file go away? Because we cannot test it easily if it requires arguments.

include/seqan/arg_parse/arg_parse_doc.h
```diff
@@ -249,6 +249,45 @@ inline CharString getShortDescription(ArgumentParser const & me)
...
+ * @brief Returns the url string.
+ *
+ * @signature TCharStringRef getUrl(parser);
+ *
+ * @param[in,out] parser The ArgumentParser to get the url string from.
+ *
+ * @return TCharString A const-ref to a @link CharString @endlink with the url string.
+ */
+
+inline CharString const & getUrl(ArgumentParser const & me)
```
rrahn: Why setting with std::string and returning CharString?
Make both to CharString.

include/seqan/arg_parse/arg_parse_parse.h
```diff
@@ -307,6 +308,31 @@ ArgumentParser::ParseResult parse(ArgumentParser & me,
         return ArgumentParser::PARSE_ERROR;
     }
 
+#if _SEQAN_VERSION_CHECK == 1
```
rrahn: Remove ```== 1```

include/seqan/arg_parse/arg_parse_parse.h
```diff
@@ -307,6 +308,31 @@ ArgumentParser::ParseResult parse(ArgumentParser & me,
...
+    // do version check if not turned off by the user
+    std::string check = "DEV"; // default
+    if(isSet(me, "version-check"))
+        getOptionValue(check, me, "version-check");
```
rrahn: Why getting the version check value if version-check is already set?
Can we not just use this as a flag option?
Like so:     
```
addOption(parser, seqan::ArgParseOption("", "version-check", "Check for updates."));
...
bool doVersionCheck = isSet(me, "vc");
```
And then only use doVersionCheck instead of testing against an expected string.

include/seqan/arg_parse/arg_parse_parse.h
```diff
@@ -307,6 +308,31 @@ ArgumentParser::ParseResult parse(ArgumentParser & me,
...
+
+    if (!(check=="OFF"))
+    {
+        VersionCheck app_version(me.appVersionCheckFuture,
+                                 toCString(me._toolDoc._name),
+                                 toCString(me._toolDoc._version),
+                                 toCString(me._toolDoc._url));
+        checkForNewerVersion(app_version);
+
+        if (!(check=="APP_ONLY"))
```
rrahn: Ah ok, so version-check can have multiple values and is not just a switch.

smehringer: Yes, it can have different values. 
The user can decide whether he wants no notifications at all (OFF), all notifications (DEV - default) or only the app notifications and no "new seqan version" info (APP_ONLY).

include/seqan/arg_parse/arg_parse_doc.h
smehringer: I changed the setUrl function.

include/seqan/arg_parse/arg_parse_parse.h
```diff
@@ -307,6 +308,31 @@ ArgumentParser::ParseResult parse(ArgumentParser & me,
         return ArgumentParser::PARSE_ERROR;
     }
 
+#if _SEQAN_VERSION_CHECK
+    // do version check if not turned off by the user
+    std::string check = "DEV"; // default
```
rrahn: This should be modelled with an enum.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+	std::string _os = "MacOS";
+#elif defined(PLATFORM_WINDOWS)
+	std::string _os = "Windows";
+#elif __FreeBSD__
+    std::string _os = "FreeBSD";
+#elif __OpenBSD__
+    std::string _os = "OpenBSD";
+#else
+    std::string _os = "unknown";
+#endif
```
rrahn: Better to call constexpr function here.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+    // ----------------------------------------------------------------------------
+    // Member Variables
+    // ----------------------------------------------------------------------------
+    std::string _url = "http://www.seqan.de/version_check/SeqAn_";
+    std::string _name;
+    std::string _version = "0.0.0";
+    std::string _program;
+    std::string _command;
+    std::string _website = "https://github.com/seqan/seqan/tree/master/apps";
+    std::future<bool> & _fut;
```
rrahn: I think it is better handled with a std::promise.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+    VersionCheck(std::future<bool> & fut,
+                 std::string const & name,
+                 std::string const & version,
+                 std::string const & website):
+    _fut(fut)
+    {
+        _name = name;
+        if (!version.empty() &&
+            std::regex_match(version, std::regex("^[[:digit:]]\\.[[:digit:]]\\.[[:digit:]].*")))
+            _version = version.substr(0,5); // in case the git revision number is given take only version number
```
rrahn: This is prone to failure, if there is a two digit version number.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+        else if (!system("which ftp > /dev/null 2>&1"))
+            _program =  "ftp -Vo";
+#endif
+        else
+            _program.clear();
+    }
+#else // windows
+    void _getProgram()
+    {
+        _program = "powershell.exe -NoLogo -NonInteractive -Command \"& {Invoke-WebRequest -erroraction 'silentlycontinue' -OutFile";
```
rrahn: Is that true on all windows platforms?

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+        if (!version.empty() &&
+            std::regex_match(version, std::regex("^[[:digit:]]\\.[[:digit:]]\\.[[:digit:]].*")))
+            _version = version.substr(0,5); // in case the git revision number is given take only version number
+        if(!website.empty())
+            _website = website;
+        _getProgram();
+        _updateCommand();
+    }
+
+    VersionCheck(VersionCheck const & rhs):
```
rrahn: I think the class should not be copy constructible if it holds a reference to a future. 
Otherwise it is possible to write to the same future from multiple threads without knowing.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+{
+    std::swap(me._url, url);
+    me._updateCommand();
+}
+
+// ----------------------------------------------------------------------------
+// Function _checkWritability()
+// ----------------------------------------------------------------------------
+#if defined(PLATFORM_WINDOWS)
+inline bool _checkWritability(std::string path)
```
rrahn: Rather pass by const-ref here?

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+
+    if (dummyFile == INVALID_HANDLE_VALUE)
+        return false;
+    
+    CloseHandle(dummyFile);
+    DeleteFile(fileName.c_str());
+    return true;
+}
+#else
+inline bool _checkWritability(std::string path)
```
rrahn: const-ref?

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+
+    return true;
+}
+#endif
+
+// ----------------------------------------------------------------------------
+// Function _getFileTimeDiff()
+// ----------------------------------------------------------------------------
+#if defined(PLATFORM_WINDOWS)
+inline double _getFileTimeDiff(std::string version_file)
```
rrahn: const ref?

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+    ul_file.LowPart  = hFileTime.dwLowDateTime;
+    ul_file.HighPart = hFileTime.dwHighDateTime;
+
+    ULONGLONG diffInTicks = ul_curr.QuadPart - ul_file.QuadPart; // get time difference
+    LONGLONG diffInS = diffInTicks / 10000000;
+    CloseHandle(hFile);
+    return((double)diffInS);
+}
+#else
+inline double _getFileTimeDiff(std::string version_file)
```
rrahn: const-ref?

include/seqan/arg_parse/arg_parse_version_check.h
rrahn: Is it not easier to write the timestamp into the file, without all this kernel voodoo?

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+// ----------------------------------------------------------------------------
+inline bool _readVersionNumbers(seqan::String<int> & version_numbers, std::string const & version_file)
+{
+    std::ifstream myfile;
+    myfile.open(version_file.c_str());
+    std::string line;
+    if (myfile.is_open())
+    {
+        std::getline(myfile,line); // get first line which should only contain the version number
+        if (std::regex_match(line, std::regex("^[[:digit:]]\\.[[:digit:]]\\.[[:digit:]]$")))
```
rrahn: Only matches numbers with single digits not multiple.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+        {
+            _getNumbersFromString(version_numbers, line);
+            myfile.close();
+        }
+        else if (line == "UNREGISTERED APP")
```
rrahn: Please use a constexpr.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+
+// ----------------------------------------------------------------------------
+// Function _getVersionNumbers()
+// ----------------------------------------------------------------------------
+inline bool _readVersionNumbers(seqan::String<int> & version_numbers, std::string const & version_file)
+{
+    std::ifstream myfile;
+    myfile.open(version_file.c_str());
+    std::string line;
+    if (myfile.is_open())
```
rrahn: Maybe:
```cpp
if (!myfile.is_open())
    return false;

bool state = false;
// check regex and set state to true if successful ...
// otherwise print comment if unregistered app.
myfile.close();
return state;
```

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+    
+    if (version_file_exists) // file exists. TODO:: ask if there should be a time limit here too
+    {
+        seqan::String<int> new_ver;
+        if (_readVersionNumbers(new_ver, version_file))
+        {
+            seqan::String<int> old_ver;
+            _getNumbersFromString(old_ver, me._version);
+
+            if (_isSmaller(old_ver, new_ver))
```
rrahn: You could use SeqAn's  Comparator instead of writing your own method.
This stores the result and you have to parse only once. See here the documentation: http://docs.seqan.de/seqan/develop/?p=Lexical
You can then ask the comparator object ```isLess() ``` or ```isGreater()``` for the result.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+inline bool _callServer(VersionCheck me, bool version_file_exists)
+{
+    // system call
+    // http response is stored in a file '.config/seqan/{app_name}_version'
+    if (system(me._command.c_str()))
+    {
+        std::string version_file = me._path + "/" + me._name + "_version.txt";
+        if (!version_file_exists)
+        {
+            std::ofstream version_file_out (version_file.c_str());
```
rrahn: no space between var name and arguments.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+            if (version_file_out.is_open())
+            {
+                version_file_out << "UNABLE TO CALL HOME.\n";
```
rrahn: constexpr

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+                version_file_out.close();
+            }
+        }
+        else // just set last modified date to current and do not override possible version information
+        {
```
rrahn: So there was no successful connection to the server, but there is already a file, so you update the timestamp of that file  to call first next day. 
But there is always a file. Then, why not storing the timestamp in the file. We can do this without the OS voodoo but with standard functions of the STL.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+        }
+        return false;
+    }
+    return true;
+}
+
+// ----------------------------------------------------------------------------
+// Function checkForNewerVersion()
+// ----------------------------------------------------------------------------
+inline bool checkForNewerVersion(VersionCheck & me)
```
rrahn: I think this entire function should run in a separate thread.
And even if the operations are quite small. There is no guarantee, that the thread is terminated before the file is written too and closed properly.
So it would be better to wrap it in a class, whose destructor can make sure, that all file handlers are closed correctly.
In case it could also utilise a rollback functionality.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+    {
+        seqan::String<int> new_ver;
+        if (_readVersionNumbers(new_ver, version_file))
+        {
+            seqan::String<int> old_ver;
+            _getNumbersFromString(old_ver, me._version);
+
+            if (_isSmaller(old_ver, new_ver))
+            {
+                if(me._name == "seqan")
```
rrahn: constexpr

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,466 @@
...
+            }
+        }
+    }
+
+    if (me._program.empty())
+        return false;
+
+    // launch a seperate thread to not defer runtime
+    // std::cout << me._command << std::endl;
+    me._fut = std::async(std::launch::async, _callServer, me, version_file_exists);
```
rrahn: Instead of using this reference to a future thing, we could use a promise object.
The promise object creates a future which is stored in the ArgumentParser and is moved to the callable invoked by the thread. Then the promise sets the bool flag to inform the future. Like in the example of: http://en.cppreference.com/w/cpp/thread/promise

include/seqan/arg_parse/tool_doc.h
```diff
@@ -649,7 +650,7 @@ class ToolDoc
         _manSection(1) {}
 
     ToolDoc(ToolDoc const & toolDoc) :
-        _name(toolDoc._name), _shortDescription(toolDoc._shortDescription),
+        _name(toolDoc._name), _shortDescription(toolDoc._shortDescription),_url(toolDoc._url),
```
rrahn: space before _url

include/seqan/arg_parse/tool_doc.h
```diff
@@ -874,6 +875,47 @@ inline CharString const & getShortDescription(ToolDoc const & doc)
...
+ *
+ * @signature void setUrl(toolDoc, str);
+ *
+ * @param[in,out] toolDoc The ToolDoc object to the set the version string for.
+ * @param[in]     str     The url string of the tool (@link CharString @endlink).
+ */
+
+inline void setUrl(ToolDoc & doc, CharString const & url)
+{
+    doc._url = url;
```
rrahn: For consistency, use copy'n'swap.
See: https://github.com/seqan/seqan/pull/1760/files#diff-5576254dc1d9732a6bc7c84aefc8b04dR186

include/seqan/arg_parse/arg_parse_version_check.h
smehringer: It is available for all windows versions that support the powershell.
If they do not have the powershell, there is no way to send a http request (as far as i researched it) so we will not be able to check the version then. (The error is redicrected to `null` so the user does not notice anything.)

include/seqan/arg_parse/arg_parse_version_check.h
smehringer: The shown code snippet is outdated.

include/seqan/arg_parse/arg_parse_version_check.h
xenigmax: According to this, https://technet.microsoft.com/library/hh849901
Invoke-WebRequest has been supported since 3.0.

According to this, https://en.wikipedia.org/wiki/Windows_PowerShell 
Powercell 3.0 has been integrated since Windows 8.
(But Windows 7 / Server 2008 users can use it with a service pack.

Just for your information.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,406 @@
...
+    void _getProgram()
+    {
+        // ask if system call for version or help is successfull
+        if (!system("wget --version > /dev/null 2>&1"))
+            _program = "wget -q -O";
+        else if (!system("curl --version > /dev/null 2>&1"))
+            _program =  "curl -o";
+#ifndef __linux  // ftp call does not work on linux
+        else if (!system("which ftp > /dev/null 2>&1"))
+            _program =  "ftp -Vo";
```
xenigmax: Do we have a plan to provide multiple servers(http and ftp ..) for call-home functionality?
Just curious.  :)

smehringer: As far as I know we do not plan on an ftp server. We set up an http server and the ftp call (you commented on) works just fine on Mac :) 
try out: 
`ftp -Vo /tmp/seqan.version  http://www.seqan.de/version_check/SeqAn_Mac_64_seqan_0.0.0`
(I just noticed that curl still has some standard output to be redirected. I need to fix that.)
But I do not know a lot about web servers so I'm not sure.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -45,131 +45,112 @@
...
 // Tags, Classes, Enums
 // ==========================================================================
 
+struct VersionControlTags
+{
+    constexpr static const char * SEQAN_NAME = "seqan";
+    constexpr static const char * UNREGISTERED_APP = "UNREGISTERED APP";
+};
+
+enum VersionControlOptions
```
rrahn: please use typed enums:
```cpp
enum class VersionControlOptions : uint8_t
{
    ...
};
```

include/seqan/arg_parse/arg_parse_version_check.h
xenigmax: Oh.. I didn't know that's possible.

h-2:
build failing travis, moving to 2.3.0

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -64,20 +64,26 @@ inline std::string _getBitSys();
...
+template <typename TVoidSpec = void>
+struct VersionControlTags_
 {
-    static const char * SEQAN_NAME;
-    static const char * UNREGISTERED_APP;
-    static const char * OPTION_OFF;
-    static const char * OPTION_DEV;
-    static const char * OPTION_APP_ONLY;
+    static constexpr char const * const SEQAN_NAME = "seqan";
+    static constexpr char const * const UNREGISTERED_APP= "UNREGISTERED APP";
```
h-2: should this maybe have an underscore, too? might be better for parsing...

include/seqan/arg_parse/argument_parser.h
```diff
@@ -237,7 +237,7 @@ class ArgumentParser
                                         ArgParseArgument::STRING,
                                         "OPTION"));
         setValidValues(*this, "version-check", "DEV OFF APP_ONLY");
-        setDefaultValue(*this, "version-check", VersionControlTags::OPTION_DEV);
+        setDefaultValue(*this, "version-check", VersionControlTags_<>::OPTION_DEV);
```
h-2: Maybe we should deactivate it here, instead of through the macro? Would make it easier to experiment with.

include/seqan/arg_parse/argument_parser.h
```diff
@@ -221,6 +227,18 @@ class ArgumentParser
...
+        addOption(*this, ArgParseOption("",
+                                        "version-check",
+                                        "Choose OFF if you don't want any notifications. "
+                                        "Choose APP_ONLY if you don't want any notification about a new SeqAn version. "
+                                        "DEV is the default value and will check for both versions.",
+                                        ArgParseArgument::STRING,
+                                        "OPTION"));
+        setValidValues(*this, "version-check", "DEV OFF APP_ONLY");
+        setDefaultValue(*this, "version-check", VersionControlTags_<>::OPTION_DEV);
+#endif
```
h-2:
```c++
#ifdef SEQAN_DISABLE_VERSION_CHECK
    hideOption(*this, "version-check", true);
    setDefaultValue(*this, "version-check", "OFF");
#else
    setDefaultValue(*this, "version-check", "APP_ONLY");
#endif
```

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,453 @@
...
+        _url = "http://www.seqan.de/version_check/SeqAn_" + _getOS() + _getBitSys() + _name + "_" + _version;
+        _getProgram();
+        _updateCommand();
+    }
+
+    // ----------------------------------------------------------------------------
+    // Member Functions
+    // ----------------------------------------------------------------------------
+
+#if defined(PLATFORM_WINDOWS)
```
h-2: this macro is deprecated, use STDLIB_VS instead.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,453 @@
...
+    // ----------------------------------------------------------------------------
+
+    VersionCheck(std::string const & name,
+                 std::string const & version,
+                 std::string const & website)
+    {
+        _name = name;
+        if (!version.empty() &&
+            std::regex_match(version, std::regex("^[[:digit:]]+\\.[[:digit:]]+\\.[[:digit:]]+.*")))
+            _version = version.substr(0,5); // in case the git revision number is given take only version number
```
h-2: what if the version is `10.23.12` ? then `substr()` will not get it all... Rather extract it with the regex module directly...

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,453 @@
...
+
+    // ----------------------------------------------------------------------------
+    // Constructors
+    // ----------------------------------------------------------------------------
+
+    VersionCheck(std::string const & name,
+                 std::string const & version,
+                 std::string const & website)
+    {
+        _name = name;
```
h-2: :nail_care: to permit move-construction, rather do this:
```c++
VersionCheck(std::string name,
                       ...)
{
    std::swap(_name, name);
    ...
```

rrahn:
Ok, I updated the test. This works now.
@h-2 any further comments?

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,459 @@
...
+#include <fstream>
+#include <string>
+#include <sys/stat.h>
+#include <time.h>
+#include <stdio.h>
+#include <regex>
+#include <future>
+#include <sys/types.h> 
+#include <chrono>
+
```
h-2: :nail_care: seperate system headers and stdlib; don't use c-style headers, but the c++ equivalents, i.e 
`#include <cstdio>` not `#include <stdio.h>`

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,459 @@
...
+{
+    std::string bitSys;
+
+#if SEQAN_IS_32_BIT
+    bitSys = "_32_";
+#else
+    bitSys = "_64_";
+#endif
+    return bitSys;
+}
```
h-2: instead:
```c++
constexpr const char * const _getBitSys()
{
#if SEQAN_IS_32_BIT
￼     return "_32_";
#else
￼     return "_64_";
￼#endif
}
```

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,459 @@
...
+        {
+            me._path = tmp_path;
+        }
+        else
+        { //GetTempPath() returns 0 on failure
+            prom.set_value(false);
+            return;
+        }
+# else // unix
+        me._path = "/tmp";
```
h-2: this should not be hard-coded, instead try the TMPDIR or TMP environment variables first. Also we have a SEQAN_DEFAULT_TMPDIR macro. Please check this out.

If we don't have it yet, add a documented function somewhere in `seqan/include/system/` that abstracts from the platform and returns the tmpdir.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,459 @@
...
+
+// ----------------------------------------------------------------------------
+// Function _callServer()
+// ----------------------------------------------------------------------------
+
+inline void _callServer(VersionCheck const me, std::promise<bool> prom)
+{
+    // update timestamp
+    std::string timestamp_filename = me._path + "/" + me._name + ".timestamp";
+    std::ofstream timestamp_file(timestamp_filename.c_str());
```
h-2: Unless there is a very good reason, try to always use brace-initialization (it is possible for both line preceeding this comment!).

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,459 @@
...
+            return;
+        }
+# else // unix
+        me._path = "/tmp";
+#endif
+        me._updateCommand();
+    }
+
+    std::string version_filename   = me._path + "/" + me._name + ".version";
+    std::string timestamp_filename = me._path + "/" + me._name + ".timestamp";
```
h-2: brace initialization ;-)

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -216,6 +216,13 @@ macro (seqan_build_system_init)
     set (SEQAN_NIGHTLY_RELEASE FALSE CACHE BOOL "Set to TRUE to enable nightly app releases.")
 
     ## options
+
+    # SeqAn Version Check
+    if (NOT SEQAN_VERSION_CHECK)
+        set (SEQAN_DEFINITIONS "${SEQAN_DEFINITIONS};-DSEQAN_DISABLE_VERSION_CHECK")
+    endif ()
```
h-2: this should be removed now

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,459 @@
...
+    os = "Windows";
+#elif __FreeBSD__
+    os = "FreeBSD";
+#elif __OpenBSD__
+    os = "OpenBSD";
+#else
+    os = "unknown";
+#endif
+    return os;
+}
```
h-2: also use conxtexpr cstring literal function for this.

h-2:
concerning the initialization you can uses braces everywhere for everything. This is a bit new, but it makes the code much more readable, because you immediately see whether a function is called or an object initialized. It is also much easier to differentiate between an assignment which takes place later and the initialization.

include/seqan/arg_parse/arg_parse_version_check.h
smehringer: I can only find the SEQAN_DEFAULT_TMPDIR defined inside some apps 
(e.g. ```./apps/micro_razers/micro_razers.cpp:32:    #define SEQAN_DEFAULT_TMPDIR "C:\\TEMP\\"```)
but nowhere globally. I then looked into the seqan macro SEQAN_TEMP_FILENAME, since it needs the temp folder of course and, as I see it, it does nothing differently from what I did:
https://github.com/seqan/seqan/blob/master/include/seqan/basic/debug_test_system.h#L681

include/seqan/arg_parse/arg_parse_doc.h
rrahn: make the return type auto in this case.

tests/arg_parse/test_arg_parse_version_check.cpp
```diff
@@ -0,0 +1,75 @@
+// ==========================================================================
+//                                 arg_parse
+// ==========================================================================
```
rrahn: Please use the following header:
//                 SeqAn - The Library for Sequence Analysis

tests/arg_parse/test_arg_parse_version_check.h
```diff
@@ -0,0 +1,332 @@
+// ==========================================================================
+//                                 arg_parse
+// ==========================================================================
```
rrahn: // ==========================================================================
//                 SeqAn - The Library for Sequence Analysis

tests/arg_parse/test_arg_parse_version_check.h
```diff
@@ -0,0 +1,332 @@
...
+const std::string SEQAN_VERSION_FILENAME(PATH + "/seqan.version");
+const std::string SEQAN_TIMESTAMP_FILENAME(PATH + "/seqan.timestamp");
+const std::string APP_VERSION_FILENAME(PATH + "/" + PARSER_NAME + ".version");
+const std::string APP_TIMESTAMP_FILENAME(PATH + "/" + PARSER_NAME + ".timestamp");
+
+constexpr const char * const APP_NAME("test");
+constexpr const char * const OPTION_VERSION_CHECK("--version-check");
+constexpr const char * const OPTION_OFF("OFF");
+constexpr const char * const OPTION_APP_ONLY("APP_ONLY");
+constexpr const char * const OPTION_DEV("DEV");
```
rrahn: Please wrap all these global variables in a struct

tests/arg_parse/test_arg_parse_version_check.cpp
```diff
@@ -0,0 +1,75 @@
...
+// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+// DAMAGE.
+//
+// ==========================================================================
+// Author: Svenja Mehringer <svenja.mehringer@fu-berlin.de>
+// ==========================================================================
+
+#define SEQAN_DEBUG
+
+
```
rrahn: double new-line

include/seqan/arg_parse/arg_parse_doc.h
h-2: why? :)

rrahn: Please ignore, I thought I deleted the comment.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,465 @@
...
+
+    HANDLE dummyFile; // check writablity by trying to create a file in GENERIC_WRITE mode
+    std::string fileName(path + "/dummy.txt");
+    dummyFile = CreateFile(fileName.c_str(), GENERIC_WRITE, 0, NULL, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
+
+    if (dummyFile == INVALID_HANDLE_VALUE)
+        return false;
+
+    CloseHandle(dummyFile);
+    DeleteFile(fileName.c_str());
```
rrahn: Please add an assertion that the return value must be non-zero. And return the with false if ```DeleteFile``` returns ```0```

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,465 @@
...
+// Function _getNumbersFromString()
+// ----------------------------------------------------------------------------
+
+inline String<int> _getNumbersFromString(std::string const & str)
+{
+    String<int> numbers;
+    StringSet<std::string> set;
+    strSplit(set, str, EqualsChar<'.'>(), false);
+    for (auto & num : set)
+        appendValue(numbers, lexicalCast<int>(num));
```
rrahn: Please add brackets around the for to make the scope clear.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,465 @@
...
+    if (timestamp_file.is_open())
+    {
+        std::string str_time;
+        std::getline(timestamp_file, str_time);
+        timestamp_file.close();
+        double d_time;
+        lexicalCast(d_time, str_time);
+        return curr - d_time;
+    }
+
```
rrahn: Maybe remove this empty line.

include/seqan/arg_parse/arg_parse_version_check.h
```diff
@@ -0,0 +1,465 @@
...
+            line.clear();
+        }
+        if (line == VersionControlTags_<>::UNREGISTERED_APP)
+        {
+            me.errorStream << VersionControlTags_<>::MESSAGE_UNREGISTERED_APP;
+            line.clear();
+        }
+        myfile.close();
+    }
+
```
rrahn: Maybe remove empty line here.

tests/arg_parse/CMakeLists.txt
```diff
@@ -37,8 +37,15 @@ add_executable (test_arg_parse
                test_argument_parser.h
                test_arg_parse_ctd_support.h)
 
+set (SEQAN_VERSION_CHECK ON FORCE)
+add_executable (test_arg_parse_version_check
+               test_arg_parse_version_check.cpp
```
rrahn: indent by one more space

tests/arg_parse/CMakeLists.txt
```diff
@@ -37,8 +37,15 @@ add_executable (test_arg_parse
...
+               test_arg_parse_version_check.h)
```
rrahn: same here

rrahn:
we are almost there 😆

h-2:
This PR has a huge number of commits with lots of small back-and-forth changes, please squash when merging or do some local cleanup of the history...

mr-c:
Will this be opt-in or opt-out?

Debian doesn't distribute opt-out "phone home"s like this, so if it will be opt-out please make it easy to switch to opt-in via configuration flag.

h-2:
@mr-c currently it is opt-in (for the upcoming  2.3.0 release), but it is planned to be opt-out in the 2.4.0 release if it has undergone more testing. Both run-time opt-out and compile-time opt-out are possible, however the compile-time-opt-out hides the option entirely so it's not a switch to opt-in.

For the run-time-opt-out we inform the user at first start about the update check and how to opt-out at run-time. We are aware that it is a sensitive topic and we had long discussions about this in the team, but we don't gather any information on the actual usage or anything related to the user's data. It's less data than any browser leaves when accessing a web-page…

We are ware that the feature is of less value to Debian users so long as the packages are regularly updated, but outside of official distribution packages many of our users use outdated software so we think it is quite useful to nag them...
Which of Debian's guidelines would this feature violate?

mr-c:
Debian is very privacy sensitive, so hence the general ban on opt-out "phone homes".

We can use the [debconf](https://wiki.debian.org/debconf) system to ask the administrator (typically the main user) at install time what their preference is and set the appropriate system wide switch to reflect that.

h-2:
@mr-c Ok, sounds good. Than lets re-discuss this again when we are going to default-activate it, ok? [2.4.0 is not planned before 2017-03]

smehringer:
@rrahn I gave the manual pages a first try. I'm not sure if this is the way you wanted this feature to be documented so please take a look.

include/seqan/arg_parse/argument_parser.h
```diff
@@ -221,6 +227,21 @@ class ArgumentParser
                                         "FORMAT"));
         hideOption(*this, "export-help", true);
         setValidValues(*this, "export-help", "html man txt");
+
+        addOption(*this, ArgParseOption("",
+                                        "version-check",
+                                        "Choose OFF if you don't want any notifications. "
+                                        "Choose APP_ONLY if you don't want any notification about a new SeqAn version. "
+                                        "DEV is the default value and will check for both versions.",
```
rrahn: Maybe we rewrite this. I think that people might not understand the difference between app_only and dev.
Maybe something like:
```
Choose OFF to disable any update notifications. 
With  APP_ONLY you'll receive update notifications about new versions of your app. 
In DEV mode you'll receive update notifications regarding new versions of your app 
and new versions of the SeqAn library.
```

rrahn:
Can you also please resolve the merge conflict in the manual?
Thanks.

CMakeLists.txt
```diff
@@ -79,6 +79,12 @@ option (SEQAN_ARCH_AVX2 "Build AVX optimized binaries." OFF)
 option (SEQAN_ARCH_NATIVE "Build natively optimized binaries." OFF)
 
 # ===========================================================================
+# Version Check Variable.
+# ===========================================================================
+
+option (SEQAN_VERSION_CHECK "SeqAn version check." ON)
```
smehringer: Is this the default value for the SEQAN_VERSION_CHECK variable? If so I should change it to OFF since in 2.3 we do now want it to be enabled by default yet, right?

rrahn: Nop, this is correct. We keep it in as compile time option, but the default of the version-check option is set to OFF. This means the people get informed, that there is such an option, when they run the help, but they have to opt-in explicitly and won't be affected if they don't.

PR #1224, opened 2015-10-08
Fixes several demos

rrahn:
Fixes build errors and warnings for some demos.

PR #1764, opened 2016-06-27
Fix intel warnings

marehr:
Fixes:
- test_basic_alphabet_adapt_builtins.h(221): warning #61: integer operation result is out of range
- platform.h(117): warning #1224: #warning directive: ICC versions older than 16 are not supported.

### DIFF
--- a/include/seqan/platform.h
+++ b/include/seqan/platform.h
@@ -113,7 +113,7 @@
  */
 #if defined(__ICC)
 #define COMPILER_INTEL
-#if _ICC < 1600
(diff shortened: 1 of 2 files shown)


h-2:
what about the other warnings like the acces behind the last element a.s.o.? Please try to bundle them as this makes review easier and we don't exhaust our slots in travis (it takes really long now for verification).

marehr:
For me other stuff is windows, this is clearly to fix the unix warnings. Two different areas. I don't mind if we merge that in 3 days.

Issue #1730, opened 2016-06-08
cmake macros and c++ macros for platforms


h-2:
bring them in sync so that e.g. COMPILER_INTEL always means the same thing.

Issue #1385, opened 2016-01-28
add support for Clang on Windows


h-2:
- [ ] get clang/C2 working on our build machines
  - [ ] adapt the platform detection to accomodate for MSCV and CLANG being defined
  - [ ] test and fix broken stuff
  - [ ] update README.RST and homepage

h-2:
@cpockrandt can you take this for 2.1.1 ? Shouldn't be a lot of work and is potentially really cool.

cpockrandt:
Couldn't get cmake working with clang.exe on our build machines.
Therefore postponed until further notice.

h-2:
Cmake has support now:
https://cmake.org/pipermail/cmake-developers/2016-March/027967.html
It is in the nightlies and will be in the next release. @cpockrandt Can you check with the nightlies, so that we have support, once the next cmake version is out?

> ￼ adapt the platform detection to accommodate for MSCV and CLANG being defined

I would propose that we remove PLATFORM_WINDOWS and that we define `PLATFORM_MSVC` only on `MSVC && != c2`. If there are sufficiently many cases where `PLATFORM_CLANG` doesn't for c2, we could add an extra `PLATFORM_C2`

h-2:
There are also fixes for clang/c2 in the new vs2015update:
https://www.visualstudio.com/en-us/news/vs2015-update2-vs.aspx
So please ask the staff to update vs2015 on our virtual machines. 

And maybe ask them to install the vs2016 preview, as well:
http://www.heise.de/newsticker/meldung/Microsoft-Build-2016-Vorschau-von-Visual-Studio-2016-veroeffentlicht-3158425.html

h-2:
Can you move the c2 build to margerite so that it is clear that it's a 32bit build? Do you have time tomorrow that we can look at the razers failure?

marehr:
I already added the c2 builds to margerite, but the intel compiler needed 8 hours to build the develop branch. Thus, c2 is still building (already 4 hours).

marehr:
@h-2  Just to let you know, I started an update on margerite to VS Update 3.

h-2:
> @h-2 Just to let you know, I started an update on margerite to VS Update 3.

:open_mouth: lets :pray: that it works

marehr:
@h-2 I rechecked the clang binaries, they apparently are 64bit. Only the compiler itself is 32bit.

```console
KLETTE+seqan_dev@KLETTE MINGW64 /c/Users/marehr
[...]

PR #1767, opened 2016-06-28
Support intel and clang on windows

marehr:
Fixes
* #1730,
* #1274 
  - add support for windows/icc and 
* #1385 
  - adapt the platform detection to accommodate for MSCV and CLANG being defined
  - test and fix broken stuff


include/seqan/misc/bit_twiddling.h
```diff
@@ -272,7 +272,7 @@ popCount(TWord word)
 // intrinsics __popcnt16, __popcnt, and __popcnt64 for 16, 32, and 64 bit words.
 
 // MSVC >= 2008, has intrinsic
-#if defined(_MSC_VER)
+#if defined(STDLIB_VS)
```
h-2: since these are compiler builtins, maybe use `defined(COMPILER_MSVC) || defined(COMPILER_WINTEL)` instead? I guess clang/c2 can use the regular builtins...

include/seqan/misc/bit_twiddling.h
```diff
@@ -328,7 +328,7 @@ _popCountImpl(TWord word, WordSize_<8> const & /*tag*/)
 }
 
 // GCC or CLANG
-#elif !defined(_MSC_VER)
+#elif !defined(STDLIB_VS)
```
h-2: see above

include/seqan/parallel/parallel_atomic_primitives.h
```diff
@@ -231,9 +231,13 @@ struct Atomic
 // Implementation in MSVC
 // ----------------------------------------------------------------------------
 
+// NOTE(marehr): clang/c2 v3.7 doesn't know #pragma intrinsic.
+#if !defined(COMPILER_CLANG)
```
h-2: if this only affects c2 and not the unix clang, than please use `!defined(__c2__)`

marehr: See further above into the file, I'm in a section of `STDLIB_VS`.

include/seqan/platform.h
```diff
@@ -365,14 +381,14 @@ typedef int8_t __int8;     // nolint
 // ==========================================================================
 // C++ branch hints
 // ==========================================================================
-#if defined(COMPILER_GCC) || defined(COMPILER_CLANG) || defined(COMPILER_INTEL)
+#if !defined(STDLIB_VS)
```
h-2: I would rather keep this as it was before, because it is a non-standard feature of these compilers. I know the effect is the same, but it is clearer if defined by compiler VS standard library (it has nothing to do with the standard library).

include/seqan/platform.h
```diff
@@ -365,14 +381,14 @@ typedef int8_t __int8;     // nolint
...
 #define SEQAN_LIKELY(expr) __builtin_expect(!!(expr), 1)
 #else
 #define SEQAN_LIKELY(x)    (x)
 #endif
 
-#if defined(COMPILER_GCC) || defined(COMPILER_CLANG) || defined(COMPILER_INTEL)
-#define SEQAN_UNLIKELY(expr) __builtin_expect(!!(expr), 1)
```
h-2: here as well [but please don't revert the chance on the expected value.

include/seqan/platform.h
```diff
@@ -393,7 +409,11 @@ typedef int8_t __int8;     // nolint
     #ifdef STDLIB_VS
     #define SEQAN_HAS_EXECINFO 0
     #elif defined(__has_include)
-    #define SEQAN_HAS_EXECINFO !!__has_include(<execinfo.h>)
+        #if __has_include(<execinfo.h>)
+        #define SEQAN_HAS_EXECINFO 1
+        #else
+        #define SEQAN_HAS_EXECINFO 0
+        #endif
     #else // assume that it is there
```
h-2: why was this revertet?

tests/pipe/test_pipe.cpp
```diff
@@ -299,22 +299,18 @@ SEQAN_DEFINE_TEST(test_pipe_test_sorter) {
...
+template <typename TStringSet, typename T>
+inline void appendValues(TStringSet &stringSet, T t)
 {
-    va_list ap;
-    va_start(ap, numArgs);
-    for(int j = 0; j < numArgs; ++j)
-        appendValue(stringSet, va_arg(ap, char const *));
-    va_end(ap);
+    appendValue(stringSet, t);
 }
```
h-2: what is the point of this function now, why can it not be removed?

tests/pipe/test_pipe.cpp
```diff
@@ -299,22 +299,18 @@ SEQAN_DEFINE_TEST(test_pipe_test_sorter) {
...
+template <typename TStringSet, typename T, typename... Args>
+inline void appendValues(TStringSet &stringSet, T t, Args... args)
 {
-    typename Size<TStrings>::Type numStrings = length(strings);
-    for(int j = 0; j < numStrings; ++j)
-        appendValue(stringSet, strings[j]);
+    // use recursive variadic function, instead of va_list
+    appendValue(stringSet, t);
+    appendValues(stringSet, args...) ;
 }
```
h-2: why where these functions changed? IIANM iterating is faster than recursion because recursive functions cannot be inlined...

include/seqan/misc/bit_twiddling.h
marehr: okay just tested, clang supports builtins

include/seqan/platform.h
marehr: We also reverted the expression down below, because it caused problems. I can't remember the expression, but something like `!SEQAN_ASYNC_IO && OTHER_CONSTANT` resulted in a compiler error.

```c++
     #elif defined(__has_include) && defined(__unix__)
         #if __has_include(<aio.h>)
         #define SEQAN_ASYNC_IO 1
         #else
         #define SEQAN_ASYNC_IO 0
         #endif
#endif
```

util/cmake/FindSeqAn.cmake
```diff
@@ -174,6 +171,8 @@ endif ()
 # Require C++14
 # ----------------------------------------------------------------------------
 
+# The standard library of visual studio defines __cplusplus still as
```
h-2: the macro is not defined by the standard library, but by the compiler. Clang/c2 correctly sets the macro!

util/cmake/FindSeqAn.cmake
```diff
@@ -354,9 +355,14 @@ if (NOT _SEQAN_FIND_OPENMP EQUAL -1)
 endif ()
 
 if (OPENMP_FOUND)
-    if (COMPILER_IS_CLANG AND (_GCC_VERSION MATCHES "^37[0-9]$"))
+    if (COMPILER_CLANG AND (_GCC_VERSION MATCHES "^37[0-9]$"))
```
h-2: Maybe we should change _GCC_VERSION to _COMPILER_VERSION or use "CMAKE_CXX_COMPILER_VERSION" directly?

util/cmake/FindSeqAn.cmake
```diff
@@ -366,6 +372,16 @@ if (OPENMP_FOUND)
     endif ()
 endif ()
 
+if (Boost_FOUND)
+  # Example warning:
+  # C:\seqan-deps\boost_1_53_0\boost/mpl/if.hpp(131,1): error : pasting formed
+  # 'BOOST_PP_TUPLE_ELEM_E_2(', an invalid preprocessing token
+  if (COMPILER_CLANG AND STDLIB_VS)
+    message (STATUS "The boost library (at least until 1.53) doesn't support the clang/c2 compiler on windows (version 3.7 and 3.8), yet!")
+    set (Boost_FOUND FALSE)
```
h-2: also unset the following, just in case someone includes boost as optional:
Boost_INCLUDE_DIRS
Boost_LIBRARY_DIRS
Boost_LIBRARIES

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -47,30 +47,40 @@
...
-# ---------------------------------------------------------------------------
-# Set architecture for MinGW.
-#
-# If we do not set i586 as the architecture for MinGW then generating atomic
-# expressions will fail.
-# ---------------------------------------------------------------------------
+if (COMPILER_MSVC OR COMPILER_WINTEL)
+    # Set /bigobj for COMPILER_MSVC and COMPILER_WINTEL, but COMPILER_CLANG on
+    # windows (clang/c2 3.7) can not handle it.
+    add_definitions (/bigobj)
```
h-2: It does work now on clang/c2 and we should set it:
https://blogs.msdn.microsoft.com/vcblog/2016/03/31/clang-with-microsoft-codegen-march-2016-released/

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -47,30 +47,40 @@
...
+endif()
 
-if (MINGW)
-	if ("${CMAKE_SIZEOF_VOID_P}" EQUAL "4")
-	    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=i586")
-	endif ("${CMAKE_SIZEOF_VOID_P}" EQUAL "4")
-endif (MINGW)
+if (STDLIB_VS)
+    add_definitions (-D_WIN32_WINNT=0x0600 -DWINVER=0x0600)
```
h-2: Whats this stuff? [I know it was there before, but why?]

tests/pipe/test_pipe.cpp
marehr:
```c++
va_list ap;
va_start(ap, numArgs);
for(int j = 0; j < numArgs; ++j)
    appendValue(stringSet, va_arg(ap, char const *));
va_end(ap);
```

va_list gave warnings (extension used or so) for clang.

Since it is only a convenient function in this test case, which adds a constant number of strings to a string set, I changed it to a variadic function.  For example:

```c++
appendValues(expectedOutput,
     "< 2 , [T A A] >",
     "< 2 , [T A A] >",
     "< 1 , [A A A] >");
     "< 1 , [A A A] >");
```

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -687,4 +716,3 @@ macro (seqan_register_tests)
         endif (IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${ENTRY})
     endforeach (ENTRY ${ENTRIES})
 endmacro (seqan_register_tests)
-
```
h-2: don't remove empty line at end!

tests/pipe/test_pipe.cpp
marehr: This are a couple of string known at compile time, this should be properly inlined. And even if not, it is only a convenient function in this test case.

util/cmake/FindSeqAn.cmake
```diff
@@ -112,54 +112,51 @@ endif ()
 
 # Recognize Clang compiler.
 
-set (COMPILER_IS_CLANG FALSE)
+set (COMPILER_CLANG FALSE)
+set (COMPILER_GNU FALSE)
```
h-2: In the source code we call it COMPILER_GCC so maybe we should do that here, as well!

include/seqan/parallel/parallel_atomic_primitives.h
h-2: ok!

util/cmake/FindSeqAn.cmake
marehr: How would you change the comment? I find it without a comment not clear, why we exclude MSVC.

include/seqan/platform.h
h-2: ok, whatever, not so important.

tests/pipe/test_pipe.cpp
h-2: > va_list gave warnings 

Yes, don't use it.

But, use initializer_list instead, it is very simple.

h-2:
This PR is a leap forward very good! The only major question I have is: If `COMPILER_WINTEL` is intel-on-windows, should `COMPILER_INTEL` really be unix-only?
Maybe we could do:
* COMPILER_WINTEL == __ICL
* COMPILER_LINTEL == __ICC
* COMPILER_INTEL == (COMPILER_WINTEL || COMPILER_LINTEL)

What do you think? Do we have any use-cases for the last one?

util/cmake/SeqAnBuildSystem.cmake
marehr: https://msdn.microsoft.com/en-us/library/6sehtctf.aspx

Seems to be there to support >= windows vista.

tests/pipe/test_pipe.cpp
marehr: This works, if you want to change it, please provide the source code. I see no benefit for a initializer_list. The change was based on http://stackoverflow.com/questions/16987493/getting-the-type-of-parameters-passed-in-a-variable-arguments-list

util/cmake/FindSeqAn.cmake
h-2: It is unrelated to thr STDLIB, check for COMPILER_MSVC and COMPILER_WINTEL (?) instead and make the comment explain that its about compiler, not stdlib ;)

marehr:
Nop, we don't have a use case for `COMPILER_INTEL == (COMPILER_WINTEL || COMPILER_LINTEL)`. As I said before, both compilers are distinct to each other.

include/seqan/misc/bit_twiddling.h
marehr: done

include/seqan/misc/bit_twiddling.h
marehr: done

include/seqan/platform.h
marehr: done

include/seqan/platform.h
marehr: done

util/cmake/FindSeqAn.cmake
marehr: nice catch, was an oversight of me. 

----
done

util/cmake/FindSeqAn.cmake
marehr: done

util/cmake/FindSeqAn.cmake
marehr: another issue/pull request, this was used before

marehr: https://github.com/seqan/seqan/issues/1770

util/cmake/FindSeqAn.cmake
marehr: We didn't do that for OpenMP

marehr: done

util/cmake/SeqAnBuildSystem.cmake
marehr: Ah! I didn't try that with clang 3.8, I will check that.

util/cmake/SeqAnBuildSystem.cmake
marehr: Here were two empty lines, thus there is still an empty line. Github would complain if it doesn't have an empty line at the end of a file.

util/cmake/FindSeqAn.cmake
h-2: ok. We don't need it for OpenMP, because we don't include the Flags anyway, but we would for Boost, because they are not part of the SeqAn-flags.

util/cmake/SeqAnBuildSystem.cmake
h-2: ok

util/cmake/SeqAnBuildSystem.cmake
h-2: hm, looks strange, but lets not break it before the release ;)

h-2:
> Nop, we don't have a use case for COMPILER_INTEL == (COMPILER_WINTEL || COMPILER_LINTEL). As I said before, both compilers are distinct to each other.

Should we still rename COMPILER_INTEL to COMPILER_LINTEL?

marehr:
:racehorse:  and now merge

util/cmake/SeqAnBuildSystem.cmake
marehr: Still failing:

```
  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\Clang 3.7\bin\x86\clang.exe -c -fdiagnostics-format=msvc -target "amd64-pc-windows-msvc" -I "C:\nightly\co-git-develop-wip\seqan-Nightly-64\include" -I "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\Clang 3.7\include" -I "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include" -I "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\atlmfc\include" -I "C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt" -I "C:\Program Files (x86)\Windows Kits\8.1\Include\um" -I "C:\Program Files (x86)\Windows Kits\8.1\Include\shared" -I "C:\Program Files (x86)\Windows Kits\8.1\Include\winrt" -o "test_misc.dir\Release\test_misc.obj" -Wall -O2 -fno-strict-aliasing -fomit-frame-pointer -ffunction-sections -fdata-sections -fstack-protector -fpic -fno-short-enums -fno-rtti -D WIN32 -D _WINDOWS -D SEQAN_GLOBAL_EXCEPTION_HANDLER=1 -D _WIN32_WINNT=0x0600 -D WINVER=0x0600 -D _SCL_SECURE_NO_WARNINGS -D _CRT_SECURE_NO_WARNINGS -D _LARGEFILE_SOURCE -D _FILE_OFFSET_BITS=64 -D SEQAN_ENABLE_TESTING=1 -D NOMINMAX -D SEQAN_HAS_EXECINFO=0 -D "CMAKE_INTDIR=\"Release\"" -D _MBCS -x c++ -fms-extensions -fno-ms-compatibility  -fms-extensions -fms-compatibility -frtti -fexceptions -W -pedantic -fstrict-aliasing -Wstrict-aliasing /bigobj  -D_MT -D_DLL -Xclang --dependent-lib=msvcrt -Xclang --dependent-lib=oldnames "C:\nightly\co-git-develop-wip\seqan-Nightly-64\tests\misc\test_misc.cpp"
  test_misc.cpp
clang.exe : error : no such file or directory: '/bigobj'
```

h-2:
Ok, codewise this looks good, but on cdash there are still
* two warnings for icc
* one error for clang
*16 warnings for clang

The last point might be  @rrahn 's responsibility, but the other ones need fixing!

marehr:
* two warnings for icc (actually only one)
  * #warning directive: "The Intel Compiler has a bug for this type of range-based for loop (at least until v16.0.2). Please reevaluate this issue for newer versions."
  * We should wait until the 16.0.3 update is installed on unix, to re-evaluate the warning.
* one error for clang
  * app_test_razers3: fails
  * Actually, I don't know why it fails and what is expected, there should someone else take a look. My only guess would be, that razers uses a seed-based random function, that behaves differently from msvc and wintel.
* 16 warnings for clang  (actually only 8)
  * warning : SSE4.1 instruction set not enabled [-W#pragma-messages]
  * is not critical
  * file_stream.h(867,28): warning : comparison of integers of different signs: 'TSize' (aka 'unsigned long long') and 'streamoff' (aka 'long long') 
  * That's the only real open issue.

marehr:
Just to mention this. The nightlies of icc and clang/c2 are based on 10aba81cfa6b9fc3e462aab0304db5aa51ea8d8d, which was committed on 09. June 2016. So after merging this there might be some new errors/warnings due to changes made since 10aba81cfa6b9fc3e462aab0304db5aa51ea8d8d.

h-2:
> We should wait until the 16.0.3 update is installed on unix, to re-evaluate the warning.

As there has been no update on the issue in our intel premier support, please just bump this one patch-level up.

> one error for clang

@rrahn can you have a look?

> warning : SSE4.1 instruction set not enabled

@rrahn can you have a look?

> That's the only real open issue.

No, everything really should be green. We take build failures too lightly which always causes more work down the road. We need to get this is into a 100% green situation and we should not treat breakage as a "TODO for the next release", but a "YOU HAVE TO FIX THIS BEFORE ANYTHING ELSE".
New platforms will not be accepted if they produce warnings or error. 
The only exceptions are bam_io where @rrahn is working on and the rare FreeBSD error which is actually unfixable by us.

marehr:
Hmm, so you suggest, that our test cases cover everything and that the clang compiler is marked as stable by Microsoft. Both is obviously not true. For me this is an experimental addition of two platforms. No one changed the docs, that says that both platforms are stable and should be used. This pr clearly aims to add an **initial** support for those platforms.

h-2:
> No one changed the docs, that says that both platforms are stable and should be used. 

http://www.seqan.de :trollface:

marehr:
Can we merge this? The new errors will be fixed by #1772 which is depending on this pr. The razers3 failure should be investigated separately.

h-2:
Ok, but I am not closing #1385 and #1274 since there are some remaining issues. Lets continue there.

h-2:
Please see also #1773 and #1774

PR #1768, opened 2016-06-28
[FIX] Deactivating PoolParameters for 64bit because too much memory is allocated

cpockrandt:


marehr:
Thank you for the fix. The benchmarks with indices works on windows again :+1:

PR #1771, opened 2016-06-29
[INTERNAL] Fix warning C4146: unary minus operator applied to unsigne…

marehr:
cdash fails with
```
warning C4146: unary minus operator applied to unsigned type, result still unsigned
```
on msvc


rrahn:
lgtm!

PR #1772, opened 2016-06-29
[FIX] Compiler errors and warnings

rrahn:
* Adds AVX2 support to cmake
* Makes basic_simd_vector requiring at least SSE4.1 and SSE4.2.
* Fixes indentation issue detected on gcc6


util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -247,14 +254,25 @@ macro (seqan_build_system_init)
...
-        if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
-            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mpopcnt")
-        endif ()
-        if (CMAKE_CXX_COMPILER_ID MATCHES "Intel")
-            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -ipo -no-prec-div -fp-model fast=2")
+    else ()
+        if (SEQAN_ARCH_SSE4)
+            if (SEQAN_ARCH_AVX2)
+                message (STATUS "Building optimized binaries up to AVX2 and POPCNT.")
+                set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mavx -march=core-avx2")
```
h-2: don't use `-march=core-avx2` use `-mavx2` instead.

h-2:
Also, please wait for #1767 since it will change some cmake-macro names. So,e checks that are currently COMPILER_MSVC will ahve to be changed to `COMPILER_MSVC || COMPILER_WINTEL`

util/cmake/SeqAnBuildSystem.cmake
h-2: `-march=` are **architecture** flags, they imply a **set** of flags for a **specific CPU (generation)**
In this case `-march=core-avx2` is the equivalent of `-march=haswell` which specifically target intel haswell (or newer) CPUS, activating the following features:
> MOVBE, MMX, SSE, SSE2, SSE3, SSSE3, SSE4.1, SSE4.2, POPCNT, AVX, AVX2, AES, PCLMUL, FSGSBASE, RDRND, FMA, BMI, BMI2 and F16C instruction set support.

which could as well be activated manually.... but which we don't want to do, because it will break on newer AMDs (e.g. newest bulldozer which support AVX2, but don't support RDRND).

→ please always use explicit `-m*` flags and not `-march=*` because the latter bring a lot of other stuff which we don't know and which will probably not work on other platforms.

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -247,14 +254,25 @@ macro (seqan_build_system_init)
...
+            endif ()
+            set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mmmx -msse -msse2 -msse3 -mssse3 -msse4")
+            if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+                set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -mpopcnt")
+            endif ()
+            if (CMAKE_CXX_COMPILER_ID MATCHES "Intel")
+                set (SEQAN_CXX_FLAGS "${SEQAN_CXX_FLAGS} -ipo -no-prec-div -fp-model fast=2")
+            endif ()
+        else ()
+            if (NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")  # Explicitly disable extended vector instruction sets.
```
h-2: I don't think we need this anymore, now that the simd_module is fixed. I have read up on the issue, and `-msse3` is set on osx by clang and gcc, but none of the other flags. On all other platforms, none of the flags are set anyway.

h-2:
this needs to be braught in sync with current develop branch

marehr:
@rrahn Did you check why the intel compiler fails?

util/cmake/SeqAnBuildSystem.cmake
rrahn: Yes you are right.

rrahn:
@marehr I will have a look

h-2:
@rrahn Have you checked whether the intel errors are fixed?

rrahn:
@h-2 @marehr Seems like a bug in Intel compiler. I fixed it and will reproduce it to send a bug report to intel.

PR #1776, opened 2016-06-30
[DOC] Added missing documentation for constructors

smehringer:
Fixes  #1638

PR #1777, opened 2016-06-30
[DOC] Added missing documentation for constructors

### LINKED ISSUE #1638: [DOX] No documentation for the constructor of the LocalAlignmentEnumerator class
If you take a look, there is no documentation of the constructor:
http://docs.seqan.de/seqan/master/index.html?p=LocalAlignmentEnumerator

same here:
class Score http://docs.seqan.de/seqan/master/index.html?p=Score
class Index http://docs.seqan.de/seqan/master/index.html?p=Index

smehringer:
master PR #1776
Fixes #1638

### DIFF
--- a/include/seqan/align/local_alignment_enumeration.h
+++ b/include/seqan/align/local_alignment_enumeration.h
@@ -63,7 +63,7 @@ class LocalAlignmentEnumerator;
  * @tparam TScore The type of the @link Score @endlink to use for the local alignment.
  * @tparam TSpec  The tag to use for specializing the enumerator.
  *
- * This is an abstract class that will only be instantiated thorugh its subclasses.
+ * This is an abstract class that will only be instantiated through its subclasses.
  * See the documentation of the specializations for examples.
  *
  * @section References
--- a/include/seqan/align/local_alignment_enumeration.h
+++ b/include/seqan/align/local_alignment_enumeration.h
@@ -63,6 +63,7 @@ class LocalAlignmentEnumerator;
  * @tparam TScore The type of the @link Score @endlink to use for the local alignment.
  * @tparam TSpec  The tag to use for specializing the enumerator.
  *
+ * This is an abstract class that will only be instantiated thorugh its subclasses.
  * See the documentation of the specializations for examples.
  *
  * @section References
--- a/include/seqan/index/index_dfi.h
+++ b/include/seqan/index/index_dfi.h
@@ -215,6 +215,22 @@ namespace seqan
         TPredHull        predHull;
         TPred            pred;
 
+        /*!
+         * @fn IndexDfi::Index
+         * @brief Constructor
+         *
+         * @signature Index::Index();
+         * @signature Index::Index(index);
+         * @signature Index::Index(text);
+         * @signature Index::Index(text, predHull);
+         * @signature Index::Index(text, predHull, pred);
+         *
+         * @param[in] index    Other Index object to copy from.
+         * @param[in] text     The text to be indexed.
+         * @param[in] predHull (TPredHull) A monotonic hull of TPred.
+         * @param[in] pred     (TPred) An arbitrary frequeny predicate.
+         */
+
         Index() {}
 
         Index(Index &other):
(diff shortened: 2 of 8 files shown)


include/seqan/align/local_alignment_enumeration.h
```diff
@@ -63,6 +63,7 @@ class LocalAlignmentEnumerator;
  * @tparam TScore The type of the @link Score @endlink to use for the local alignment.
  * @tparam TSpec  The tag to use for specializing the enumerator.
  *
+ * This is an abstract class that will only be instantiated thorugh its subclasses.
```
rrahn: typo: through

include/seqan/index/index_dfi.h
```diff
@@ -215,6 +215,22 @@ namespace seqan
         TPredHull        predHull;
         TPred            pred;
 
+        /*!
+         * @fn IndexDfi::Index
+         * @brief Constructor
+         *
+         * @signature Index::Index();
+         * @signature Index::Index(index);
+         * @signature Index::Index(text);
```
rrahn: Make only one signature, with optional arguments like:
Index(text [, predHull] [,pred]) and write in the param section that this is an optional parameter.

include/seqan/index/index_fm.h
```diff
@@ -205,6 +205,16 @@ class Index<TText, FMIndex<TSpec, TConfig> >
...
 
+    /*!
+     * @fn FMIndex::Index
+     * @brief Constructor
+     *
+     * @signature Index::Index();
+     * @signature Index::Index(text);
+     *
+     * @param[in] text The text to be indexed.
+     */
```
rrahn: @h2 @cpockrandt  is it true, that the text must be in reversed for the FM-Index. If so we should drop a node in the description of the descriptor.

include/seqan/index/index_qgram.h
```diff
@@ -257,6 +257,20 @@ class Index<TText_, IndexQGram<TShapeSpec, TSpec> > {
...
     TSize            stepSize;    // store every <stepSize>'th q-gram in the index
 
+    /*!
+     * @fn IndexQGram::Index
+     * @brief Constructor
+     *
+     * @signature Index::Index();
+     * @signature Index::Index(index);
+     * @signature Index::Index(text);
+     * @signature Index::Index(text, shape);
```
rrahn: Make shape as optional parameter in one signature.

include/seqan/index/index_qgram_openaddressing.h
```diff
@@ -116,6 +116,21 @@ namespace seqan
...
 
+        /*!
+         * @fn OpenAddressingQGramIndex::Index
+         * @brief Constructor
+         *
+         * @signature Index::Index();
+         * @signature Index::Index(index);
+         * @signature Index::Index(text);
+         * @signature Index::Index(text, predHull);
+         * @signature Index::Index(text, predHull, pred);
```
rrahn: Copy'n'Paste error. shape instead predHull and pred

include/seqan/index/index_fm.h
cpockrandt: Yes, the text of the FM-index has to be reversed

smehringer: so shall I change the documentation?
`Index::Index(revText)`
`revText   the reversed text to be indexed`

maybe with a link to the `reverse` function for strings?

cpockrandt: Or we could refer to ModifiedStrings.

I think @h-2 uses the reverse function in Lambda. Maybe he can explain 
why he doesn't use a ModifiedString instead. Performance reasons?

h-2: > If so, we should drop a node in the description of the descriptor.

Yes, please document that the all the talk about SATree and FibreSA is misleading, because the FM-Index in the current implementation does not provide a suffx trie interface, but a prefix trie interface. 

> Maybe he can explain why he doesn't use a ModifiedString instead. Performance reasons?

Yes, I recommond using `reverse()` because it moves the transformation to time-of-indexing, instead of time-of-access.

smehringer: @cpockrandt will add a tag ("suffix") to the FMindex interface s.t. the user will be able to use the FMindex as a suffix-trie and the text does not need to be reversed beforehand. Since the documentation of this should be done in another pull request, the current documentation here is sufficient I think?

Issue #1797, opened 2016-07-11
Razers3 -- differences between serial and parallel mode


h-2:
Razers3 has two different codepaths, one serial, one parallel. Selected a thread count of 0 (`-tc 0`) chooses the serial one, and choosing a thread count of 1 (`-tc 1`) , chooses the parallel codepath, but with exactly one thread (also serial execution).

If run with `-rr 100` (full sensitivity) they produce the same results, but if the recognition rate is lowered to 99, than the parallel codepath misses results that the serial one finds. Unless the serial codepath actually ignores the rr setting, this hints at a bug in the parallel codepath.

h-2:
This example reproduces the problem:
`bin/razers3 adeno-genome.fa adeno-reads36_1.fa adeno-reads36_2.fa -i 98 -rr 99`
Depending on whether `-tc 0` or `-tc 1` is specified, the results will differ.

[adeno-genome.fa.txt](https://github.com/seqan/seqan/files/357365/adeno-genome.fa.txt)
[adeno-reads36_1.fa.txt](https://github.com/seqan/seqan/files/357367/adeno-reads36_1.fa.txt)
[adeno-reads36_2.fa.txt](https://github.com/seqan/seqan/files/357368/adeno-reads36_2.fa.txt)

holtgrewe:
I don't think that this is a bug.

The sequential and the parallel code path both only guarantee that at least one match per equivalence class is returned. As the parallel processing comes from window-wise processing, different matches are generated and thus different results are generated.

h-2:
This is not about equivalence classes, but about number of mapped reads which is approximately half in parallel mode VS sequential. 
But maybe I misunderstood something?

holtgrewe:
I see.

Maybe @weese can comment here?

h-2:
Closed because inactivity.

PR #1799, opened 2016-07-11
fix multiple app errors on windows

h-2:
contains multiple fixes, also a workaround for RazerS.


apps/razers3/paramChooser.h
```diff
@@ -102,7 +102,8 @@ struct ParamChooserOptions
         // global input parameters
         totalN = 32;                // sequence length
         totalK = 2;                 // errors
-        optionLossRate = (TFloat)0.01;      // in
+//         optionLossRate = (TFloat)0.01;      // in
```
rrahn: Remove comment

rrahn:
Ok, lgtm. Merge after addressing the minor issue.

apps/razers3/paramChooser.h
h-2: this comment is intentional, because the original is the actually intended and documented behaviour. Once #1797 is addressed, this can be reverted.

I will merge this now, as is. If you disagree, I can make a [NOP] PR tomorrow,

PR #1819, opened 2016-07-22
[COMPILER WARNING] Added deprecated macro to trigger compiler warnings.

### LINKED ISSUE #1794: Proper deprecation
Search the codebase for things that are deprecated and mark them with `[[deprecated]]` so that we actually get warnings when they are used.

smehringer:
Fixes #1794 

The PR will cause about 14035 warnings due to the usage of deprecated code. This will be fixed in later commits.


include/seqan/arg_parse/arg_parse_argument.h
```diff
@@ -136,7 +136,6 @@ class ArgParseArgument
         OUTPUT_FILE,  // .. an outputfile (implicitly also a string)
         INPUT_PREFIX, // .. an inputprefix (implicitly also a string)
         ///@deprecated use INPUT_PREFIX instead
-        [[deprecated("Use INPUT_PREFIX instead.")]]
```
smehringer: @h-2 I can't find any documentation on how to deprecate a single enum value.

h-2: Hm, it works by just placing the compiler hint after the value:
```c++
#include <iostream>

enum FOO
{
    BAR,
    BAX [[deprecated]],
    BAT
};

int main()
{
    FOO foo{BAX};
    return 0;
}
```
But of course this will probably not be detected if people are using static_casts or something like that.

h-2: For that case the syntax would be
```c++
enum FOO
{
    BAR,
    BAX [[deprecated("Use FOO::BAR instead.")]] = BAR,
    BAT
};
```

h-2:
@smehringer It seems like the last two commits introduced warnings again. Can you have a look?

PS: try to reduce the number of commits a little in the future please, e.g. subsume all the `[INTERNAL] Replaced getValue() by operator*().` into one. → a commit should be a logical unit, but if it is too fine-grained it is more confusing than helpful. Also use `git commit --amend` instead of revert when "undoing" previous changes in PRs. See, also https://nathanhoad.net/git-amend-your-last-commit

PR #1822, opened 2016-07-26
Add ARM compatibility

rrahn:
Disables SIMD code and fixes parallel module on ARM architecture.
Note, we still need to deactivate the tests for bs_tools for ARM, due to some rounding errors.

### DIFF
--- a/include/seqan/basic/basic_simd_vector.h
+++ b/include/seqan/basic/basic_simd_vector.h
@@ -43,15 +43,26 @@
 #include <tuple>
 //#include <seqan/basic/iterator_zip.h>
 
-#if defined(PLATFORM_WINDOWS_VS)
-  /* Microsoft C/C++-compatible compiler */
-  #include <intrin.h>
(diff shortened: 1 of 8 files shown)


include/seqan/basic/basic_simd_vector.h
```diff
@@ -43,15 +43,26 @@
...
-#if defined(PLATFORM_WINDOWS_VS)
-  /* Microsoft C/C++-compatible compiler */
-  #include <intrin.h>
-#elif defined(PLATFORM_GCC) && (defined(__x86_64__) || defined(__i386__))
-  /* GCC-compatible compiler, targeting x86/x86-64 */
-  #include <x86intrin.h>
-#else
- #warning "No supported platform for SIMD vectorization!"
-#endif
+#if !defined(__arm__) && !defined(_M_ARM)  // We do not support arm architecture.
```
h-2: instead check for __amd64__ || x86__64 || i386

because we don't want to fix this again, if someone targets MISC or powerpc or ...

PR #1829, opened 2016-08-11
[FIX] Fixing platform detection for basic simd vector.

rrahn:


include/seqan/basic/basic_simd_vector.h
```diff
@@ -42,26 +42,25 @@
...
+    #elif defined(__SSE4_1__) && defined(__SSE4_2__)
+        #define SEQAN_SSE4
+        #define SEQAN_SIZEOF_MAX_VECTOR 16
+    #else  // defined(__AVX2__)
+        #undef SEQAN_SIMD_ENABLED  // Disable simd if instruction set is not supported.
+    #endif  // defined(__AVX2__)
+#endif  // defined(COMPILER_GCC) || defined(COMPILER_CLANG) || defined(COMPILER_LINTEL)
+
+#ifdef SEQAN_SIMD_ENABLED  // Include header with intrinsics.
     #include <x86intrin.h>
```
h-2: don't you have to include this **before** checking the `__AVX2__` macro?

rrahn: I always assumed, that the compiler sets these flags, and that they are not dependent on an included header.

h-2: I just double-checked and you are right.

Issue #1830, opened 2016-08-12
Extension list in argument parser gets really long with compressions


h-2:
e.g:
```
    -q, --query IN
          Query sequences. Valid filetypes are: .fq, .fq.bgzf, .fq.gz, .fq.bz2, .fastq, .fastq.bgzf, .fastq.gz,
          .fastq.bz2, .fa, .fa.bgzf, .fa.gz, .fa.bz2, .fasta, .fasta.bgzf, .fasta.gz, .fasta.bz2, .faa, .faa.bgzf,
          .faa.gz, .faa.bz2, .ffn, .ffn.bgzf, .ffn.gz, .ffn.bz2, .fna, .fna.bgzf, .fna.gz, .fna.bz2, .frn, .frn.bgzf,
          .frn.gz, .frn.bz2, .embl, .embl.bgzf, .embl.gz, .embl.bz2, .gbk, .gbk.bgzf, .gbk.gz, .gbk.bz2, .raw,
          .raw.bgzf, .raw.gz, and .raw.bz2.
```
Could we instead display something like:
```
    -q, --query IN
          Query sequences. Valid filetypes are: .fq[.bgzf|.gz|.bz2], .fastq[.bgzf|.gz|.bz2], .fa[.bgzf|.gz|.bz2], 
          .fasta[.bgzf|.gz|.bz2], .faa[.bgzf|.gz|.bz2], .ffn[.bgzf|.gz|.bz2], .fna[.bgzf|.gz|.bz2], .frn[.bgzf|.gz|.bz2],
          .embl[.bgzf|.gz|.bz2], .gbk[.bgzf|.gz|.bz2], .raw[.bgzf|.gz|.bz2]
```

Or:
```
    -q, --query IN
          Query sequences. Valid filetypes are: .fq, .fastq, .fa, .fasta, .faa, .ffn, .fna, .frn, .embl, .gbk, .raw
          Additionally any of the listed extensions may be extended by .bgzf, .gz or .bz2 for transparent 
          (de)compression.
```

The last approach looks nicer, but it does have the problem that sometimes not all extensions have all compression formats, e.g. `.bam` cannot by extended by anything, because it already is compressed.

@rrahn @smehringer what do you think?

smehringer:
Do we always allow compressed files and want the compressed file extensions always to be shown or does the user needs to specify compression formats (/for each file format)?

rrahn:
[...]

PR #1843, opened 2016-08-31
[FEATURE] seperated file extensions and compression extensions.

smehringer:
Solves #1830


include/seqan/arg_parse/arg_parse_doc.h
```diff
@@ -38,6 +38,10 @@
 #include <seqan/arg_parse/tool_doc.h>
 #include <seqan/arg_parse/argument_parser.h>
 
+#include <regex>
+#include <vector>
+#include <algorithm>
+
```
h-2: Please move to toplevel include header

include/seqan/arg_parse/arg_parse_doc.h
```diff
@@ -729,6 +733,44 @@ inline void _addDefaultValues(std::string & text, ArgParseOption const & opt)
...
+    {
+        std::regex rgx("^(\\.)?([A-z]+)(\\.)?([A-z]+)?");
+        std::smatch result;
+
+        std::regex_search(validValues[i], result, rgx);
+
+        if (!result[4].str().empty())
+        {
+            comp_ext.push_back(result[4].str());
+            file_ext.push_back("." + result[2].str() + "[.x]");
```
h-2: since .x could actually be a valid extension, it is not immediately clear that it is meant as a placeholder. Could we use `.?` or `.*` or `.#` instead?

h-2:
nice use of c++ regular expressions :)

include/seqan/arg_parse/arg_parse_doc.h
smehringer: Sure. I think the asterisk would fit best. @rrahn what do you think?

rrahn: 👍

include/seqan/arg_parse/arg_parse_doc.h
smehringer: I did. Git just didn't recognize it.. :information_desk_person:

Issue #1834, opened 2016-08-24
"isOpen" function for FormattedFile


holtgrewe:
This would be super useful.

smehringer:
I gave it a try. Works with a little example for me. @holtgrewe @rrahn can you take a look at it?

holtgrewe:
What about adding a unit test?

smehringer:
Sure. Any suggestions to which existing tests this might fit?
Since ```formattedfile.h``` belongs to ```include/seqan/stream/```, I looked into the stream tests but I'm having a little trouble to understand all the test cases and figure out where to best place an isOpen-test.

smehringer:
I added unit tests for each of the formatted file IO's. (See PR)

h-2:
fixed by #1844

PR #1844, opened 2016-09-01
[FEATURE] isOpen() function for formatted file.

smehringer:
suggestion to #1834

@rrahn what do you think?

### DIFF
--- a/include/seqan/stream/formatted_file.h
+++ b/include/seqan/stream/formatted_file.h
@@ -718,8 +718,19 @@ inline bool open(FormattedFile<TFileFormat, TDirection, TSpec> & file,
 // ----------------------------------------------------------------------------
 // Function isOpen()
 // ----------------------------------------------------------------------------
+
+/*!
+ * @fn FormattedFile#isOpen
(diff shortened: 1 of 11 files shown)


include/seqan/stream/formatted_file.h
```diff
@@ -716,6 +716,15 @@ inline bool open(FormattedFile<TFileFormat, TDirection, TSpec> & file,
 }
 
 // ----------------------------------------------------------------------------
+// Function isOpen()
+// ----------------------------------------------------------------------------
+template <typename TFileFormat, typename TDirection, typename TSpec>
+inline bool isOpen(FormattedFile<TFileFormat, TDirection, TSpec> & file)
```
rrahn: Should be ```const &```

rrahn: And add documentation, in case we have the formatted file documented.

rrahn: 👍 😄

h-2:
+1

PS: The last commit should have had `[TEST] ...` label, only the first one was `[FEATURE] ...` :nail_care:

PR #1846, opened 2016-09-06
[INTERNAL] rename PAGESIZE variable

pohlmax:


h-2:
Description:

- this makes SeqAn compatible with the **musl libc** which is a generic c library, but also used by the emscripten compiler.

h-2:
Thanks for the PR @pohlmax, but can you please use `SEQAN_PAGESIZE` as replacement for `PAGESIZE` and not `pAGESIZE`?

h-2:
superseded by #1848

PR #1847, opened 2016-09-07
Add check before dereferencing iterator

rrwick:
I ran into a crash that I traced to the `getAlignmentStatistics` function in the `graph_align_tcoffee_library.h` file. This function was being called with an empty `matches` string and crashing when this line tried to dereference the iterator:
`TId id1 = sequenceId(*itFrag, 0);`

I should mention that I ran into this bug after making a number of modifications to SeqAn, so perhaps it can never happen in actual, unmodified SeqAn. But it seems like a safe change, so I thought I'll make this PR, in case you'd like to use it!

### DIFF
--- a/include/seqan/graph_msa/graph_align_tcoffee_library.h
+++ b/include/seqan/graph_msa/graph_align_tcoffee_library.h
@@ -543,7 +543,7 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
     if (length(matches) > 0)
     {
         TFragIter itFrag = begin(matches, Standard()) + from;
-        TFragIter itFragEnd = itFrag + to;
+        TFragIter itFragEnd = begin(matches, Standard()) + to;
 
         TId id1 = sequenceId(*itFrag, 0);
         TId id2 = sequenceId(*itFrag, 1);
--- a/include/seqan/graph_msa/graph_align_tcoffee_library.h
+++ b/include/seqan/graph_msa/graph_align_tcoffee_library.h
@@ -541,33 +541,36 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
     TSize maxId2 = 0;
     TSize matchMismatch_length = 0;
 
-    TFragIter itFrag = begin(matches, Standard());
-    TFragIter itFragEnd = itFrag;
-    itFrag += from;
-    itFragEnd += to;
-    TId id1 = sequenceId(*itFrag, 0);
-    TId id2 = sequenceId(*itFrag, 1);
-    TSize fragLen = 0;
-    TSize beginI = 0;
-    TSize beginJ = 0;
-    for(;itFrag != itFragEnd; ++itFrag) {
-        fragLen = fragmentLength(*itFrag, id1);
-        beginI = fragmentBegin(*itFrag, id1);
-        beginJ = fragmentBegin(*itFrag, id2);
-        if (beginI < minId1) minId1 = beginI;
-        if (beginJ < minId2) minId2 = beginJ;
-        if (beginI + fragLen > maxId1) maxId1 = beginI + fragLen;
-        if (beginJ + fragLen > maxId2) maxId2 = beginJ + fragLen;
-        typedef typename Infix<TString>::Type TInfix;
-        typedef typename Iterator<TInfix, Standard>::Type TInfixIter;
-        TInfix inf1 = label(*itFrag, str, id1);
-        TInfix inf2 = label(*itFrag, str, id2);
-        TInfixIter sIt1 = begin(inf1, Standard());
-        TInfixIter sIt2 = begin(inf2, Standard());
-        TInfixIter sIt1End = end(inf1, Standard());
-        matchMismatch_length += fragLen;
-        for(;sIt1 != sIt1End; ++sIt1, ++sIt2)
-            if ( (TAlphabet) *sIt1  == (TAlphabet) *sIt2) ++matchLength;
+    if (length(matches) > 0)
+    {
+        TFragIter itFrag = begin(matches, Standard());
+        TFragIter itFragEnd = itFrag;
+        itFrag += from;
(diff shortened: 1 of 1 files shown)


rrwick:
And I did try to complete the CLA agreement, but it gives me an error message: 'We're sorry, but something went wrong.'

h-2:
Thanks for the PR and sorry about the CLAHUB, it just seems to be really broken. We have deactivated the integration for now. It shouldn't be a problem, because the patch is very minimal anyway.

Since you have touch this code section, could you still improve the formatting to make it more readable? I will comment inline:

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -541,33 +541,36 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
...
+        TFragIter itFrag = begin(matches, Standard());
+        TFragIter itFragEnd = itFrag;
+        itFrag += from;
+        itFragEnd += to;
+        TId id1 = sequenceId(*itFrag, 0);
+        TId id2 = sequenceId(*itFrag, 1);
+        TSize fragLen = 0;
+        TSize beginI = 0;
+        TSize beginJ = 0;
+        for(;itFrag != itFragEnd; ++itFrag) {
```
h-2: space after `;` and `{` to beginning of next line

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -541,33 +541,36 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
...
-        TInfixIter sIt1End = end(inf1, Standard());
-        matchMismatch_length += fragLen;
-        for(;sIt1 != sIt1End; ++sIt1, ++sIt2)
-            if ( (TAlphabet) *sIt1  == (TAlphabet) *sIt2) ++matchLength;
+    if (length(matches) > 0)
+    {
+        TFragIter itFrag = begin(matches, Standard());
+        TFragIter itFragEnd = itFrag;
+        itFrag += from;
+        itFragEnd += to;
```
h-2: compact these four lines into two and introduce a blank line after them.

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -541,33 +541,36 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
...
+        TSize beginI = 0;
+        TSize beginJ = 0;
+        for(;itFrag != itFragEnd; ++itFrag) {
+            fragLen = fragmentLength(*itFrag, id1);
+            beginI = fragmentBegin(*itFrag, id1);
+            beginJ = fragmentBegin(*itFrag, id2);
+            if (beginI < minId1) minId1 = beginI;
+            if (beginJ < minId2) minId2 = beginJ;
+            if (beginI + fragLen > maxId1) maxId1 = beginI + fragLen;
+            if (beginJ + fragLen > maxId2) maxId2 = beginJ + fragLen;
```
h-2: insert newlines after the conditions in the previous four statements...

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -541,33 +541,36 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
...
+            typedef typename Infix<TString>::Type TInfix;
```
h-2: blank line before typedef

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -541,33 +541,36 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
...
+            typedef typename Iterator<TInfix, Standard>::Type TInfixIter;
+            TInfix inf1 = label(*itFrag, str, id1);
+            TInfix inf2 = label(*itFrag, str, id2);
+            TInfixIter sIt1 = begin(inf1, Standard());
+            TInfixIter sIt2 = begin(inf2, Standard());
+            TInfixIter sIt1End = end(inf1, Standard());
+            matchMismatch_length += fragLen;
+            for(;sIt1 != sIt1End; ++sIt1, ++sIt2)
+                if ( (TAlphabet) *sIt1  == (TAlphabet) *sIt2) ++matchLength;
```
h-2: insert newline...

rrwick:
Thanks - I made the readability changes you suggested!

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -543,23 +542,29 @@ getAlignmentStatistics(String<TFragment, TSpec1> const& matches,
 
     if (length(matches) > 0)
     {
-        TFragIter itFrag = begin(matches, Standard());
-        TFragIter itFragEnd = itFrag;
-        itFrag += from;
-        itFragEnd += to;
+        TFragIter itFrag = begin(matches, Standard()) + from;
+        TFragIter itFragEnd = itFrag + to;
+
```
h-2: This changes actually had impact, as itFragEnd is now different (as you can see some tests failed). Can you change this line to:
```c++
TFragIter itFragEnd = begin(matches, Standard()) + to;
```

That should restore the correct behaviour.

rrwick:
Whoops - I was a bit too hasty with that last update! Thanks for spotting that and it's fixed now.

h-2:
Great! One more thing: we have re-included the clahub, can you check if signing it works? If it doesn't I will merge anyway, but it would be good to know if it works now :)

Thanks!

h-2:
@rrwick friendly ping :smile:

rrwick:
Thanks for the ping... I had a round trip time of about 200,000,000 milliseconds, so I don't think I'd make a very good web server :flushed:

The CLAHub thing worked, so it should all be good to go now!

PR #1848, opened 2016-09-07
[INTERNAL] rename PAGESIZE to SEQAN_PAGESIZE for emscripten compiler

pohlmax:


h-2:
This should make SeqAn compatible with MUSL LIBC.

ps: @pohlmax you could have just pushed to the same branch as before, and it would have updated the previous pull request [so you don't need to make a new one]. Just for the future :)

Issue #1841, opened 2016-08-30
SeqAn::T-Coffee library input


bestrauc:
Hi,
there seems to be a small bug in Seqan's T-Coffee implementation when you supply an externally generated T-Coffee library that specifies alignments and residue pairs that you want T-Coffee to use. The `T-COFFEE_LIB_FORMAT_01` I'm referring to is specified [here](http://www.tcoffee.org/Documentation/t_coffee/t_coffee_doc.pdf) on page 71 and as you can see, it ends with `!<space>SEQ_1_TO_N `, where SeqAn::T-Coffee loops endlessly, at least for my [example files](https://github.com/seqan/seqan/files/445002/example.tar.gz).

The issue occurred for me while trying to run `seqan_tcoffee -s rna.fasta -l rna.lib` .

I fixed the issue for me by inserting a break statement after [this line](https://github.com/seqan/seqan/blob/master/include/seqan/graph_msa/graph_align_tcoffee_io.h#L197), else it doesn't skip past the "!" and the "!" indicates the last line anyway, so it should be okay to exit the loop.

h-2:
@joergi-w can you have a look at this?

joergi-w:
Sure. Thank you @bestrauc for pointing that out. 
I suggest skipping comment lines (starting with '!') completely.

joergi-w:
The issue was solved in Pull Request #1850 and can be closed.

h-2:
@bestrauc If the issue persists, please reopen.

PR #1849, opened 2016-09-07
[FIX-#1841] skip comment lines in tcoffee_io

joergi-w:


h-2:
We only change these things in develop nowadays!

PR #1850, opened 2016-09-07
[FIX-#1841] skip comment lines in tcoffee_io

joergi-w:


tests/graph_msa/test_graph_tcoffee.h
```diff
@@ -201,7 +201,7 @@ void Test_ExternalLibraries() {
     strmRead.open(fileName.c_str(), std::ios_base::in | std::ios_base::binary);
     read(strmRead,seqSet,nameSet,TCoffeeLib());
     strmRead.close();
-    std::cout << g << std::endl;
+    //std::cout << g << std::endl;
```
h-2: Is this debugging code or what is its purpose?

joergi-w: The code was there before and polluted the test output. So I just commented it out, as there are several similar comments in that file.

PR #1853, opened 2016-09-19
[FIX] fix open/save overloads, add generic directional goDown, add representative()

h-2:
@cpockrandt please also add tests for this stuff :) thanks!


h-2:
> Should we really allow the developer to call goDown(it, Rev()) on a unidirectional index? I agree to allow it for the forward tag Fwd(). But offering an interface that is contradictory to its functionality doesn't make sense to me.

I would add it, but clearly state in the (future) documentation how this behaves. The thing is that we want to be able to use the same interface, otherwise it is really difficult to use the bi-directional index as "drop-in" replacement for the unidirectional index.

h-2:
@cpockrandt do you agree with this now?

cpockrandt:
@h-2 :+1:

PR #1858, opened 2016-09-21
[FIX] BiFM: Reversed StringSet had wrong limits

cpockrandt:
I kept the RevTextFibre for now in case we want to switch back to a ModifiedString.


h-2:
1. a test fails, please check that.
  2. please adapt the construction code so that it clears the revText after construction, because it is not the expected behaviour for the index to contain the text. And please add source code comments that this behaviour will be changed in the future.

include/seqan/index/index_bifm.h
```diff
@@ -56,6 +56,8 @@ SEQAN_CONCEPT_IMPL((Index<TText, BidirectionalIndex<FMIndex<TSpec, TConfig> > >
 template <typename TText, typename TSpec, typename TConfig>
 inline bool indexCreate(Index<TText, BidirectionalIndex<FMIndex<TSpec, TConfig> > > & index)
 {
+    indexText(index.rev) = indexText(index.fwd);
+    reverse(indexText(index.rev));
```
h-2: If the index was created with the strings as argument, it will be reversed twice, see above.

PR #1870, opened 2016-10-06
[FIX] mark deprecated code and fix corresponding warnings

smehringer:
fixes #1819


apps/fiona/compute_gain.cpp
```diff
@@ -991,7 +991,7 @@ int main(int argc, char const ** argv)
     bool error = false;
 
     // Format recognition tag used for sequence file I/O.
-    seqan::AutoSeqFormat seqFormatTag;
+    // seqan::AutoSeqFormat seqFormatTag; // deprecated
```
h-2: Whats up with this? Is it not used anywhere?

apps/fiona/fiona.cpp
```diff
@@ -1254,7 +1254,7 @@ getFoundCorrections(
     unsigned numCorrections = 0;
 
     TValue currentPos = firstCorrectionForRead[erroneousReadId];
-    while (currentPos != maxValue<TValue>())
+    while (currentPos != MaxValue<TValue>::VALUE)
```
h-2: Why not `std::numeric_limits<TValue>::max()` as well?

apps/razers/razers_spliced.h
```diff
@@ -991,7 +991,7 @@ findBestSplitPosition(String<Pair<TScore,int> > & maxColsL,
 	::std::cout << "findBestSplitEditReverse\n";
 #endif
 
-	TScore maxSum = minValue<TScore>();
+	TScore maxSum = MinValue<TScore>::VALUE;
```
h-2: see above

apps/searchjoin/db.h
```diff
@@ -130,25 +130,25 @@ namespace seqan
 template <>
 struct Fibre<TDbDnaSaSmall, FibreSA>
 {
-    typedef String<Pair<unsigned int, unsigned char, BitPacked<24, 8> >, DefaultIndexStringSpec<TDbDnaSaSmall>::Type> Type;
+    typedef String<Pair<unsigned int, unsigned char, BitPacked<24, 8> >, StringSpec<TDbDnaSaSmall>::Type> Type;
```
h-2: how is this related to anything?

include/seqan/align/evaluate_alignment.h
```diff
@@ -318,6 +318,7 @@ TScoreVal computeAlignmentStats(AlignmentStats & stats,
 
 // NOTE(h-2): this interface is deprecated. Don't use it.
 template <typename TGaps, typename TAlignSpec, typename TScoreVal, typename TScoreSpec>
+[[deprecated]]
```
h-2: add a reason and remove my comment

include/seqan/basic/metaprogramming_enable_if.h
```diff
@@ -187,16 +187,16 @@ struct DisableIf<False, T>
 //    }
 
 template <bool b, typename T>
-struct EnableIf2;
+struct /*[[deprecated("Will be renamed to EnableIfC.")]]*/ EnableIf2;
```
h-2: ?

include/seqan/index/index_qgram.h
```diff
@@ -1404,6 +1404,7 @@ typename TBucketMap,
 typename TText,
 typename TShape,
 typename TStepSize >
+/*[[deprecated("Use createQGramIndex(index) instead.")]]*/
```
h-2: ?

include/seqan/sequence/string_set_base.h
```diff
@@ -1477,6 +1477,7 @@ return iter(me, length(me), tag);
  */
 
 template<typename TString, typename TSpec, typename TId>
+// [[deprecated("Use the subscript operator (operator[]) instead.")]]
```
h-2: ?

include/seqan/sequence/string_set_dependent_tight.h
```diff
@@ -91,7 +91,7 @@ struct Dependent;
 
 // StringSet with individual sequences in a tight string of string pointers and corr. IDs
 template <typename TString>
-class StringSet<TString, Dependent<Tight> >
+class /*[[deprecated]]*/ StringSet<TString, Dependent<Tight> >
```
h-2: ?

include/seqan/store/store_contig.h
```diff
@@ -192,13 +192,13 @@ struct ContigFile
     static const TId INVALID_ID;
 
     CharString        fileName;
-    AutoSeqFormat    format;
+    //AutoSeqFormat    format; // is deprecated
```
h-2: ?

h-2: If everything works without it, why not remove it?

tests/basic/test_basic_alphabet_math.h
```diff
@@ -61,34 +61,37 @@ struct MyNumber_
...
 
-inline MyNumber_ const &
-infimumValueImpl(MyNumber_ * /*tag*/)
+namespace seqan
 {
-    static const MyNumber_ result(-1);
-    return result;
-}
+template <>
+struct MinValue<MyNumber_>
```
h-2: Whats up with these Metafunctions, why not STL?

tests/sequence/test_sequence.h
```diff
@@ -1400,39 +1400,41 @@ std::cout << __LINE__ << std::endl;
...
+//    TValue dummy_ = 'T';
+//    TValue & value_ = dummy_;
+//    SEQAN_ASSERT_EQ(value_, TValue('T'));
+//
+//    value_ = getValue(string, 0);
+//    SEQAN_ASSERT_EQ(value_, TValue('A'));
+//
+//    value_ = 'T';
+//    SEQAN_ASSERT_EQ(getValue(string, 0), TValue('A'));
+//}
```
h-2: If it is not used anymore, remove it.

tests/sequence/test_string_set.h
```diff
@@ -1353,50 +1353,52 @@ SEQAN_TYPED_TEST(StringSetTestCommon, Front)
...
+//    SEQAN_ASSERT(TString(getValue(stringSet, 1)) == TString(str2));
+//}
+//
+//// TODO(singer): No appendValue for string sets of packed strings
+//template <typename TValue, typename TStringSetSpec>
+//void testStringSetGetValue(StringSet<String<TValue, Packed<> >, TStringSetSpec> & /*Tag*/) {}
+//template <typename TValue, typename TStringSetSpec>
+//void testStringSetGetValue(StringSet<String<TValue, Packed<> >, TStringSetSpec> const & /*Tag*/) {}
+//template <typename TValue, typename TStringSetSpec>
+//void testStringSetGetValue(StringSet<String<TValue, Array<100> >, TStringSetSpec> & /*Tag*/) {}
```
h-2: ...

tests/store/test_store_io_bam.h
```diff
@@ -44,15 +44,15 @@ SEQAN_DEFINE_TEST(test_store_io_bam_read)
...
     strcat(fastaBuffer, "/projects/tests/store/toy.fa");
     // Construct file name to SAM file.
     char samBuffer[1023];
-    strcpy(samBuffer, SEQAN_PATH_TO_ROOT());
+    strcpy(samBuffer, getAbsolutePath(""));
     strcat(samBuffer, "/projects/tests/store/toy.sam");
     // Construct file name to BAM file.
     char bamBuffer[1023];
-    strcpy(bamBuffer, SEQAN_PATH_TO_ROOT());
+    strcpy(bamBuffer, getAbsolutePath(""));
```
h-2: ?

apps/fiona/compute_gain.cpp
smehringer: Not that I can tell. I originally commented it out to see any warnings or error that might follow but nothing happend :no_mouth: ..

apps/fiona/fiona.cpp
smehringer: Not every type is a primitive type that can be resolved with numeric_limits. For example When a DNA/DNA5 alphabet is used the MaxValue<> function will return 'T'/'N'. To avoid any errors I only replaced primitive types with numeric_limits and used the seqan meta function for generic types.

include/seqan/basic/metaprogramming_enable_if.h
smehringer: That's what the comments in this file said.. I thought to add the deprecation as a comment to be commented in as soon as EnableIfC is available.

include/seqan/sequence/string_set_dependent_tight.h
smehringer: The tight dependent stringset is deprecated but it is still the default and used in several applications. One cannot change the default to generic since it introduces several warnings and failed tests and as far as I can tell will require quite the amount of work to change. But it should become deprecated as soon as possible (hence the comment).

apps/fiona/fiona.cpp
h-2: wow, I didn't know it was used like that. Are you sure this behaviour is intended? Maybe they meant `unknownValue<T>()` instead?

include/seqan/sequence/string_set_dependent_tight.h
h-2: Hm, this is a difficult situation. Making the default deprecated, but not changing the default seems inconsistent to me :confused:

tests/store/test_store_io_bam.h
h-2: No, the question was why there is this test that it not used :D

include/seqan/sequence/string_set_dependent_tight.h
smehringer: Well I can remove the deprecation of course.. but as I said changing the default will require some thought and work. Maybe we can talk this through on monday.

include/seqan/basic/metaprogramming_enable_if.h
smehringer: OK. :wastebasket:

apps/searchjoin/db.h
smehringer: ```DefaultIndexStringSpec``` is marked deprecated in favor of StringSpec here: https://github.com/seqan/seqan/blob/master/include/seqan/index/index_base.h#L197
so I replaced it by StringSpec (and added the deprecation to the documentation).

include/seqan/store/store_contig.h
smehringer: Well I have no idea where this is used, since commenting it out did not lead to any warnings or errors, so I didn't know if I could delete it?

include/seqan/index/index_qgram.h
smehringer: The function ```createQGramIndex(sa, dir , ...)``` is deprecated but used by the app searchjoin. It requires some work in my POV to change the usage so thought leave the comment such that in seqan3 where we might change or remove the application anyway we can include the deprecation macro then.
Shall I remove the comment?

include/seqan/sequence/string_set_base.h
smehringer: The documentation marks those functions (```*ById()```) as deprecated with the comment that the ID is no different than the position. When I tried to replace those functions I encountered several errors and it seems this is related to the dependent tight StringSet. So I commented those macros out too, since changes to the stringset will probably not happen until seqan3.

apps/searchjoin/db.h
h-2: ok

include/seqan/index/index_qgram.h
h-2: no, but please include this reason in the source code comment!

include/seqan/sequence/string_set_base.h
h-2: Hm, so the ById() functions are not deprecated now?

include/seqan/sequence/string_set_dependent_tight.h
h-2: lets discuss this again

include/seqan/store/store_contig.h
h-2: commenting out and removing it has exactly the same effect :nerd_face: 
so please delete it!

include/seqan/sequence/string_set_base.h
smehringer: According to the documentation yes, in my POV (and concerning the macro) no.

include/seqan/index/index_qgram.h
smehringer: ok

tests/store/test_store_io_bam.h
smehringer: I don't know :confused:. I regexed the replacement so I accidentally changed unused code. 

There is a ```TODO(holtgrew): Actually check contents.``` at the end so maybe he knows why..

h-2:
Can you double-check which of the following is true by making a small test:
  - [ ] `Dependant<Tight>` cannot be used reliably with `[]` 
  - [ ] `Dependant<Generous>` cannot be used reliably with `[]`

If both are true, we don't deprecate anything. If only one is true, we deprecate the other and make the non-deprecated version default.

smehringer:
While looking into the Deendent SringSet I noticed the follwoing


Dependent Tight - erase
---------------------------------
Stated [here](http://docs.seqan.de/seqan/master/class_StringSet.html#StringSet#removeValueById), the removeValueById is deprecated in favour of erase. Erase does not work on the Dependent String Set though. The follwing code snippet does not work (or did I use erase the wrong way?):

```
    StringSet<DnaString, Dependent<Tight> > stringSet;
    DnaString str0 = "TATA";
    DnaString str1 = "CGCG";
    appendValue(stringSet, str0);
    appendValue(stringSet, str1);
    
    std::cout << stringSet[0] << std::endl;
    erase(stringSet, 0);
    std::cout << stringSet[0] << std::endl;

```
It gives the error
```
/home/svenja/Projekte/AlBi_work/seqan/include/seqan/sequence/sequence_interface.h:1207:23: error: no matching function for call to ‘resizeSpace(seqan::StringSet<seqan::String<seqan::SimpleType<unsigned char, seqan::Dna_>, seqan::Alloc<> >, seqan::Dependent<seqan::Tag<seqan::TagInsist_> > >&, int&, int&, int&, seqan::Tag<seqan::TagGenerous_>&, seqan::DefaultOverflowExplicit<seqan::StringSet<seqan::String<seqan::SimpleType<unsigned char, seqan::Dna_>, seqan::Alloc<> >, seqan::Dependent<seqan::Tag<seqan::TagInsist_> > > >::Type)’
     return resizeSpace(me, size, pos_begin, pos_end, limit, typename DefaultOverflowExplicit<
                       ^
``` 
But it works when I replace erase with ```removeValueById()```.

Dependent Tight - [] operator
---------------------------------------
Considering the same String set from above
followed by the following lines of code:
```
    std::cout << stringSet[0] << " " << stringSet[1] << std::endl;
    std::cout << getValueById(stringSet, 0) << " " << getValueById(stringSet, 1) << std::endl;
    
    removeValueById(stringSet, 0);
    std::cout << "after removing string 0:\n";
    std::cout << stringSet[0] << " " << stringSet[1] << std::endl;
    
    std::cout << getValueById(stringSet, 1) << std::endl;
    std::cout << getValueById(stringSet, 0) << std::endl;
```
outputs this:

```
TATA CGCG
TATA CGCG
after removing string 0:
CGCG CGCG
CGCG
[1]    4237 segmentation fault  ./bin/demo_test
```
I interpret this as non reliable. I don't even know why stringSet[1] does exist ?? Since the Tight structure should move str1 to the position of str0... 

Dependent Generous - [] operator
----------------------------------------------
Given the same stringSet but with the TSpec Dependent<Generous.
I altered the output slightly to better display the behaviour:
```
    std::cout << stringSet[0] << " " << stringSet[1] << std::endl;
    std::cout << getValueById(stringSet, 0) << " " << getValueById(stringSet, 1) << std::endl;
    
    removeValueById(stringSet, 0);
    std::cout << "after removing string 0:\n";
    std::cout << stringSet[0] << std::endl;
    std::cout << stringSet[1] << std::endl;
    
    std::cout << getValueById(stringSet, 0) << std::endl;
    std::cout << getValueById(stringSet, 1) << std::endl;
```
output:
```
TATA CGCG
TATA CGCG
after removing string 0:
CGCG  // stringSet[0]
              // stringSet[1]
CGCG // getValueById[0]
             // getValueById[1]
```
 Note: The Generous Dependent StringSet returns an empty string on any out-of-range or deleted-value error.

I don't know if this is reliable? I depends how you specify the behaviour you want to provide with the []-operator. (But it certainly does not represent getValueById() as stated in the documentation...)

Dependent Generous - erase
----------------------------------------
erase works the same as removeValueById in this case.

smehringer:
@rrahn maybe you also want to take a look at my last comment on the behaviour of Dependent StringSets.

h-2:
@smehringer 
Except that erase doesn't work, all of the other behaviour seems to be correct, or not? If you have two elements in the StringSet, and you delete the first one, than `stringSet[0]` should return the former second one. This works.
`stringSet[1]` on the other hand is now undefined behaviour. In Debug mode it should fail an assertion, but other than that the results are expected (you still get the original value in some cases, becaue it's still there, only the length of the internal data structure was reduced).
To be it looks like it works correctly, and I don't understand what the difference between both stringsets is. But if you can make `erase()` work and `[]` work than we don't need to deprecate anything.

smehringer:
hmm.. in my POV the usual behaviour of any vector is not equal to the dependent tight stringset without the byId() functions. 

You would have to manually keep track of the position changes of your sequences: 
E.g.  
- you might found a pattern in sequence 1 and it is saved as ```hit1(seq1, pos5)``` , 
- then you delete seq 0 because there was no pattern, 
- but then you would have to manualy change your hit to ```hit1(seq0, pos5)```.

I think that's why the tight stringset keeps a map and has the functions *byId() such that one can still work with it like a normal vector. This contradicts the []-operator. Do you see what I mean? It all depends on the definition of "correct behaviour", which I'm not sure of. What kind of behaviour do we really want to provide?

so considering the just said, I don't know what you mean by 

> if you can make ```[]``` work 

make it work in which way?

h-2:
From my understanding
`StringSet<Dependant<Tight>>`: store pointers to other strings; if `i` is deleted, than the indexes are out-of-sync, but that's ok
`StringSet<Dependant<Generous>>`: store pointers to other strings; if 'i' is deleted, than querying for i in the future will return empty string

This behaviour should work with `erase()` and `operator[]`. I think we are pretty close. If this works, I don't think the tight stringset should be deprecated.

rrahn:
@h-2 and @smehringer funnily the erase function is not implemented for dependent string sets.
So it is not likely to work 😨

h-2:
@rrahn that that's the thing that needs to be fixed :)

include/seqan/basic/alphabet_adapt_builtins.h
```diff
@@ -203,27 +208,31 @@ supremumValueImpl(float *)
 // ----------------------------------------------------------------------------
 
 template <typename T>
+[[deprecated("Use MinValue instead.")]]
```
h-2: Recommend numeric_limits not our Metafunctions.

include/seqan/basic/alphabet_math.h
```diff
@@ -220,26 +220,26 @@ template <typename T> inline T const & supremumValueImpl(T *);
...
 
 template <typename T>
-inline T const &
+[[deprecated("Use MaxValue instead.")]]
+inline T
 maxValue()
 {
-    T * _tag = 0;
-    return supremumValueImpl(_tag);
+    return MaxValue<T>::VALUE;
```
h-2: directly use the numeric_limits and also recommend them in the deprecation notice

include/seqan/basic/alphabet_simple_type.h
```diff
@@ -319,7 +380,7 @@ struct MinValue<SimpleType<TValue, TSpec> >
 };
 
 template <typename TValue, typename TSpec>
-const SimpleType<TValue, TSpec> MinValue<SimpleType<TValue, TSpec> >::VALUE = SimpleType<TValue, TSpec>(0);
+const SimpleType<TValue, TSpec> MinValue<SimpleType<TValue, TSpec> >::VALUE = std::numeric_limits<SimpleType<TValue, TSpec>>::min();
```
h-2: Please check whether these specializations can be remove now, also MaxValue, MinValue, infimumValueImpl(), supremumValueImpl()

smehringer: when removing the MIn/MaxValue specialization for SimpleType I end up with the following error:

```
/home/mi/smehringer/AGABI/seqan/include/seqan/basic/alphabet_math.h:78:49: error: ambiguous overload for ‘operator<<’ (operand types are ‘seqan::SimpleType<unsigned char, seqan::Dna_>’ and ‘int’)
 const T MaximumValueSigned_<T>::VALUE = ((((T)1 << (BitsPerValue<T>::VALUE - 2)) - 1) << 1) + 1;
```
I don't quite understand what is the problem here.

h-2:
@smehringer Do we still have open questions here? Also it needs to be rebased since there are conflicts.

include/seqan/basic/alphabet_simple_type.h
smehringer: @h-2 There is an outstanding decision about deleting some code here.

h-2:
@rrahn ok, lets move this to 2.4

h-2:
@smehringer This was pretty complete, right? Can you rebase it? Than I will do a final review and merge.

smehringer:
Uf, it feels like ages that I worked on this. I remember that this was tedious, but I think I only included the deprecation macro on functions/classes that did not introduce warnings or errors so I guess it's save to merge. I'll try to review it on the weekend.

h-2:
Any news on this? We will likely do a release next week and it would be great to have this in. You can reduce it to whatever subset doesn't create any problems :)

smehringer:
I resolved rebasing issues and get no warning or failed tests on my system. Let's see what CI says. 
I think it was complete when we left it (to rot) some month ago so I think it's save since although many changes, most of them are simple replacements.

About the commit history: I know we usually squash it, but the commit messages actually helped me here to recover what I was doing... so maybe, because it is such a large PR we shouldn't squash it (or maybe not all of it) @h-2 what do you think?

h-2:
> About the commit history: I know we usually squash it, but the commit messages actually helped me here to recover what I was doing... so maybe, because it is such a large PR we shouldn't squash it (or maybe not all of it) @h-2 what do you think?

We don't squash per se and indeed it is better to have a proper history...

To have two commits for 
```
[INTERNAL] Replaced deprecated usage of getIdByName().
[INTERNAL] Replaced deprecated usage of appendName(). 
```
is totally fine, but the problem is that there is lots of back-and-forth a la "UNDO", "revert", "remove ... again...", "FIX problem introduced previously ..." these would have to be cleaned up (and this is nothing I can do from the interface).
If you want to clean the history up, I recommend:
  1. make a copy of your branch and switch to it
  2. do an interactive rebase on the current develop
  3. in the history editor, re-order all the commits so that related ones are "en-bloc"
  4. mark all but the first one of a "bloc" as fixup (f in the first) column
  5. pray that it doesn't introduce new consistency errors and close the editor
    * → have a nice history where duplicates a.s.o are merged; *or*
    * → have to resolve some conflicts

smehringer:
Only one revert is missing, but I wanted to push the changes so that didn't cause troubles.
Can I just delete the original and it's reverting commit ?

smehringer:
Sorry for spamming travis. There is this one commit (322a243fdac349506b0c9c72df0ff8d24aee0f98) that snuck in my PR and I don't know why. When I interactivly rebase and delete the commit, I change the commit messages of all commits that happened in between from other authors..
@h-2 what do I do?

h-2:
> When I interactivly rebase and delete the commit, I change the commit messages of all commits that happened in between from other authors..

What do you mean, you change the commit messages? The commit IDs will change, but the messages are independent. And what do you mean by "commits from other authors",  all of them are from you!?

smehringer:
Yes it will only alter the commit id, marking me as co-author. is that fine?

h-2:
Hm, I still don't understand this. You are the author of all the commits in this PR anyway, how can you become co-author of other commits?
Do you have to time chat some time today?

smehringer:
sorry. I didn't have time today. This is what happens. I deleted one commit ([FIX] Corrected cmake argument to disable the version check.).. and now the history is screwed up..
I kept the prior version locally and can revert this. 
I'll also take my laptop to work tomorrow so we can talk about it.

smehringer:
Ok done. @h-2 Please do a final review.

PR #1876, opened 2016-10-10
[INTERNAL, FIX-#699] Add check for possible score overflow in banded chain alignment

### LINKED ISSUE #699: Fatal Error in Banded Chain Alignment Traceback

rrahn:
Checks if a score overflow is likely for a given alignment sub matrix in the banded chain alignment procedure.

Fixes #699.

### DIFF
--- a/include/seqan/journaled_set/score_biaffine.h
+++ b/include/seqan/journaled_set/score_biaffine.h
@@ -52,51 +52,44 @@ template <typename TScoreValue>
 class Score<TScoreValue, BiAffine>
 {
     public:
-    TScoreValue _match;
-    TScoreValue _mismatch;
+    TScoreValue data_match;
+    TScoreValue data_mismatch;
     TScoreValue _gapExtendHorizontal;
     TScoreValue _gapOpenHorizontal;
     TScoreValue _gapExtendVertical;
     TScoreValue _gapOpenVertical;
 
 
-    Score() : _match(0),
-              _mismatch(0),
+    Score() : data_match(0),
+              data_mismatch(0),
               _gapExtendHorizontal(0),
               _gapOpenHorizontal(0),
               _gapExtendVertical(0),
               _gapOpenVertical(0) {}
 
     Score(TScoreValue match, TScoreValue mismatch, TScoreValue gap) :
-                                    _match(match),
-                                    _mismatch(mismatch),
+                                    data_match(match),
+                                    data_mismatch(mismatch),
                                     _gapExtendHorizontal(gap),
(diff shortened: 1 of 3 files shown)


h-2:
@rrahn please do not rename the member variables just for the fun of it :sleepy:

rrahn:
It's not for the fun of it, but to use the global scoreMatch and scoreMismatch function on that score object. But I agree at least the other variables need to be renamed in the same scheme.

rrahn:
Ok, I reverted the renaming. Apparently, there where already setter and getter functions but not complete.

include/seqan/journaled_set/score_biaffine.h
```diff
@@ -90,13 +90,6 @@ class Score<TScoreValue, BiAffine>
...
                                                                         _gapExtendVertical(gapExtendVertical),
                                                                         _gapOpenVertical(gapOpenVertical) {}
-
-    Score(Score const & other) : _match(other._match),
-                                 _mismatch(other._mismatch),
-                                 _gapExtendHorizontal(other._gapExtendHorizontal),
-                                 _gapOpenHorizontal(other._gapOpenHorizontal),
-                                 _gapExtendVertical(other._gapExtendVertical),
-                                 _gapOpenVertical(other._gapOpenVertical) {}
 };
```
h-2: Why was this removed?

h-2: I think there definitely should be a copy constructor, but it could also be defaulted. TBH, if you change this you could as well make it an aggregate type and remove all the constructors. But this require changes in other places...

rrahn: No, the copy constructor will be defaulted since the custom constructs won't affect this.
Otherwise rule-of-5- applies 😫. Probably the best would be aggregate type, but then it is not consistent with other score types and I don't see the point of changing all this for this small class, which is used in only one place at the moment. The other constructors are for consistency with the Simple score class.

Issue #1861, opened 2016-09-25
Debian related package issues


mr-c:
Open Debian Issues:
- [x] CPU architectures
  - [x] amd64 
  - [x] arm64 
  - [x] armel : out of memory while building apps, disabled ~~WONTFIX because of GCC BUG: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64735~~
  - [x] armhf:  ~~using `-O3`, out of memory during yara_mapper `cc1plus: out of memory allocating 
1761548 bytes after a total of 66174976 bytes`~~
Builds on Ubuntu now, try again? https://launchpad.net/ubuntu/+source/seqan2/2.3.1+dfsg-5ubuntu2/+build/13316820 Nope, still not enough: `cc1plus: out of memory allocating 2174444 bytes after a total of 64348160 bytes`
  - [x] i386  
  - ~~mips : out of memory during yara_mapper even with `-O1 -mxgot`, also see https://github.com/seqan/seqan/pull/1887#issuecomment-255703564~~
  - [x] mips64el
  -  ~~mipsel : out of memory during yara_mapper even with `-O1 -mxgot`: `virtual memory exhausted: Cannot allocate memory`~~
  - [x] powerpc 
  - [x] ppc64el 
  - [x] s390x
  - [x] alpha: ~~[`internal compiler error: Segmentation fault` & `internal compiler error` (out of memory?)](https://buildd.debian.org/status/fetch.php?pkg=seqan2&arch=alpha&ver=2.3.0~1.20161019~af07bc1%2Bdfsg-4&stamp=1480555510)~~
  - [x] hppa 
  - [x] hurd-i386
  - [x] kfreebsd-amd64
  - [ ] kfreebsd-i386 : looks like out of memory
  - [x] m68k : [boost error?](https://buildd.debian.org/status/fetch.php?pkg=seqan2&arch=m68k&ver=2.3.0~1.20161019~af07bc1%2Bdfsg-4&stamp=1480590848) could be fixed by https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=846617
  - [x] powerpcspe: ~~waiting to build~~
  - [x] ppc64 
  - [x] sh4 ~~waiting to build~~
  - [x] sparc64 
  - [x] x32
- [x] `libseqan2-dev` contains the whole cmake folder from the repo, but should only contain `seqan-config.cmake`
[...]

Issue #830, opened 2015-02-05
Support ARM architecture


sagitter:
Hi all.

Are there incompatibilities of the seqan-2.0.0-rc1's tests on ARM architectures?

```
[  9%] Building CXX object tests/parallel/CMakeFiles/test_parallel.dir/test_parallel.cpp.o
cd /builddir/build/BUILD/seqan-seqan-v2.0.0-rc1/build/Release/tests/parallel && /usr/bin/c++   -DSEQAN_ENABLE_TESTING=1 -DSEQAN_HAS_EXECINFO=1 -DSEQAN_HAS_OPENMP=1 -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches  -march=armv7-a -mfpu=vfpv3-d16  -mfloat-abi=hard  -W -Wall -Wno-long-long -fstrict-aliasing -Wstrict-aliasing -pedantic -Wno-variadic-macros -fopenmp -std=c++11 -I/builddir/build/BUILD/seqan-seqan-v2.0.0-rc1/include    -o CMakeFiles/test_parallel.dir/test_parallel.cpp.o -c /builddir/build/BUILD/seqan-seqan-v2.0.0-rc1/tests/parallel/test_parallel.cpp
{standard input}: Assembler messages:
{standard input}:10611: Error: bad instruction `rep'
```

Full log here: https://kojipkgs.fedoraproject.org//work/tasks/3293/8833293/build.log

esiragusa:
Hi @sagitter. I guess that most of the library could work out of the box, but the parallel primitives won't... Do all SeqAn 1.4.x tests compile on ARM?

sagitter:
SeqAn 1.4.x did not compile tests: https://kojipkgs.fedoraproject.org//packages/seqan/1.4.2/9.fc22/data/logs/armv7hl/build.log, this problem was not established. Or am I wrong?

sagitter:
Same problem during compilation of application searchjoin (on ARM only)
```
[  7%] Building CXX object apps/searchjoin/CMakeFiles/s4_join.dir/join.cpp.o
[...]

PR #1881, opened 2016-10-11
[FIX,TEST] only run certain tests on Linux+x86/amd64

h-2:
This was responsible for some test failures on Debian and non-x86 architectures. See also #1861


h-2:
See also #830

rrahn:
ok.

Issue #1865, opened 2016-10-05
check implicit fallthroughs


h-2:
New compilers now warn if one has an implicit fallthrough, i.e.
```c++
switch(i)
{
    case 0:
        cout << "foo\n";
        break;
    case 1:
        cout << "bar\n";
        // falls through to case 2
        // generates warning on gcc7
    case 2:
        cout << "bax\n";
        break;
}
```
This is often a bug, so to explicitly state that this is desired behaviour, you can now add:
```c++
switch(i)
{
    case 0:
        cout << "foo\n";
        break;
    case 1:
        cout << "bar\n";
        [[fallthrough]]
        // falls through to case 2
        // generates no warning on gcc7
    case 2:
        cout << "bax\n";
        break;
}
```
This is formally a c++17 feature, but unknown attributes should be ignored on older versions so it should work anyway on c++14, too.

We need to 
* [x] check that the is the case for all our platforms (i.e. every supported compiler must not complain about the attribute)
* [x] add the attribute to our implicit fallthroughs in the library that currently produce warnings / investigate if the fallthrough is actually intended or maybe a bug

These are the current warnings:
http://cdash.seqan.de/viewBuildError.php?type=1&buildid=195900

joergi-w:
Issue was solved with Pull Request #1882 and can be closed.

h-2:
@joergi-w  Even the second pull request introduced new warnings. Please resolve them and in the future always check the cdash after one of your PRs has been merged to see its effect on other platforms! Thanks!

joergi-w:
Now the warnings are gone, so we can close this issue.

PR #1882, opened 2016-10-12
[FIX-#1865] set fallthrough attribute

joergi-w:


apps/bs_tools/casbar.h
```diff
@@ -722,13 +722,13 @@ int readMatchesFromSamBam(
...
-                       // case 'S': std::cout << 'S' << record.cigar[i].count;
-                       case 'X': std::cout << 'X' << record.cigar[i].count;
+                        case 'D': std::cout << 'D' << record.cigar[i].count; break;
+                       // case 'N': std::cout << 'N' << record.cigar[i].count; break;
+                       // case 'P': std::cout << 'P' << record.cigar[i].count; break;
+                        case 'I': std::cout << 'I' << record.cigar[i].count; break;
+                        case 'M': std::cout << 'M' << record.cigar[i].count; break;
+                       // case 'S': std::cout << 'S' << record.cigar[i].count; break;
+                       case 'X': std::cout << 'X' << record.cigar[i].count; break;
                     }
```
h-2: wow, this really looks like a bug :open_mouth:

apps/bs_tools/casbar_score.h
```diff
@@ -188,6 +188,7 @@ assignTargetFreqs(Score<TValue, BsTagList<TBsProfileScore, TModel, TCellDescript
                 me.targetFreqs[i][1] += (estMethLevel/sum)/4.0;         // C
                 me.targetFreqs[i][2] += (1.0/sum)/4.0;                  // G
                 me.targetFreqs[i][3] += ((1.0-estMethLevel)/sum)/4.0;   // T
+                SEQAN_FALLTHROUGH;
```
h-2: Hm, not sure about this one.

apps/dfi/dfi.cpp
```diff
@@ -836,13 +836,15 @@ int main(int argc, const char *argv[])
 				case 1: return runDFI<PredMinFreq, PredMaxFreq, AminoAcid> (getArgumentValues(parser, 0), options.minFreq, options.maxFreq, options.maximal, options);
 				case 2: return runDFI<PredMinFreq, PredMaxFreq, Dna> (getArgumentValues(parser, 0), options.minFreq, options.maxFreq, options.maximal, options);
 			}
+            break;
```
h-2: Although the code does not follow our guidelines, please indent to the same level as the line above

apps/razers/readSimulator.h
```diff
@@ -305,6 +305,7 @@ void simulateReads(
 	
 					case SEQAN_DELETE:
 						++del;
+                        SEQAN_FALLTHROUGH;
```
h-2: same here: please use the same indention.

include/seqan/bam_io/cigar.h
```diff
@@ -149,14 +149,14 @@ uint32_t toBamCigarElement(CigarElement<TOperation, TCount> const & cigarElement
...
+        case '=': operation += 1; SEQAN_FALLTHROUGH;
+        case 'P': operation += 1; SEQAN_FALLTHROUGH;
+        case 'H': operation += 1; SEQAN_FALLTHROUGH;
+        case 'S': operation += 1; SEQAN_FALLTHROUGH;
+        case 'N': operation += 1; SEQAN_FALLTHROUGH;
+        case 'D': operation += 1; SEQAN_FALLTHROUGH;
+        case 'I': operation += 1; SEQAN_FALLTHROUGH;
         case 'M': break;
     }
     return (cigarElement.count << 4) | operation;
```
h-2: Wow, this looks really bad, I have no clue what the intended behaviour should be.
I just checked and this function is
* not documented
* not used anyway

I would propose adding the FALLTHROUGHs, but also adding a deprecated notice saying that the behaviour is not clear and that this function should not be used anymore.

include/seqan/bam_io/cigar.h
```diff
@@ -648,6 +648,7 @@ unsigned cigarToGapAnchorRead(TGaps & gaps, TCigarString const & cigar)
                 if (atBegin)
                     beginGaps += cigar[i].count;
                 insertGaps(it, cigar[i].count);
+                SEQAN_FALLTHROUGH;
```
h-2: this seems likely, but I am not sure...

include/seqan/blast/blast_tabular_in.h
```diff
@@ -548,6 +548,7 @@ _readField(BlastMatch<TAlignRow0, TAlignRow1, TPos, TQId, TSId> & match,
             break;
         case BlastMatchField<>::Enum::P_POS:
             match.alignStats.alignmentSimilarity = lexicalCast<double>(context._stringBuffer);
+            SEQAN_FALLTHROUGH;
```
h-2: this should be a `break;`

include/seqan/platform.h
```diff
@@ -398,6 +398,17 @@ typedef int8_t __int8;     // nolint
 // backwards compatibility
 #define SEQAN_UNUSED_TYPEDEF SEQAN_UNUSED
 
+// eliminate fallthrough warnings
+#if defined __has_cpp_attribute && __has_cpp_attribute(fallthrough)
+    #if __cplusplus < 201500
+        #define SEQAN_FALLTHROUGH [[gnu::fallthrough]]
```
h-2: On which compiler/std combinations is this supported?

include/seqan/platform.h
```diff
@@ -398,6 +398,17 @@ typedef int8_t __int8;     // nolint
...
+    #else
+        #define SEQAN_FALLTHROUGH [[fallthrough]]
```
h-2: please include the `;` in the macro since it is always needed anyways

include/seqan/index/pump_extender7.h
```diff
@@ -393,28 +393,28 @@ namespace seqan
...
-
+                SEQAN_FALLTHROUGH;
             case 1:
     /* 1 */                o1.i2[0] = o2.i2[1] = o3.i2[1] = o4.i2[2] = o5.i2[2] = o6.i2[2] = (*nameIn1).i2; ++nameIn1; o1.i1 = p; ++p;
                         cp___(o1,0,cp___(o2,1,cp___(o3,2,                                     *textIn))); ++textIn;
                 if (r >= 6) push(out6, o6);
                 if (r >= 5) push(out5, o5);
                 if (r >= 4) push(out124, o4);
-
+                SEQAN_FALLTHROUGH;
```
h-2: Wow, what is all of this? :open_mouth:

include/seqan/platform.h
h-2: I already checked, this is good.

include/seqan/platform.h
```diff
@@ -398,6 +398,17 @@ typedef int8_t __int8;     // nolint
 // backwards compatibility
 #define SEQAN_UNUSED_TYPEDEF SEQAN_UNUSED
 
+// eliminate fallthrough warnings
+#if defined __has_cpp_attribute && __has_cpp_attribute(fallthrough)
```
h-2: You need to make this in two consecutive if statements or it will fail on some of our platforms (see travis). It is probably easiest to do it like this:
```c++
#define SEQAN_FALLTHROUGH
#if defined(__has_cpp_attribute)
  #if __has_cpp_attribute(fallthrough)
    #undef SEQAN_FALLTHROUGH
    #if __cplusplus < 201500
      #define SEQAN_FALLTHROUGH [[gnu::fallthrough]];
    #else
      #define SEQAN_FALLTHROUGH [[fallthrough]];
    #endif
  #endif
#endif
```

Issue #1891, opened 2016-10-24
Make it possible to turn individual apps off


h-2:
Here we add our individual apps to the build:
https://github.com/seqan/seqan/blob/master/util/cmake/SeqAnBuildSystem.cmake#L182

It would be good if we add another check to the if-clause, something like:
```
(NOT ${${APP}_NO_BUILD})
```
that enables you to selectively block certain apps from being built. NOTE that I am not sure this syntax is supported by cmake. If it doesn't work, one might need to add a `SKIP_APP_LIST` and then check if the app-name is a member of the list.

@joergi-w can you look into this?

joergi-w:
see PR #1894

PR #1894, opened 2016-10-26
[FIX-#1891] suppress build of specified apps

joergi-w:


h-2:
Please join it with the surrounding if statement instead of nesting once more, otherwise :+1:

h-2:
PS: please use git commit --amend or git rebase so that you don't have multiple commits for small changes! thanks!

PR #1895, opened 2016-10-27
[FIX] displaying of extensions that contain numbers

h-2:
@smehringer can you have a look? Your changes mangled extension that contained numbers before...


smehringer:
Which extensions? Can you give an example? Because for example in yara_mapper (--output-format) .bz2 is somehow correctly displayed.

h-2:
> Which extensions? Can you give an example? 

`.accession2taxid` `.accession2taxid.gz`...
it just prints
`.accession`

 > Do we want do allow other sign beside characters and numbers? 

I think alphanumerical is enough.

Issue #1889, opened 2016-10-20
Empty CStyle Strings not properly initialized


fschopp:
When a `CStyle` string is constructed (or assigned), memory allocation happens in method [`allocate`](https://github.com/seqan/seqan/blob/seqan-v2.2.0/include/seqan/basic/allocator_interface.h#L230). Backtrace at that point:

```
#0  seqan::allocate<seqan::String<char, seqan::Tag<seqan::CStyle_> >, char, unsigned long, seqan::AllocateStorage_> (data=@0x7fff5fbff8e8: 0x0, count=1) at include/seqan/basic/allocator_interface.h:230
#1  0x00000001000029b0 in seqan::_reallocateStorage<char> (me=..., new_capacity=0) at include/seqan/sequence/string_cstyle.h:371
#2  0x000000010000280e in seqan::CreateArrayStringExpand_<seqan::Tag<seqan::TagExact_> >::create_<seqan::String<char, seqan::Tag<seqan::CStyle_> >, char const*> (target=..., source=@0x7fff5fbff7d8: 0x100004cf8 "") at include/seqan/sequence/string_cstyle.h:579
#3  0x00000001000027a0 in seqan::create<char, char, seqan::TagExact_> (target=..., source=0x100004cf8 "") at include/seqan/sequence/string_cstyle.h:758
#4  0x000000010000276d in seqan::assign<char, char, seqan::TagExact_> (target=..., source=0x100004cf8 "", tag=...) at include/seqan/sequence/string_cstyle.h:467
#5  0x000000010000273d in seqan::assign<char, seqan::Tag<seqan::CStyle_>, char [1]> (target=..., source=...) at include/seqan/sequence/string_base.h:992
#6  0x000000010000270f in seqan::String<char, seqan::Tag<seqan::CStyle_> >::String<char [1]> (this=0x7fff5fbff8e8, str=...) at include/seqan/sequence/string_cstyle.h:143
#7  0x000000010000153d in seqan::String<char, seqan::Tag<seqan::CStyle_> >::String<char [1]> (this=0x7fff5fbff8e8, str=...) at include/seqan/sequence/string_cstyle.h:142
```

If the string is empty, a memory block of 1 byte is allocated. However (and this is the bug), it is never initialized to zero.

h-2:
@kneubert can you have a look at this?

[...]

PR #1904, opened 2016-10-28
[FIX-#1889] proper initialization of empty CStyle Strings

kneubert:
removed the if clause in string_cstyle.h (line 585) because it results in incomplete initialization of a CStyle string with an empty string:

bildschirmfoto vom 2016-10-28 16 59 24


h-2:
@kneubert this PR contains the changes from #1903 as well. Can you recreate it, just with commits for this fix?
The easiest way to do this is to just checkout a new branch from upstream/develop and then `git cherry-pick 51d01d7`.

PR #1909, opened 2016-11-02
[FEATURE, INTERNAL] Altered Display of ArgParseOption in help page 

smehringer:
- corrected behaviour of getArgumentLabel()
- added function getArgumentType()
- added function getArgumentTypeAsString()
- Altered display of an ArgParseOption on the help page: 
... - the type of the option is displayed instead of the argumentLabel
... - instead of ```NUM NUM NUM``` the help page displays ```List of 3 NUMBER's```

### DIFF
--- a/include/seqan/arg_parse/arg_parse_doc.h
+++ b/include/seqan/arg_parse/arg_parse_doc.h
@@ -813,7 +813,7 @@ inline void _addValidValuesRestrictions(std::string & text, ArgParseArgument con
 
 inline void _addTypeAndListInfo(std::string & text, ArgParseArgument const & arg)
 {
-    std::string type = _getArgumentType(arg);
+    std::string type = getArgumentType(arg);
 
     // Write arguments to term line -> only exception, boolean flags
     if (!empty(type))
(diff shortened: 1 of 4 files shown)


h-2:
Can you post an example of an help page, before and after? I am a little confused by all the different changes now...
Are you sure this doesn't affect any publicly documented interfaces?

smehringer:
This was the display of options before:
testapp_develop

and this is the display afterwards:
testapp_develop_new

I don't know if this changes any public documentation since I don't think the display is documented?

otherwise we now have the new functions ```isArgumentType()``` and ```isArgumentTypeAsString()```
which I newly documented

and the change of  ```isArgumentLabel()``` which only now really behaves as the documentation states. (Documentation states: *returns the argument label*. Before my changes it returned the type, or even several types, if the label was empty.... which actually contradicted the documentation.)

Sorry for the amount of fixes. I had some problems with inheritance which I did not consider previously. :disappointed:

include/seqan/arg_parse/arg_parse_doc.h
```diff
@@ -846,13 +852,20 @@ inline void _addValidValuesRestrictions(std::string & text, ArgParseArgument con
...
+
 // ----------------------------------------------------------------------------
 // Function _addTypeAndListInfo()
 // ----------------------------------------------------------------------------
 
 inline void _addTypeAndListInfo(std::string & text, ArgParseArgument const & arg)
 {
-    std::string type = _getArgumentType(arg);
+    std::string type = getArgumentTypeAsString(arg);
+    for (auto & c: type) c = toupper(c);
```
h-2: insert linebreak

PR #1910, opened 2016-11-02
[FIX] Windows warning: EPR-dictionary: shift negative or too big

### LINKED ISSUE #1908: EPR changes create new warnings on windows
Please double-check the cdash yourself after larger PRs have been merged!

cpockrandt:
(hopefully) resolves #1908 

Thanks for hinting at it, h-2!

### DIFF
--- a/include/seqan/index/index_fm_rank_dictionary_levels.h
+++ b/include/seqan/index/index_fm_rank_dictionary_levels.h
@@ -507,8 +507,8 @@ struct RankDictionary<TValue, Levels<TSpec, TConfig> >
         // TODO(cpockrandt): make all bitmasks a const-expr
         constexpr unsigned maxValue = (1 << _BITS_PER_VALUE) - 1;
         constexpr unsigned padding = _BITS_PER_WORD % _BITS_PER_VALUE;
-	    // max() workaround for windows C4293 warning
-        constexpr TWordType paddingWord = (padding > 0) ? (1ull << (_BITS_PER_WORD - std::max(padding, 1u))) : 0ull;
+        // max() workaround for windows C4293 warning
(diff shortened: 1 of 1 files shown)


include/seqan/index/index_fm_rank_dictionary_levels.h
```diff
@@ -505,7 +505,8 @@ struct RankDictionary<TValue, Levels<TSpec, TConfig> >
         // TODO(cpockrandt): make all bitmasks a const-expr
         const unsigned maxValue = (1 << _BITS_PER_VALUE) - 1;
         const unsigned padding = _BITS_PER_WORD % _BITS_PER_VALUE;
```
h-2: don't use the max metafunction, use the stl's max function instead and use it up here. You could also make it constexpr

h-2:
Out of curiosity, have you tried making paddingWord constexpr as well and whether the warning is still created, when padding is constexpr when they all are constexpr? Because then the compiler should know all combinations at compile time....

cpockrandt:
I tried making ``paddingWord`` constexpr but it didn't work because of ``std::max`` I guess (see commit ``f9755d6``)

h-2:
std::max should be constexpr since c++14 so it should work, but I think it should even work without, because the windows compiler know at compile time that the value is >= 1. Can you try either way?

h-2:
@ping

cpockrandt:
@h-2 Introducing ``constexpr`` without ``max()`` does not work on Windows

h-2:
@cpockrandt why does windows suck so much :disappointed: 

Did you introduce make the original solution? then I will merge it...

cpockrandt:
@h-2 As discussed in person, this is the final version, so you're free to merge it :)

PR #1911, opened 2016-11-02
Fix/finder disallow non containerconcepts

### LINKED ISSUE #1683: Finder on an index sometimes allows to search Dna5 and sometimes not.
This code snippet works:

``` c++
Dna5String genome = "ACGTGCTAGCGATGACGATGCAGACTGACTGGACAGGCATGA";

Index<Dna5String, IndexEsa<> > index(genome);
Finder<Index<Dna5String, IndexEsa<> > > finder(index);
Dna5 needle(0);
find(finder, needle);
```

This doesn't:

``` c++
Dna5String genome = "ACGTGCTAGCGATGACGATGCAGACTGACTGGACAGGCATGA";

Index<Dna5String, FMIndex<> > index(genome);
Finder<Index<Dna5String, FMIndex<> > > finder(index);
Dna5 needle(0);
find(finder, needle);
```

I know that `needle` should be `Dna5String needle = "A"`. But I think it should be either allowed or disallowed in both cases.

cpockrandt:
Resolves #1683

### DIFF
--- a/tests/index/test_index_finder.cpp
+++ b/tests/index/test_index_finder.cpp
@@ -72,7 +72,7 @@ SEQAN_TYPED_TEST_CASE(IndexFinderTest, UnidirectionalIndexTypes);
 SEQAN_TYPED_TEST(IndexFinderTest, FindFirstChar)
 {
     SEQAN_ASSERT(find(this->finder, prefix(concat(this->text), 1u)));
-    SEQAN_ASSERT(find(this->finder, (const char*)"A"));
+    SEQAN_ASSERT(find(this->finder, "A"));
 }
 
 SEQAN_TYPED_TEST(IndexFinderTest, DefaultFinder)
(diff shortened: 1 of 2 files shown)


tests/index/test_index_finder.cpp
```diff
@@ -72,7 +72,7 @@ SEQAN_TYPED_TEST_CASE(IndexFinderTest, UnidirectionalIndexTypes);
 SEQAN_TYPED_TEST(IndexFinderTest, FindFirstChar)
 {
     SEQAN_ASSERT(find(this->finder, prefix(concat(this->text), 1u)));
-    SEQAN_ASSERT(find(this->finder, (const char*)"A"));
+    SEQAN_ASSERT(find(this->finder, "A"));
```
h-2: shouldn't an array of length one still be containerconcept? I though only `'A'` wouldn't be...

cpockrandt: I didn't notice the asterix :grin:

@h-2 Should I simply allow ``TPattern`` to be a pointer in ``find()``? Because `(const char*)` didn't implement the ContainerConcept.

h-2:
@cpockrandt ping, can you reply to the question?

Issue #1809, opened 2016-07-13
Improve/fix native packaging


h-2:
In https://github.com/seqan/seqan/blob/master/util/cmake/SeqAnBuildSystem.cmake#L528 add:
- correct `CPACK_DEBIAN_PACKAGE_ARCHITECTURE` and `CPACK_RPM_PACKAGE_ARCHITECTURE` depending on target platform
- other DEBIAN and RPM specfifc flags?
- CPACK_WIX instead of CPACK_NSIS for windows installers?

xenigmax:
I don't think there are many difficulties regarding this.
But I'm just curious. :)

(1) For deb/rpm (CPACK options, DEBIAN/REDHAT specific flags)
You mentioned that generation of deb/rpm/exe/dmg packages is just for our own apps.
So, do we need this? May I ask you why?

(2) For CPACK_WIX (windows)
May I ask what is the benefit of this change?

h-2:
> (1) For deb/rpm (CPACK options, DEBIAN/REDHAT specific flags)
> You mentioned that generation of deb/rpm/exe/dmg packages is just for our own apps.
> So, do we need this? May I ask you why?

We overwrite the system architecture from `amd64` to `x86_64` in https://github.com/seqan/seqan/blob/master/util/cmake/SetCPackSystemName.cmake . This breaks the packaging, because the .deb and .rpm generator don't "know" which architecture is used anymore. To make sure that the packages are correctly detected as "64bit amd/intel", we need to manually also set the flags mentioned above.

> (2) For CPACK_WIX (windows)
> May I ask what is the benefit of this change? 

This will generate an `.msi` file which I think is the recommend thing nowadays on Windows? I really am not an expert for windows, but heard they are better / integrate with the system settings or some such...

xenigmax:
@h-2, I will make a pull-request after merging #1900 .

I understand that architectural part. I will do that.  

However, for other DEB/RPM specific tags, would you let me know which tags do you want to add? 
I found several tags in here : https://cmake.org/Wiki/CMake:CPackPackageGenerators

.DEB, for example, these are already set.
CPACK_DEBIAN_PACKAGE_MAINTAINER
CPACK_PACKAGE_DESCRIPTION_SUMMARY
CPACK_DEBIAN_PACKAGE_VERSION

while these are not.
CPACK_PACKAGE_DESCRIPTION
CPACK_DEBIAN_PACKAGE_SECTION
[...]

PR #1915, opened 2016-11-02
[FIX] change CPACK_GENERATOR on windows & change maintainer in deb packaging

xenigmax:
This fix is based on @h-2  's suggestion (https://github.com/seqan/seqan/issues/1809).

I would like to ask you to consider two things before merging.

1. Dependency & license
Unlike CPACK_NSIS, CPACK_WIX depends on wixtoolset. (http://wixtoolset.org/releases/)
So a user has to install wixtoolset before using this.
wixtoolset uses Microsoft Reciprocal License (MS-RL). http://wixtoolset.org/about/license/
I'm not sure whether it's OK or not.

2. LICENSE file format
LICENSE file must be provided with ".txt" extension in UTF-8. Or .rtf must be provided. This is a mandatory information. https://cmake.org/cmake/help/v3.1/module/CPackWIX.html
All the apps that do not contain .txt(UTF-8) or .rtf can't ship their license to the package.

Regarding 2, we have to change LICENSE file to **UTF-8** with **.txt** extension. (LICENSE.txt)
But I don't know the consequences of this change in other systems.
> kjk@giraffe:~/work/seqan$ file -bi LICENSE 
> text/plain; charset=us-ascii

Or we can provide another file - LICENSE.rtf.

Let me know your preference. Then I will change them accordingly.


h-2:
> Dependency & license
> Unlike CPACK_NSIS, CPACK_WIX depends on wixtoolset. (http://wixtoolset.org/releases/)
> So a user has to install wixtoolset before using this.

Are you sure about this? I think it only needs to be available on the system that generates the installer, not on the target system. Can you check this?

> wixtoolset uses Microsoft Reciprocal License (MS-RL). http://wixtoolset.org/about/license/
> I'm not sure whether it's OK or not.

It's approved as a Free and Open Source license and it doesn't affect the license of our software so it's ok.

> Regarding 2, we have to change LICENSE file to UTF-8 with .txt extension. (LICENSE.txt)
> But I don't know the consequences of this change in other systems.

The encoding is irrelevant here. Regarding the extension, I don't want to change it for this little thing. Can you have CMake just copy `LICENSE`  to ` LICENSE.txt` in the build directory and then add that? [this would have to be done as part of seqan_configure_cpack_app]

xenigmax:
@h-2 .

About the installer I tested it in Klette and got a mesaage like "cannot find wix".
This was resolved after I installed the toolchian and added the path of binaries to the system variables.

I also tried to copy LICENSE to LICENSE.txt and it gave me errors and it couldn't produce the package.
According to the link ..
"If CPACK_RESOURCE_FILE_LICENSE has an .txt extension it is implicitly converted to RTF by the WiX Generator. The expected encoding of the .txt file is UTF-8."
So I thought that it should be utf-8.
May I try this with utf-8?

xenigmax:
I tried utf-8/.rtf but none of them worked.
The problem was not about encoding as you said.
It was about a path problem.  So It will ship license like before by copying them to .txt.
I will update PR about this.

But.. I have another issue. I realized a warning that I ignored before.
**EXEC : CPack warning : CPACK_WIX_UPGRADE_GUID implicitly set to 99EFC039-2326-44AB-BFDC-AA240BD79768 .** 
So it sounds like we have to set GUID for all apps if we don't want to see this.
Do you want to remove this? 
If so, how do you wish to set and maintain this?

xenigmax:
image

h-2:
> So it sounds like we have to set GUID for all apps if we don't want to see this.
> Do you want to set this?
> If so, how do you wish to set and maintain this?

I have no clue what you are talking about :sweat_smile: can you explain it?

xenigmax:
OK. Here is the screenshot. There is a warning message (yellow).
We should provide GUID to WIX or it generate them randomly with that warning message.

And.. Since GUID has to be unique by definition, we have to provide different GUIDs for all apps if we decide to provide them. That's what I'm asking.

Please feel free to let me know if I misunderstood something.


screenshot from 2016-11-04 14 19 42

xenigmax:
https://cmake.org/cmake/help/v3.0/module/CPackWIX.html
Windows uses it for checking pre-existing installation or replacing existing installations.
So it has to be maintained by us in theory..

h-2:
@xenigmax let's ignore the warning for now and take pre-generated GUIDs...

h-2:
@xenigmax are there open TODOs for this PR? What about the renaming of the license file?

h-2:
@xenigmax ping

rrahn:
@xenigmax ping

xenigmax:
@h-2 , @rrahn  I've updated it and tested it on Klette.

Issue #1289, opened 2015-11-18
deprecate Dependent<Tight> StringSets


h-2:
... and also undocument `valueById()` `removeValueById()` as no one knows exactly what they are good for, their interfaces don't align with SequenceConcept, there are currently no tests being run and no app uses them.
Also redo StringSets Tutorial.

esiragusa:
If you plan to keep these functions, you better mark them as deprecated in the dox! Or do you want to remove these functions?

h-2:
@esiragusa deprecating would probably be enough. But they need to be removed from the tutorial, too.

h-2:
"The Dependent-Tight StringSet is deprecated and will likely be removed within the SeqAn-2.x lifecycle."

  - [ ] add above  notice to the dox entry of `Spec Tight Dependent StringSet`
  - [ ] add notice to the `[set|get|remove][vV]alueById()` that they are deprecated and that one should use  version without `ById()` instead
  - [ ] add above notice the StringSets manual, just below `Specialization Dependent<Tight>`

h-2:
fixed by #1467

PR #1923, opened 2016-11-09
[DOX] Corrected and expanded the (dependent) string set documentation

smehringer:
- Removed the deprecation of the *ById() functions
- Added a detailed description about the behaviour of the Dependent StringSet

this PR is related to #1794


smehringer:
@h-2 I notice that this reverts a former issue #1289 


The documentation will now look like the following:

dox_dss

h-2:
Does this also include documentation for the different specializations? 

Also you wanted to change `[n+1]` to undefined behaviour and `getValueById(DELETED_ELEMENT)` to empty element...

include/seqan/sequence/string_set_dependent_tight.h
```diff
@@ -63,8 +63,30 @@ namespace seqan {
  * @tparam TString The type of the string to store in the string set.
  * @tparam TSpec   Tag for further specializing the string set.
  *
- * The class is not usable itself, only its subclasses @link TightDependentStringSet @endlink and
+ * Important: The class is not usable itself, only its subclasses @link TightDependentStringSet @endlink and
```
rrahn: Please rather write something: ```This is an abstract class. Use one of the following specializations of an dependent string set: ... .

smehringer:
@h-2 
not yet. I forgot to ask whether I should include the documentation for the different specializations on the same page (Dependent StringSet) or if I should document directly on each of the specialized sets?

I will add another pull request for the empty element issue since it's not related to the manual.

h-2:
>  I forgot to ask whether I should include the documentation for the different specializations on the same page (Dependent StringSet) or if I should document directly on each of the specialized sets?

Add it to the specialization pages, but please include them in this PR.

> I will add another pull request for the empty element issue since it's not related to the manual.

But the documentation should also reflect this.

smehringer:
> >I forgot to ask whether I should include the documentation for the different specializations on the same page (Dependent StringSet) or if I should document directly on each of the specialized sets?

> Add it to the specialization pages, but please include them in this PR.

I think my last commit (3e2e62d) did this?

> > I will add another pull request for the empty element issue since it's not related to the manual.

> But the documentation should also reflect this.

As we discussed last week, there will be no alteration of the behaviour of the different string set specializations (but I included in my last commit a documentation of how the string sets behave on semantic errors like empty element.)

h-2:
AFAICT the getValueById for Tight dss is in O(log(n)) and not amortized constant. Can you double-check these values please?

PR #1932, opened 2016-11-09
Fixed reading/writing of contig names in VCF header

esiragusa:
Note that parsing could be smarter: now it expects ID as first field. Contig lengths could also be parsed.

### DIFF
--- a/include/seqan/vcf_io/write_vcf.h
+++ b/include/seqan/vcf_io/write_vcf.h
@@ -61,6 +61,15 @@ writeHeader(TTarget & target,
         writeValue(target, '\n');
     }
 
+    for (unsigned i = 0; i < length(contigNames(context)); ++i)
+    {
+        write(target, "##");
+        write(target, "contig=<ID=");
+        write(target, contigNames(context)[i]);
+        writeValue(target, '>');
+        writeValue(target, '\n');
+    }
+
     write(target, "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT");
     for (unsigned i = 0; i < length(sampleNames(context)); ++i)
     {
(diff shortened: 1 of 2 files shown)


esiragusa:
Okay a couple of VCF test files must be updated as the header format changed.

Mason and Gustaf seem to have this functionality implemented on their own (!) as they now output twice contig header lines. @holtgrewe @ktrappe suggestions?

ktrappe:
I think that was the way you had to do it with the (old) I/O module. Is this only an internal seqan update or have the VCF specs changed somehow (I don't see a difference at first glance when I open the VCF spec 4.0).

include/seqan/vcf_io/write_vcf.h
```diff
@@ -61,6 +61,15 @@ writeHeader(TTarget & target,
         writeValue(target, '\n');
     }
 
+    for (unsigned i = 0; i < length(contigNames(context)); ++i)
```
rrahn: Why not using range based for loop here?

rrahn:
No, it just seems, regarding the submitted changes, that we never wrote out the contig names. 
Probably you have written your own method to do this in Gustaf/Mason?

include/seqan/vcf_io/write_vcf.h
rrahn: Really 😞

rrahn: Than just leave it.

ktrappe:
Yes, I have written my own method to write VCF headers and records:
https://github.com/seqan/seqan/blob/develop/apps/gustaf/msplazer_out.h#L971
and I do fill in the sequence names there too
https://github.com/seqan/seqan/blob/develop/apps/gustaf/msplazer_out.h#L1009

esiragusa:
Closing in favor of #1941.

Issue #1820, opened 2016-07-22
Remove `seqan.dddoc` and deprecated stuff under util/py_lib


kdmurray91:
Hi,

I'm packaging Seqan 2.x for the Debian and derivative operating systems. So far, so good, however it appears that there are many minified javascript and even Flash programs within the documentation helper code (specifically under [dddoc/tpl](https://github.com/seqan/seqan/tree/master/util/py_lib/seqan/dddoc/tpl)). These are not allowed (even in sources) within debian, as they are in essence not source code but compiled output.

My question is: are these used for anything user-facing? I notice the `dddoc` module has not been modified in some time, and there is a second `dox` helper. Is there any harm in excluding this module from the debian packages? Will the user-facing documentation still build?

Thanks,
Kevin (On behalf of [Debian Med](https://wiki.debian.org/DebianMed))

h-2:
Hi Kevin,

thanks for working on porting SeqAn2 first of all. As I mentioned previously to @mr-c of Debian Med in #1478 we do recommend strongly against using a full checkout of the repository for building the library packages. Having a huge meta-repository with all sorts of utils and apps is a legacy that we really rather not pass on to downstream.
Precisely for this purpose we offer "library source packages" on http://packages.seqan.de
These contain only the source of the library and corresponding html documentation and no python code or applications, they also have no build step and no external dependencies. Please see also our recommendations for downstream packaging here: http://seqan.readthedocs.io/en/master/Infrastructure/Manage/Releases.html#downstream-packaging

If you insist on building from the full repository checkout, you can remove all the dddoc stuff, as it is all legacy code, however, the dox stuff is required for generating the API documentation and some unit tests might use python, as well, I am not sure right now.

In any case we would be extremely happy to see a more recent version come to Debian and derivates.

Regards,
Hannes

rrahn:
@kdmurray91 @h-2 I think the dox still has some dependencies to the dddoc. So removing the subdirectory may cause some issues. But I am not so sure. It was some time ago I looked into it.

h-2:
@xenigmax since you have worked with the dox system recently I am reassigning this to you, I hope thats ok!

xenigmax:
I can't predict exact consequences after removing that parts, but I would say @rrahn is correct about this.
You will encounter problems on building docs which we can't cope with immediatly. 
I would like to recommand you to follow @h-2 's suggestion about this.

h-2:
[...]

PR #1936, opened 2016-11-10
[FIX,DOX] replace flash to html5

### LINKED ISSUE #1924: Replace flash using HTML5
Replace clipboard function based on flash using HTML5/javascript (https://clipboardjs.com/)

Refer : #1820

xenigmax:
This patch will replace flashes to html5.
This will close #1924 .
Also check #1820.

1
There are known compatibility issues between clipboard.js and bootsrap/jquery.
https://github.com/zenorocha/clipboard.js/wiki/Known-Issues
Since we are using bootstrap for modal dialog, I adopted the author's solution.
https://github.com/seqan/seqan/compare/develop...xenigmax:fix/1924?expand=1#diff-8ee174ca75ccedf003dd7ca9768032f0R10

2
I found that common.js uses tab and space for indentation together which is annoying.
I changed all tab to 4-space.


h-2:
Thanks! There seem to be some files that have big whitespace only changes. Can you revert them so the diff stays minimal? 
[I know the changes are improvements, but if something breaks later on reverting it would be difficult]

PR #1941, opened 2016-11-15
Fixed reading of contig names in VCF header

esiragusa:
Was PR #1932.

I removed contig names writing as it requires heavier changes. It would be cool if somebody implements that feature though.

Notes:
* VcfIOContext lacks contig lengths. I added code to parse that field as soon as VcfIOContext has contig lengths.
* VcfIOContext and BamIOContext would share enough functionality to have a common super context.
* Right now contig header line is stored as an unparsed header record. If we store contig names/lengths in the context and write those instead, other keys [like these](https://github.com/seqan/seqan/blob/master/tests/vcf_io/example.vcf#L5) will be lost.

### DIFF
--- a/include/seqan/vcf_io/read_vcf.h
+++ b/include/seqan/vcf_io/read_vcf.h
@@ -64,21 +64,43 @@ typedef Tag<Vcf_> Vcf;
 // Function readRecord()                                            [VcfHeader]
 // ----------------------------------------------------------------------------
 
-template <typename TString>
+template <typename TNameStore, typename TNameStoreCache, typename TStorageSpec, typename TString>
 inline void
-_parseVcfContigName(TString & contigName, TString const & headerValue)
+_readVcfContig(VcfIOContext<TNameStore, TNameStoreCache, TStorageSpec> & context, TString const & headerValue)
 {
     typedef OrFunctor<EqualsChar<','>, EqualsChar<'>'> >            IsCommaOrGt;
     typedef typename DirectionIterator<TString const, Input>::Type  TIter;
 
     TIter headerIter = directionIterator(headerValue, Input());
+    CharString &buffer = context.buffer;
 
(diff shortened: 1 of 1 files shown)


esiragusa:
@rrahn ping

include/seqan/vcf_io/read_vcf.h
```diff
@@ -64,21 +64,43 @@ typedef Tag<Vcf_> Vcf;
...
+        SEQAN_THROW(ParseError("Contig ID key not found in header."));
+    skipOne(headerIter);
+
+    // Read ID value.
+    clear(buffer);
+    readUntil(buffer, headerIter, IsCommaOrGt());
+    if (empty(buffer))
+        SEQAN_THROW(ParseError("Contig ID key is empty."));
+    appendName(contigNamesCache(context), buffer);
+
```
rrahn: Mhmm, I think it is not specified, whether ID key is the first or length key is the second element in the contig field, is it? So it could lead to an error, when the order of the tags is mixed. Or is it somewhere explicitly specified? I just could find for vcf4.3 that the ID tag is required. So maybe adapt it to parse all tags and then check for the corresponding key.

include/seqan/vcf_io/read_vcf.h
```diff
@@ -64,21 +64,43 @@ typedef Tag<Vcf_> Vcf;
...
+    // Seek length key.
```
rrahn: Could you add a note to remove comments, when this is reflected by the IOContext.

rrahn: @esiragusa ping

rrahn:
@esiragusa ping

esiragusa:
@rrahn pong

rrahn:
@esiragusa :ping_pong:

PR #1942, opened 2016-11-16
[FIX] Added length function for const dependent-generous-stringset.

### LINKED ISSUE #1939: No length() function for const string set dependent generous?
The `length()` function is overloaded by the dependent generous string set ([here](https://github.com/seqan/seqan/blob/master/include/seqan/sequence/string_set_dependent_generous.h#L185)) but only for **non const** string sets. Is this intentionally?

(same applies for `getValueById()`)

smehringer:
Fixes #1939

### DIFF
--- a/include/seqan/sequence/string_set_dependent_generous.h
+++ b/include/seqan/sequence/string_set_dependent_generous.h
@@ -187,6 +187,13 @@ length(StringSet<TString, Dependent<Generous> > & me)
     return length(me.limits) - 1;
 }
 
+template <typename TString >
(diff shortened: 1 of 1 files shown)


include/seqan/sequence/string_set_dependent_generous.h
```diff
@@ -187,6 +187,13 @@ length(StringSet<TString, Dependent<Generous> > & me)
     return length(me.limits) - 1;
 }
 
+template <typename TString >
```
rrahn: please remove the last space in the ```<>```

PR #1947, opened 2016-11-23
[FEATURE] Added Argument types: bool, input_directory and output_directory

### LINKED ISSUE #1946: Some additions to arg parser
- [x] add boolean arguments that can be set with `{on,1,true,yes}` or `{off,0,false,no}`
  - [ ] add INPUT_DIRECTORY and OUTPUT_DIRECTORY which are the same as _FILE but with different label

smehringer:
Fixes #1946


smehringer:
@h-2  only the last two commits are relevant (and appear in the diff). I don't know why the history of commits before are shown..

take a first look. I'll write tests on friday.

h-2:
> I don't know why the history of commits before are shown.. 

Maybe you did not update your remote before creating the branch? You can easily fix this by updating the remote now and then doing a git rebase on upstream/develop

> take a first look. I'll write tests on friday.

Looks good :)

h-2:
For the bool type does it show the valid values on the command line?

smehringer:
Yes. It looks like the following:
```
    -b, --mybool BOOL
          turn me on. One of 0, OFF, FALSE, F, 1, ON, TRUE, and T.
```
I cannot (or not easily through the interface) put everything to upper or lower case before checking the user input. That means the user actually has to put in the specific term OFF correctly. But I think it is not much to ask ..
or do you want me to find a work around to still provide this functionality?

smehringer:
I did not include tests for 
- checking if value of boolean opetion is on/off/.. because this is covered by the test for validValues
- checking for file extension behaviour for the type input/ouputdirectory since this is covered for normal files and the dorectory behaves as such.

h-2:
I am not yet sure how this actually works from the code, i.e. where is the "equivalence class" of `{0, FALSE, OFF}` actually defined? Can I call `getOptionValue(someBoolType, parser, "mybool")` and it will work? How?

> or do you want me to find a work around to still provide this functionality?

Hm, I am not 100% sure, but I think it would be good. Ideally we would internally support more combinations than are actually shown on the command line...

rrahn:
@h-2 What is the motivation for a bool type. I thought we have a simple tag version of an Option, which is either set or not. In this case this works for all equivalence classes of ```false```, ```0``` and so on and there is no misunderstanding.

h-2:
@rrahn In pipelines and scripts it is common to change options via environment variables. This can only be facilitated via bool types, not via existance/non-existance. Take for example:
```sh
./my_program --second-realign ${REALIGN} --other-option 3 INPUT.file
```
Of course, you could instead replace `--second-realign` with the variable itself, but it would make it more confusing. What if the default is `on` and the non-default would be `off`? What if this changes from one version to the next?

rrahn:
If it changes between versions you have to adapt your script anyway, so that's no argument 😏 
But ok, it would at least be a shortcut for an own STRING-Option which defines certain on/off values.

smehringer:
Replaced in favour pf #1956

PR #1951, opened 2016-11-24
Fix travis CI

rrahn:
- Fixes issue with upgrading pip
- Uses apt install for clang++-3.5 and clang++-3.8
- Adds gcc6 to build matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,63 +23,29 @@ matrix:
       install: export CXX="g++-5" CC="gcc-5"
 
     # package not whitelisted currently
-    #- os: linux
-      #compiler: gcc-6
-      #addons:
-        #apt:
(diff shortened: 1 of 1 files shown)


h-2:
you still exceed the time limit, I think you shouldnt add the line about "precise"...

rrahn:
@h-2 quite possible. Let's see if it is fixed now.

rrahn:
@h-2 great now it is using container based infrastructure but still caps the build at 50 min. Any idea?

h-2:
@rrahn unfortunately not
reassigning, since I will be on vacation

h-2:
Maybe the timeout is related to what is currently setup as the default, i.e. a PR update of the yml won't change this behaviour. I will merge this to verify this. If it doesn't help I think you need to contact travis support.

PR #1956, opened 2016-12-05
[FEATURE] Added Argument types: bool, input_directory and output_directory

smehringer:
Replaces #1947

### DIFF
--- a/include/seqan/arg_parse/arg_parse_type_support.h
+++ b/include/seqan/arg_parse/arg_parse_type_support.h
@@ -44,6 +44,28 @@
 
 namespace seqan {
 
+
+// ==========================================================================
+// Tags, Classes, Enums
+// ==========================================================================
+
+template <typename TVoidSpec = void>
+struct BooleanArgumentValues_
+{
+    static const std::vector<std::string> LIST_TRUE;
+    static const std::vector<std::string> LIST_FALSE;
+};
+
+template <typename TVoidSpec>
+const std::vector<std::string> BooleanArgumentValues_<TVoidSpec>::LIST_TRUE{"1", "ON", "TRUE", "T"};
+
+template <typename TVoidSpec>
+const std::vector<std::string> BooleanArgumentValues_<TVoidSpec>::LIST_FALSE{"0", "OFF", "FALSE", "F"};
+
+// ==========================================================================
+// Functions
+// ==========================================================================
+
 // ----------------------------------------------------------------------------
 // Function _tryCast()
 // ----------------------------------------------------------------------------
@@ -68,6 +90,23 @@ inline TTarget _cast(TString const s)
     bool result = (!(stream >> dst).fail()) && (stream.rdbuf()->in_avail() == 0);
     SEQAN_CHECK(result, "could not cast %s", toCString(s));
     return dst;
+}
+
+// ----------------------------------------------------------------------------
+// Function _castBool()
+// ----------------------------------------------------------------------------
+
+template <typename TString>
+inline bool _castBool(bool & dst, TString const s)
+{
+    // since the validValue check already verifies that the value if one of
+    // BooleanArgumentValues_<>::LIST, one only needs to check one equivalent class
+    // dst = std::find()
+    dst = (std::find(BooleanArgumentValues_<>::LIST_TRUE.begin(),
+                     BooleanArgumentValues_<>::LIST_TRUE.end(),
+                     s)
+           != BooleanArgumentValues_<>::LIST_TRUE.end());
+    return true;
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


include/seqan/arg_parse/arg_parse_type_support.h
```diff
@@ -44,6 +44,28 @@
...
+
+template <typename TVoidSpec = void>
+struct BooleanArgumentValues_
+{
+    static const std::vector<std::string> LIST_TRUE;
+    static const std::vector<std::string> LIST_FALSE;
+};
+
+template <typename TVoidSpec>
+const std::vector<std::string> BooleanArgumentValues_<TVoidSpec>::LIST_TRUE{"1", "ON", "TRUE", "T"};
```
h-2: please add "YES"

include/seqan/arg_parse/arg_parse_type_support.h
```diff
@@ -44,6 +44,28 @@
...
+
+template <typename TVoidSpec>
+const std::vector<std::string> BooleanArgumentValues_<TVoidSpec>::LIST_FALSE{"0", "OFF", "FALSE", "F"};
```
h-2: please add "NO"

include/seqan/arg_parse/arg_parse_type_support.h
```diff
@@ -71,6 +93,23 @@ inline TTarget _cast(TString const s)
 }
 
 // ----------------------------------------------------------------------------
+// Function _castBool()
+// ----------------------------------------------------------------------------
+
+template <typename TString>
+inline bool _castBool(bool & dst, TString const s)
+{
+    // since the validValue check already verifies that the value if one of
```
h-2: make a copy of TString and std::transform it to uppercase so we have the lower and mixed cases, as well!

include/seqan/arg_parse/arg_parse_type_support.h
```diff
@@ -71,6 +93,23 @@ inline TTarget _cast(TString const s)
...
+    // BooleanArgumentValues_<>::LIST, one only needs to check one equivalent class
+    // dst = std::find()
+    dst = (std::find(BooleanArgumentValues_<>::LIST_TRUE.begin(),
+                     BooleanArgumentValues_<>::LIST_TRUE.end(),
+                     s)
+           != BooleanArgumentValues_<>::LIST_TRUE.end());
+    return true;
```
h-2: I would prefer, if we check both the negative and the positive value and return false if neither match. It forces people to use the interface right and makes detecting errors easier.

include/seqan/arg_parse/arg_parse_type_support.h
```diff
@@ -126,6 +165,15 @@ inline bool isIntegerArgument(ArgParseArgument const & me);
...
 inline bool isDoubleArgument(ArgParseArgument const & me);
 inline bool isStringArgument(ArgParseArgument const & me);
+inline bool isBooleanArgument(ArgParseArgument const & me);
+
+inline bool _convertArgumentValue(bool & dst, ArgParseArgument const & opt, std::string const & src)
+{
+    if (!isBooleanArgument(opt))
+        return false;
+
+    return _castBool(dst, src);
```
h-2: why is it called _castBool and not _tryCast like the others?

include/seqan/arg_parse/arg_parse_type_support.h
smehringer: This is already done by the validValues functionality: Since the list of Yes, No, etc is given as validValues, the argument parser checks the user input first.

include/seqan/arg_parse/arg_parse_type_support.h
smehringer: There is no possibility do to this here. Every user input is checked automatically if it's value can be found in the (non empty) `option._validValues` list (look [here](https://github.com/seqan/seqan/blob/master/include/seqan/arg_parse/arg_parse_argument.h#L764) ). One would have to add a special case only for boolean values which I think is not consistent with the general use of the argument parser.

If the user input is not inside the valid values the argument parser will indicate such.
```
myapp: the given value 'off' is not in the list of allowed values [1, ON, TRUE, T, YES, 0, OFF, FALSE, F, NO]
```

include/seqan/arg_parse/arg_parse_type_support.h
smehringer: There if only [one generic _tryCast() function](https://github.com/seqan/seqan/blob/master/include/seqan/arg_parse/arg_parse_type_support.h#L51) that handles every cast so far. Since the boolean functionality does not actually cast the string directly to a bool, I thought to keep the difference in this behaviour by a different name instead of overloading the function _tryCast() for boolean type.
Maybe the name _tryCastFromList or _tryCastFromEquivalenceClass or some other would fit better?

include/seqan/arg_parse/arg_parse_type_support.h
h-2: > One would have to add a special case only for boolean values which I think is not consistent with the general use of the argument parser.

Well at the other place it also says in the code:
```c+++
// we only check valid values for files and string arguments, but not for prefix arguments
```
So it's the perfect place to introduce a transform for bools and changing the comment or not? :laughing:

include/seqan/arg_parse/arg_parse_type_support.h
h-2: Hm, I think you could just call it `_tryCast` and specialize for bool, but the name is not so important.

smehringer: ok

include/seqan/arg_parse/arg_parse_type_support.h
h-2: @smehringer this is the only outstanding change, right?

smehringer: yes, I'll work on that today

PR #1958, opened 2016-12-08
[FEATURE] Added macros for disabling, opt-in or opt-out version update check

rrahn:
Introduces 2 new macros/definition flags to select the behavior for the version update checker:

- ```SEQAN_VERSION_CHECK_OPT_IN```: enables version check but leaves it as opt-in option.
- ```SEQAN_DISABLE_VERSION_CHECK```: disables version check completely.

If non is specified the default will be enabled and fully activated.

Revises the cmake option ```SEQAN_VERSION_CHECK = disable|opt-in|opt-out```
```opt-out``` is the default.

The update version is disabled for all tests and demos but ```test_arg_parse_version_check``` which is explicitly enabled locally in the cpp.
I tested all different behaviors and adapted the tests accordingly.
@smehringer can you verify, that the test behavior matches the expected one?

### DIFF
--- a/manual/source/Infrastructure/Use/CustomBuildSystem.rst
+++ b/manual/source/Infrastructure/Use/CustomBuildSystem.rst
@@ -192,17 +192,24 @@ meaning
  If set to 1 then OpenMP is expected to be available.
  You might have to add ``-fopenmp`` and possibly ``-lgomp`` to your build. And OpenMP needs to be supported by your compiler.
 
-SEQAN_DISABLE_VERSION_VCHECK
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+SEQAN_VERSION_CHECK_OPT_IN
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-possible value
-  0, 1
+meaning 
+ If set then the version update feature is enabled but deactivated in the argument parser options and must be explicitly activated by the user.
+ If not specified the update feature is enabled and activated by default.
 
-default
-  0 (from release 2.4.0 onwards)
+usage
+ Add compiler flag: ``-DSEQAN_VERSION_CHCK_OPT_IN``
+
+SEQAN_DISABLE_VERSION_CHECK
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 meaning
-  Starting from the release 2.4.0 every application or script that uses the SeqAn argument parser (and only those) will have a new feature to check whether your SeqAn version is outdated. This feature can be disabled by setting this flag to 1. Note: You can also disble this feature at runtime by specifying the command line option ``--version-check OFF``.
+ If set the version update feature is disabled in the argument parser.
+
+usage
+ add compiler flag: ``-DSEQAN_DISABLE_VERSION_CHEK`` 
 
 Settings Projects Using Seqan
 -----------------------------
--- a/manual/source/Infrastructure/Use/CustomBuildSystem.rst
+++ b/manual/source/Infrastructure/Use/CustomBuildSystem.rst
@@ -200,7 +200,7 @@ meaning
  If not specified the update feature is enabled and activated by default.
 
 usage
- Add compiler flag: ``-DSEQAN_VERSION_CHCK_OPT_IN``
+ Add compiler flag: ``-DSEQAN_VERSION_CHECK_OPT_IN``
 
 SEQAN_DISABLE_VERSION_CHECK
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: manual/source/Infrastructure/Use/CustomBuildSystem.rst [hunk 1]
```diff
@@ -192,17 +192,24 @@ meaning
...
+
+SEQAN_DISABLE_VERSION_CHECK
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 meaning
-  Starting from the release 2.4.0 every application or script that uses the SeqAn argument parser (and only those) will have a new feature to check whether your SeqAn version is outdated. This feature can be disabled by setting this flag to 1. Note: You can also disble this feature at runtime by specifying the command line option ``--version-check OFF``.
+ If set the version update feature is disabled in the argument parser.
+
+usage
+ add compiler flag: ``-DSEQAN_DISABLE_VERSION_CHEK``
```
h-2: Here two variables are described (at c-macro-level)

manual/source/Infrastructure/Use/FindSeqAnCMake.rst
```diff
@@ -145,11 +145,21 @@ See :ref:`this page <infra-use-cmake-build-dirs>` for more details.
...
+  ===================================  ============================================
+  ``-DSEQAN_VERSION_CHECK="opt-out"``  The update feature is enabled and switched
+                                       on in the argument parser options. (default)
 
-  #. You can disable the feature locally for every command line application by specifying ``--version-check OFF``
+  ``-DSEQAN_VERSION_CHECK="opt-in"``   The update feature is enabled but switched
+                                       off in the argument parser options.
+
+  ``-DSEQAN_VERSION_CHECK="disable"``  The update feature is disabled.
+  ===================================  ============================================
```
h-2: here a single three-variate variable is described (at c-macro-level)

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -248,8 +248,10 @@ macro (seqan_build_system_init)
     ## options
 
     # SeqAn Version Check
-    if (NOT SEQAN_VERSION_CHECK)
-        set (SEQAN_DEFINITIONS "${SEQAN_DEFINITIONS};-DSEQAN_DISABLE_VERSION_CHECK")
+    if ("${SEQAN_VERSION_CHECK}" STREQUAL "disable")  # Disable completely
+        set (SEQAN_DEFINITIONS ${SEQAN_DEFINITIONS} -DSEQAN_DISABLE_VERSION_CHECK)
+    elseif ("${SEQAN_VERSION_CHECK}" STREQUAL "opt-in")  # Build it but make it opt-in
+        set (SEQAN_DEFINITIONS ${SEQAN_DEFINITIONS} -DSEQAN_VERSION_CHECK_OPT_IN)
     endif ()
```
h-2: Here there seems to be a single three-variate variable at cmake level, but then two variables at c-macro-level

h-2:
I am not really sure if we need three states, in which case is "disabled" preferable to "opt-in"?

Another possibility for the release would be to just turn it on manually for our binaries, but not for the source package so that we get reports for our apps, but down-stream developers are not yet exposed to the feature. But I don't know if that is sufficient, @rrahn ?

rrahn:
Well, the disabling part is actually for external app developers who doesn't want to offer this kind of feature at all. Second, I am not sure if there are some package managers who rather keep that out. Of course, opt-in would be sufficient for me.

manual/source/Infrastructure/Use/CustomBuildSystem.rst
rrahn: Well, yes, I wanted to give the default a name. But I agree I could rather use two option variables.

h-2:
travis app failure, is this related?

rrahn:
Mhmm it's timeout of two apps. Maybe related to the version check, but only gcc-4.9

h-2:
@rrahn I have restarted that individual build, but if it stays broken we need to investigate. Might be related to C++11 threading and waiting in the version check.

rrahn:
Well, but it seems to be those two apps all the time. All others work fine. I can disable the feature and see what happens then. I'll try to compile with gcc 4.9 on docker images see if I can reproduce it

rrahn:
@smehringer I could not figure out, why he is not writing the seqan.version. Can we check if this is correctly setup on the server?

smehringer:
@rrahn Even when you wait for the future object in the destructor of the argument parser?

h-2:
@rrahn  @rrahn  the builds are now green. Does this PR still contain TODOs or can it be merged? [its good from my POV]

rrahn:
Ok, it seems all ok then.

PR #1960, opened 2016-12-08
Fix two outstanding cmake issues

h-2:


h-2:
@temehi can you double check on a linux box that this creates a correct static binaries on linux? Thx!

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -311,6 +282,38 @@ macro (seqan_build_system_init)
...
+            set(CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS)
+            # make sure -fPIC isn't added automatically
+            set(CMAKE_SHARED_LIBRARY_CXX_FLAGS)
+
+            # For unknown reasons finding .a only seems to work for libz and
+            # libbzip2; cmake than proceeds to wrap these in
+            # -Wl,-Bstatic -lz -lbz2 -Wl,-Bdynamic
+            # the latter reactivates dynamic linking for the system libs
+            # we override this behaviour here:
+            set(CMAKE_EXE_LINK_DYNAMIC_CXX_FLAGS)
```
rrahn: Your PR is quite difficult to read, since only the ```if not develop ``` expression is removed? Did you change some indentation or something?
So my question: was this query just obsolete, because SEQAN_STATIC_APPS is not set in develop mode?
But which behavior changes than?

h-2: The main change is that the entire static block has been **moved** from the register_apps() to the init_build_system() call.
Regarding `CMAKE_EXE_LINK_DYNAMIC_CXX_FLAGS` the previous comment already illustrated that I am not really sure why these things are happening. This behaviour now also changed to where it is always necessary to overwrite this flag. So it is still weird, but at least it is now consistently weird. That's why the comment changed.

As the previous comment

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -311,6 +282,38 @@ macro (seqan_build_system_init)
...
+        endif (APPLE)
+    endif (SEQAN_STATIC_APPS AND (NOT CMAKE_SYSTEM_NAME MATCHES "Windows"))
+
+    # strip binaries when packaging
+    if ((CMAKE_BUILD_TYPE STREQUAL "Release") AND
+        (NOT SEQAN_BUILD_SYSTEM STREQUAL "DEVELOP") AND
+        (COMPILER_CLANG OR COMPILER_GCC))
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s")
```
h-2: This is entirely new and solves a different issues, as can be seen from the individual commits.

rrahn: Sure

util/cmake/SeqAnBuildSystem.cmake
rrahn: Ok, I missed the move to another location. but seems to be ok-ish

PR #1824, opened 2016-07-28
Snp_store

serosko:
Änderungen am SnpStore im Verlauf meiner Masterarbeit.

PR #1961, opened 2016-12-09
Improvement of the snp_store

rrahn:
This is a mirror of #1824 which hat some unresolved merge conflicts.


apps/snp_store/snp_store.cpp
```diff
@@ -703,82 +734,161 @@ int detectSNPs(SNPCallingOptions<TSpec> &options)
...
             }
         }
+    }
     // Allocate one BamAlignmentRecord for each input BAM file.
     std::vector<BamAlignmentRecord> records(length(options.readFNames));
 
     /////////////////////////////////////////////////////////////////////
+    std::time_t t = std::time(nullptr);                                         //Get time and date for vcf header
+    char now[24];
+    strftime(now, sizeof(now), "%Y%m%d", std::localtime(&t));
```
rrahn: Do you really write the timestamp of that file into the vcf file?
Is that common praxis and really necessary?
We can make tests with timestamps. At least not with the current system.
So you need make the timestamp writing optional (using a macro: like ```SEQAN_SNP_STORE_DISABLE_TIME_STAMP```) and let your cmake script add this, when ```SEQAN_BUILD_SYSTEM``` equals ```DEVELOP```, like:
```shell
if ("${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
   message (STATUS "Disabling time stamp feature for writing vcf files!")
   set (SEQAN_DEFINITIONS ${SEQAN_DEFINITIONS} -DSEQAN_SNP_STORE_DISABLE_TIME_STAMP)
endif ()
```
Thus you are able to fix the app tests.

apps/snp_store/snp_store.cpp
```diff
@@ -798,94 +908,160 @@ int detectSNPs(SNPCallingOptions<TSpec> &options)
...
+//                                     tmpReads,
+//                                     tmpReadCounts,
+//                                     tmpReadCounts,
+//                                     tmpReadClips,
+//                                     tmpReadCigars)
+//                             firstprivate()
+//                             shared(options, sumWindows, i)
+//                             reduction(+:sumwindows)
+//                             schedule(dynamic, chunk)
+//                             ordered)
```
rrahn: can this be removed

apps/snp_store/snp_store.cpp
```diff
@@ -798,94 +908,160 @@ int detectSNPs(SNPCallingOptions<TSpec> &options)
...
+        for (currentWindowBegin = 0;
+             currentWindowBegin < (TContigPos)length(genomes[i]);
+             currentWindowBegin += options.windowSize)
         {
             TContigPos currentWindowEnd = currentWindowBegin + options.windowSize;
-            if(currentWindowEnd > (TContigPos)length(genomes[i])) currentWindowEnd = (TContigPos)length(genomes[i]);
-
-            if(positionStatsOnly && !(currentWindowBegin <= *inspectPosIt && *inspectPosIt< currentWindowEnd))
+            if (currentWindowEnd > (TContigPos)length(genomes[i]))
+                currentWindowEnd = (TContigPos)length(genomes[i]);
```
rrahn: add surrounding ```{}```

apps/snp_store/snp_store.cpp
```diff
@@ -938,103 +1128,177 @@ int detectSNPs(SNPCallingOptions<TSpec> &options)
...
+                //                if (*options.outputCNV != 0)
+                //                {
+                //                    dumpCopyNumberPolymorphismsBatch(fragmentStore,
+                //                                                     genomeNames[i],
+                //                                                     startCoord,
+                //                                                     currentWindowBegin,
+                //                                                     currentWindowEnd,
+                //                                                     cnvFileStream,
+                //                                                     options);
+                //                }
```
rrahn: can be removed? Otherwise document what it is for and when to add it.

serosko:
I should manage to correct the issues today.

apps/snp_store/snp_store.cpp
```diff
@@ -81,39 +96,42 @@ bool loadGenomes(TGenomeSet &genomes, StringSet<CharString> &fileNameList, ::std
...
         for (; !atEnd(seqFileIn); ++seqCount)
         {
-            readRecord(temp, seq, seqFileIn);
+            try
+            {
+                readRecord(temp, seq, seqFileIn);
+            }
+            catch (ParseError const & e)
+            {
+                //std::cout << e.what() << "Base will be considered as 'N'." << std::endl;
```
serosko: If I remember correctly, the standard output was flooded with error messages I used SNPStore on a FASTA file containing not only A,C,G,T,N.  I fixed it now, s.t. the warning will only be written to ouput when it occurs for the first time.

apps/snp_store/snp_store.cpp
serosko: I cannot compile or run the code where I am working from at the moment, hence I would almost certainly introduce some errors if i tried to switch to chrono library now.
As for the time stamp: Yes it is common to write it to the header of the vcf file (see https://samtools.github.io/hts-specs/VCFv4.2.pdf). I cannot change this now for the same reason as mentioned above.

apps/snp_store/snp_store.cpp
serosko: As it seems there endeavours to add copy-number-variation detection or something similar to SNP-Store. I never touched these parts of the code myself, but all of them seem to be commented out. My guess woul be that the might come in handy, if someone wants to add this functionality in the future.

serosko:
I applied the suggested changes, and made a new pull request after merging everything with upstream/develop. Because I cannot compile or run the code where I am working from at the moment, I did not switch to chrono library and also did not change the behaviour of the time stamp. Consequently no newly generated test files were added.

apps/snp_store/snp_store.cpp
rrahn: @serosko any chance you can work on that before Monday?

rrahn:
@serosko do you think you can do the last changes until Monday?

serosko:
Sorry, I didn't check my emails during the weekend. Also I won't be able to make any changes until at least thursday.
________________________________
Von: rrahn [notifications@github.com]
Gesendet: Freitag, 9. Dezember 2016 20:00
An: seqan/seqan
Cc: Roskosch, Sebastian; Mention
Betreff: Re: [seqan/seqan] Improvement of the snp_store (#1961)


@serosko<https://github.com/serosko> do you think you can do the last changes until Monday?

rrahn:
closed in favor of #1962

PR #1963, opened 2016-12-09
[FEATURE] Rename FindSeqAn to SeqAnConfig.

rrahn:
Renames FindSeqAn to SeqAnConfig as actually wanted by cmake.
This increases the chance that SeqAn can be used as external project by simply calling 
```find_package (SeqAn)```, since cmake looks in many system wide locations for the config file not for a module file.
Solves issues with macports packaging.
@all If someone maintains a seqan package for some other system, please be aware of the name change.

### CODE REVIEW

File: manual/source/Infrastructure/Use/FindSeqAnCMake.rst [hunk 1]
```diff
@@ -155,8 +156,8 @@ Starting from the release 2.4.0 the argument parser will have a new feature to c
 
     This does only affect applications or scipts that use the SeqAn :ref:`Argument Parser <tutorial-getting-started-parsing-command-line-arguments>`!
 
-Details of the FindSeqAn Module
--------------------------------
+Details of the SeqAnConfig Module
```
h-2: just call it SeqAn Module, maybe?

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -373,8 +373,8 @@ macro (seqan_setup_library)
             configure_file("util/pkgconfig/seqan.pc.in" "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION_MAJOR}.pc" @ONLY)
             install(FILES "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION_MAJOR}.pc" DESTINATION share/pkgconfig)
         endif (NOT CMAKE_SYSTEM_NAME MATCHES Windows)
-        # Install FindSeqAn TODO(h-2) rename FindSeqAn.cmake to FindSeqAn${SEQAN_VERSION_MAJOR}.cmake after 2.x cycle
-        install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/util/cmake/FindSeqAn.cmake" DESTINATION share/cmake/Modules)
+        # Install FindSeqAn TODO(h-2) rename SeqAnConfig.cmake to SeqAnConfig${SEQAN_VERSION_MAJOR}.cmake after 2.x cycle
+        install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/util/cmake/SeqAnConfig.cmake" DESTINATION share/cmake/)
```
h-2: If we use the Config name then we shouldn't install to `$PREFIX/share/cmake`, but to `$PREFIX/share/seqan`, or not?

manual/source/Infrastructure/Use/FindSeqAnCMake.rst
rrahn: ok

util/cmake/SeqAnBuildSystem.cmake
rrahn: Well I will rename it to ```seqan-config.cmake ``` and it should live probably in ```${prefix}/share/cmake/seqan```. Actually both would be sufficient, but it is likely that there is a global cmake folder with all the projects.

rrahn:
I tested it on my computer and later on travis. Currently looking into #1958 issues.

rrahn:
I've tried as well with an external project.

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -321,7 +322,7 @@ macro (seqan_build_system_init)
         find_package(ZLIB)
         find_package(BZip2)
         find_package(Boost)
-        find_package(SeqAn REQUIRED)
+        find_package(SeqAn CONFIG REQUIRED)
```
h-2: Why is this required? And is it also required for the apps when built individually, i.e. did you try `SEQAN_BUILD_SYSTEM=SEQAN_RELEASE_APPS` ?

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -373,8 +374,8 @@ macro (seqan_setup_library)
             configure_file("util/pkgconfig/seqan.pc.in" "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION_MAJOR}.pc" @ONLY)
             install(FILES "${CMAKE_BINARY_DIR}/util/pkgconfig/seqan-${SEQAN_VERSION_MAJOR}.pc" DESTINATION share/pkgconfig)
         endif (NOT CMAKE_SYSTEM_NAME MATCHES Windows)
-        # Install FindSeqAn TODO(h-2) rename FindSeqAn.cmake to FindSeqAn${SEQAN_VERSION_MAJOR}.cmake after 2.x cycle
-        install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/util/cmake/FindSeqAn.cmake" DESTINATION share/cmake/Modules)
+        # Install FindSeqAn TODO(h-2) rename seqan-config.cmake to seqan-config${SEQAN_VERSION_MAJOR}.cmake after 2.x cycle
+        install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/util/cmake/seqan-config.cmake" DESTINATION share/cmake/)
```
h-2: you didn't address my previous question regarding this line: IMHO it should be `share/seqan`, not `share/cmake`.

h-2: I just read up on it:
https://cmake.org/cmake/help/v3.0/command/find_package.html

Both work, but I would strongly suggest `share/seqan/` as we install other stuff to that directory, too, and packaging systems explicitly wanted us not to write to `share/cmake/`

rrahn: Why not into ```share/cmake```? 
It should actually go into ```share/cmake/seqan/``` I think. 
Also the .pc file is installed into ```share/pkgconfig``` and not ```share/seqan/```
AFAIK, we install the docs into ```share/docs/``` too. So why changing the pattern for cmake?

util/cmake/SeqAnBuildSystem.cmake
rrahn: I tested several different things. Might be lost in action.

rrahn:
I know that SeqAnConfig is the same as seqan-config, but I generally prefer the all lower-case variant.

rrahn:
@h-2 I tried to package apps, and the library. Everything seems to work for me.

h-2:
Ok, well concerning the paths multiple things seem to be possible. This is the style Debian uses:
```
/usr/share/llvm-3.5/cmake/LLVMConfig.cmake
```
i.e. $PREFIX/share/$name/cmake/$NameConfig.cmake

This is the style FreeBSD uses:
```
/usr/local/lib/cmake/libxml2/libxml2-config.cmake
```
i.e. $PREFIX/lib/cmake/$name/$name-config.cmake

MacPorts uses same as FreeBSD, but also without cmake:
```
/opt/local/lib/cmake/libxml2/libxml2-config.cmake
/opt/local/lib/isl/isl-config.cmake
```

You propose `$PREFIX/share/cmake/$name/$name-config.cmake`. I would propose either

  * `$PREFIX/lib/cmake/$name/$name-config.cmake` (lib instead of share, mac+bsd style); or
  * `$PREFIX/share/$name/$name-config.cmake` (because it's simplest)

The second thing would also make sense if we install our license and readme there -- which I think would be good, because right now they are installed with the documentation which is a little strange. In any case all the directories should work and downstream packagers can move it to wherever they want anyway!

h-2:
[but this is not really important, feel free to merge as you see fit]

rrahn:
@h-2 Well I would then propose to put both cmake and pkgconfig below ```$PREFIX/lib/```

PR #1966, opened 2016-12-12
[FIX] Fix windows error with static const initialization with brace-initialization.

rrahn:
Uses constexpr with std::array to get rid of windows error/bug?
Seems to be a bug with brace-initilaization in static const expression?

### DIFF
--- a/include/seqan/arg_parse/arg_parse_argument.h
+++ b/include/seqan/arg_parse/arg_parse_argument.h
@@ -197,10 +197,10 @@ public:
     {
         if (argumentType == ArgParseArgument::BOOL)
         {
-            validValues = BooleanArgumentValues_<>::LIST_TRUE;
-            validValues.insert(validValues.end(),
-                               BooleanArgumentValues_<>::LIST_FALSE.begin(),
-                               BooleanArgumentValues_<>::LIST_FALSE.end());
+            copy(BooleanArgumentValues_<>::LIST_TRUE.begin(), BooleanArgumentValues_<>::LIST_TRUE.end(),
+                 std::back_inserter(validValues));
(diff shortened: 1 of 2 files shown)


h-2:
I had a similar issue before and I think it can be solved with less changes by wrapping the entire clause in two pairs of `{}` and without the assignment operator, See this example:
https://github.com/seqan/seqan/blob/master/include/seqan/blast/blast_tabular.h#L399

rrahn:
@h-2 I tried with the current develop version with two braces, cause I thought the same. Did not work for me!

rrahn:
Is there anything that favors  ```Type val {}``` over ```Type val = {}``` in the member initialization?

h-2:
> Is there anything that favors Type val {} over Type val = {} in the member initialization?

Direct initialization and copy initialization need not be identical, the latter allows for (chains of) user defined conversions and implicit casts, also for class types, the former allows this only for non-class types.

In the CPP session we agreed that we prefer `Type val{arg1};` as the generic way, because it works for everything, i.e. PODs, aggregates and classes with custom constructors. Copy initialization we only wanted to use for PODs and iterators IIRC -- or if it is necessary to allow for abovementioned conversions.

> I tried with the current develop version with two braces, cause I thought the same. Did not work for me!

Hm, that sounds strange. Did you include the out-of-class declerators:
https://github.com/seqan/seqan/blob/master/include/seqan/blast/blast_tabular.h#L639

smehringer:
Yes he did
https://github.com/seqan/seqan/pull/1966/files#diff-0ee3e8fe683494862e7022fde6feea36R60

rrahn:
@h-2 last say?

PR #1916, opened 2016-11-02
[FIX] Find Python 2.x and use it.

xenigmax:
This will close #1482.
If python is set to python 3.x, cmake will find 2.x and use it.

PR #1969, opened 2016-12-12
[FIX] Change python path in .sh files 

### LINKED ISSUE #1482: Python scripts incompatible with Python 3
Python scripts seem not to be compatible with Python 3. On some computers, `#!/usr/bin/env python` returns Python 3. It happens, e.g., with Anaconda which sometimes prepends its directory to `PATH`.

``` shell
Scanning dependencies of target dox
Traceback (most recent call last):
  File "/home/karel/github/cmake-ext/tests/cmake-ext/seqan-prefix/src/seqan/dox/../util/bin/dox.py", line 14, in <module>
    sys.exit(main())
  File "/home/karel/github/cmake-ext/tests/cmake-ext/seqan-prefix/src/seqan/dox/../util/bin/dox.py", line 10, in main
    import seqan.dox.pure
  File "/home/karel/github/cmake-ext/tests/cmake-ext/seqan-prefix/src/seqan/util/py_lib/seqan/dox/pure.py", line 129
    print 'Scanning %s...' % path
                         ^
SyntaxError: Missing parentheses in call to 'print'
dox/CMakeFiles/dox.dir/build.make:57: recipe for target 'dox/CMakeFiles/dox' failed
make[5]: *** [dox/CMakeFiles/dox] Error 1
CMakeFiles/Makefile2:167: recipe for target 'dox/CMakeFiles/dox.dir/all' failed
make[4]: *** [dox/CMakeFiles/dox.dir/all] Error 2
Makefile:160: recipe for target 'all' failed
make[3]: *** [all] Error 2
CMakeFiles/seqan.dir/build.make:110: recipe for target 'cmake-ext/seqan-prefix/src/seqan-stamp/seqan-build' failed
make[2]: *** [cmake-ext/seqan-prefix/src/seqan-stamp/seqan-build] Error 2
CMakeFiles/Makefile2:215: recipe for target 'CMakeFiles/seqan.dir/all' failed
make[1]: *** [CMakeFiles/seqan.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

```

xenigmax:
It will find python2.x and change python path in .sh accordingly.

This will close #1482.
It is also related to #1916.

### DIFF
--- a/dox/CMakeLists.txt
+++ b/dox/CMakeLists.txt
@@ -18,7 +18,6 @@ endif ()
 # find python 2.x
 if(NOT(${PYTHON_VERSION_MAJOR} MATCHES "2"))
     set(_PYTHON2_VERSIONS 2.7 2.6 2.5 2.4 2.3 2.2 2.1 2.0)
-    #set(_PYTHON2_VERSIONS i12.7 12.6 12.5 12.4 12.3 12.2 12.1 12.0)
     unset(_Python_NAMES)
     foreach(_CURRENT_VERSION IN LISTS _PYTHON2_VERSIONS)
         list(APPEND _Python_NAMES "python${_CURRENT_VERSION}")
--- a/dox/CMakeLists.txt
+++ b/dox/CMakeLists.txt
@@ -13,7 +13,7 @@ find_package (PythonInterp)
 if (NOT PYTHONINTERP_FOUND)
     message (STATUS "  Python not found, not building dox as a tests.")
     return ()
-endif ()
+endif (NOT PYTHONINTERP_FOUND)
 
 # find python 2.x
 if(NOT(${PYTHON_VERSION_MAJOR} MATCHES "2"))
--- a/dox/CMakeLists.txt
+++ b/dox/CMakeLists.txt
@@ -30,7 +30,15 @@ if(NOT(${PYTHON_VERSION_MAJOR} MATCHES "2"))
         return()
     endif()
     set(PYTHON_EXECUTABLE "${PYTHON2_EXECUTABLE}")
-    message (STATUS "Found Python 2.x: ${PYTHON_EXECUTABLE}") 
+    message (STATUS "Found Python 2.x: ${PYTHON_EXECUTABLE}")
+
+    # change shell scripts for building dox
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/../dox/run.sh" run_content)
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/../dox/dox_only.sh" dox_only_content)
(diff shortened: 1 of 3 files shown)


dox/CMakeLists.txt
```diff
@@ -15,6 +15,32 @@ if (NOT PYTHONINTERP_FOUND)
...
+    set(PYTHON_EXECUTABLE "${PYTHON2_EXECUTABLE}")
+    message (STATUS "Found Python 2.x: ${PYTHON_EXECUTABLE}")
+
+    # change shell scripts for building dox
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/../dox/run.sh" run_content)
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/../dox/dox_only.sh" dox_only_content)
+    string(REPLACE "python2 ../" "${PYTHON_EXECUTABLE} ../" run_content "${run_content}")
+    string(REPLACE "python2 ../" "${PYTHON_EXECUTABLE} ../" dox_only_content "${dox_only_content}")
+    file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/../dox/run.sh" "${run_content}")
+    file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/../dox/dox_only.sh" "${dox_only_content}")
```
rrahn: Mhmm, would be also possible to set and use an environment variable, which defaults to the standard interpreter if not set?

dox/dox_only.sh
```diff
@@ -1,5 +1,3 @@
 #!/bin/bash
-
 # Run doxygen-style documentation system.
-
```
rrahn: Here you would set the py_interpreter with the Envvar set by cmake and ic not take the default interpr.

dox/CMakeLists.txt
rrahn: ??? No it won't. You can use a special env like ```SEQAN_PYTHON_INTERPRETER```, which is unlikely to interfere with the user's env settings. This is only set by the cmake script. In the shell you can now set: 
```set py_interpreter="${SEQAN_PYTHON_INTERPRETER}-python2}"```
So the behavior is exactly the same as when reading the script replacing the value and writing it out again. But in case of our build system we have set the env and use the correct py version. Or do I miss something?

xenigmax: Then the shell scripts only works after cmake run. For example, if a user reboot their machine it may not have SEQAN_PYTHON_ ... Is this correct?

xenigmax: Ah.. you want me to set env. variables in the shell script. I see.

rrahn:
Ok, I see that there is probably no better solution for that.

PR #2000, opened 2017-01-15
[INTERNAL,FIX] Drop unused CMAKE variable

mr-c:


h-2:
replaced this by #2030 which is against master and should go in before the release... Thanks for submitting the original patch in any case!

PR #2002, opened 2017-01-16
[FIX] run.sh, dox_only.sh : do not change source files

### LINKED ISSUE #1992: cmake edits files that are under version control
#1969 changes existing files that are under version control. This is not good practice, because it causes files locally to change again and again, making it very likely that someone commits them by mistake. If the files really need to be system dependent, than the should be generated completely by cmake, not edited in place.
This could be done by tracking a `dox/dox_only.sh.in` in the repo (but not without the `.in`) and copy-rewriting that to `dox/dox_only.sh`. Or the code could be completely contained in the cmake script and generated upon execution.

@xenigmax can you look into this, as you did the original changes, too? Thx!

xenigmax:
This will close #1992.


xenigmax:
I will make PR to master.

PR #2006, opened 2017-01-16
[FIX] always include pthread with whole-archive

### LINKED ISSUE #1994: static builds with the version checker crash immediately on linux
... this is at least true for lambda, but it probably affects other apps, too.

Deactivating the version check at run-time or compile time fixes the issue, early debugging suggests that the std::promise is doing something it shouldn't,

h-2:
Since the static flag might be included after the seqan config is parsed,
we need to include this unconditionally. It doesn't seem to affect
dynamic linking so there is no problem.

fixes #1994


tests/rna_io/CMakeLists.txt
```diff
@@ -15,7 +15,9 @@ message (STATUS "Configuring tests/rna_io")
 # ----------------------------------------------------------------------------
 
 # Search SeqAn and select dependencies.
-find_package (SeqAn REQUIRED)
+if (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
+    find_package (SeqAn REQUIRED)
+endif ()
```
rrahn: Why is linking against SeqAn needed in the first place? 
Doesn't the test build system take care of it?

h-2: This is the standard for all our tests and apps. In theory we could remove it for the tests, because they are always built "internally", but I see no point in changing it now.
This diff just brings rna_io in line with all other tests.

h-2:
@rrahn why in the world does this break the manual build? And why can I not restart builds on travis anymore?

h-2:
Wow, this PR was falsely created against master 😱

PR #2011, opened 2017-01-19
Fix intel compiler17 warnings

marehr:
Fix intel compiler17 warnings/errors (i.e. ranged based for loops warning and bs_tools rounding issue)

### DIFF
--- a/apps/bs_tools/CMakeLists.txt
+++ b/apps/bs_tools/CMakeLists.txt
@@ -114,6 +114,11 @@ install (FILES LICENSE
 # App Test
 # ----------------------------------------------------------------------------
 
+if (SEQAN_BUILD_SYSTEM STREQUAL "DEVELOP")
+    include("SeqAnConsistentFPFlags")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SEQAN_CONSISTENT_FP_FLAGS}")
+endif()
+
 # (weese:) temporarily disabled app test
 seqan_add_app_test (bs_tools)
 
--- a/tests/platform/test_platform_range_based_for_loops.h
+++ b/tests/platform/test_platform_range_based_for_loops.h
@@ -100,8 +100,6 @@ SEQAN_DEFINE_TEST(test_platform_range_based_for_loops2) {
     #if defined(__INTEL_COMPILER)
         #if __INTEL_COMPILER < 1600 || (__INTEL_COMPILER == 1600 && __INTEL_COMPILER_UPDATE <= 3)
             SEQAN_SKIP_TEST;
-        #else
-            #warning "The Intel Compiler has a bug for this type of range-based for loop (at least until v16.0.3). Please reevaluate this issue for newer versions."
         #endif
     #endif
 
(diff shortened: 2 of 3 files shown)


h-2:
this introduces warnings on clang which doesn't support any of the new options, I think this would then make it difficult to synchronize.
I think it's cool that you researched the different options, but I would rather not add a new cmake file, add new compiler options et cetera just to workaround a bug in the test. @rrahn what do you think?

marehr:
That's interesting there were no errors on our local installed clang compilers.

rrahn:
The answer [here](http://stackoverflow.com/questions/7517588/different-floating-point-result-with-optimization-enabled-compiler-bug) suggest using double casting correctly to get same precision with different optimization levels. I am not sure, if that might help in the case of bs_tools. 
It seems that llvm removed the `-ffloat-store` option some time ago ([read](https://github.com/JuliaLang/julia/issues/8812)). Not sure, why this works with our clang compilers?

h-2:
Maybe we should just remove the test case altogether. :hankey:

marehr:
@rrahn @h-2 I take it back, I overlooked the warning

```
cd /seqan-builds/clang/apps/bs_tools && /usr/bin/clang++-3.9   -DSEQAN_APP_VERSION=\"0.1.8\" -DSEQAN_DATE=\"2017-01-18_23:00:37_+0100\" -DSEQAN_HAS_BZIP2=1 -DSEQAN_HAS_EXECINFO=1 -DSEQAN_HAS_ZLIB=1 -DSEQAN_REVISION=\"077b260\" -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -std=c++14  -W -Wall -pedantic -fstrict-aliasing -Wstrict-aliasing -ffloat-store -ffp-contract=off -fexcess-precision=standard -O3 -DNDEBUG -I/seqan-src/include     -DSEQAN_GLOBAL_EXCEPTION_HANDLER=1 -o CMakeFiles/bisar.dir/bisar.cpp.o -c /seqan-src/apps/bs_tools/bisar.cpp
clang: warning: optimization flag '-ffloat-store' is not supported
clang: warning: optimization flag '-fexcess-precision=standard' is not supported
Linking CXX executable ../../bin/bisar
```

marehr:
@rrahn, @h-2 ping? I removed '-ffloat-store' and '-fexcess-precision=standard' from clang options

rrahn:
@marehr so this fixes the precision problem on all platforms for this app_test? Or is it still necessary to deactivate the test?

marehr:
@rrahn No, it just fixes it for the intel compiler. I forgot the results, but I think it sill produced wrong results for msvc, clang/c2 and wintel. I would have to recheck it.

h-2:
I still propose to remove the test since it obviously depends on behaviour that is not well-defined.

h-2:
@marehr please comment, we really need to release.

h-2:
also @rrahn what is your opinion on the matter?

marehr:
Ah, sorry I thought my point was clear. This, PR fixes the problem for the Intel Compiler on Linux and only adds one new file, which seems from my point of view rather minor. Whether you want to remove the test case altogether could be decided / done in a different PR. I think it is better to keep them, since we know that everything still works - except some minor rounding issues that sometimes occur.

rrahn:
I think we should not disable the test for a patch level release. It should fix things not remove things. The compiler flags do not seem to be very harmful, and since they are in a separate file we can discard it any time easily. So from my point of view, this can be merged.

PR #2013, opened 2017-01-20
[DOC] Updated comment in tutorial (wrong AminoAcid conversion)

cpockrandt:
The AminoAcid table probably changed at some time :grin:


smehringer:
Suggestion: 
I think we should not comment the expected output in the cpp file but append the stdout file at the end, like in many other examples. (The output looks like this: .. includefrags:: a_first_example.stdout)

This way we avoid problems like this.

cpockrandt:
There are alreads *.stdout files. Then we should remove all comments that contain output to avoid differences in the future. What do you think @smehringer?

smehringer:
Yes, that's just what I meant. 😆

cpockrandt:
:grin: :grin: :grin: 

@smehringer Better now? :)

PR #2021, opened 2017-02-07
[FIXES-#2017] Disables simd feature for gcc < 5.0.0

### LINKED ISSUE #2017: severe performance issues of SIMD code with GCC 4.9
Something is  bugged in GCC 4.9 or in our SIMD code in relation to GCC 4.9.
Lambda run times with SIMD enabled are on average 10-15% better than without SIMD. This is true for GCC>=5 and clang.
For gcc-4.9 however the run-time **increases** by a factor of 2x to over 10x! I am not sure what that amount of slow-down depends on, it seems to just increase with the amout of SIMD that you are doing (in a non-linear fashion).

I have verified this with gcc-4.9.4 on FreeBSD and gcc-4.9.2 on Debian.

I propose 
* [ ] we deactivate SIMD on gcc-4.9 for now [ for 2.3.2 ?]
* [ ] we debug the problem if possible [ for 2.4.0 ?]

rrahn:
fixes #2017

### DIFF
--- a/include/seqan/basic/basic_simd_vector.h
+++ b/include/seqan/basic/basic_simd_vector.h
@@ -43,7 +43,7 @@
 #include <tuple>
 
 // Currently only support the following compilers
-#if defined(COMPILER_GCC) || defined(COMPILER_CLANG) || defined(COMPILER_LINTEL)
+#if (defined(COMPILER_GCC) && __GNUC__ >= 5) || defined(COMPILER_CLANG) || defined(COMPILER_LINTEL)
     // Define global macro to check if simd instructions are enabled.
     #define SEQAN_SIMD_ENABLED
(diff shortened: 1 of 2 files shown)


marehr:
Can we do that after my pull request? I changed the structure.

rrahn:
@marehr well, no. Because this is a hotfix and your's is not. :godmode: :

marehr:
Where is this a hotfix? This is completely undocumented code.

rrahn:
But we can keep this on hold and first fix it for 2.4 I guess. Since we only put it here, as 2.3.2 is close to be released.

rrahn:
@marehr can you then just add this to you PR. I'll close this issue then.

PR #2022, opened 2017-02-07
[FIXES-#1997] Reverts documentation about including the FindSeqAn.cmake.

### LINKED ISSUE #1997: Wrong example in doc for cmake projects in manual
In the [example](http://seqan.readthedocs.io/en/develop/Infrastructure/Use/FindSeqAnCMake.html#building-the-project) for using the sedan-config.cmake module, we have an outdated example using the old -DCMAKE_MODULE_PATH=~/devel/seqan/util/cmake

rrahn:
fixes #1997 

Reverts 44159bef2c793dbf75b65a8ff8a8ed3e3abc7713

### DIFF
--- a/manual/source/Infrastructure/Use/FindSeqAnCMake.rst
+++ b/manual/source/Infrastructure/Use/FindSeqAnCMake.rst
@@ -77,26 +77,54 @@ First you should create a build directory, i.e. for cmake-builds everything happ
    # mkdir -p ~/devel/my_project-build/release
    # cd ~/devel/my_project-build/release
 
-By default, the ``cmake`` program will look for ``seqan-config.cmake`` in a predefined set of directories.
-Please read the documentation `find_project <https://cmake.org/cmake/help/v3.0/command/find_package.html>`_ to find out, which standard paths are searched.
-Depending on how you :ref:`installed SeqAn <infra-use-install>` it might be found by cmake automatically. 
-If not, you have to give the path containing the config file to cmake via the ``SeqAn_DIR`` argument on the command line, or you add the installation prefix of ``SeqAn`` via the ``CMAKE_PREFIX_PATH`` variable.
+Cmake supports two different modes to load settings from an external project: The **module** and the **config** mode. 
+Please read the `cmake documentation <https://cmake.org/cmake/help/v3.0/command/find_package.html>`_ to learn more about this feature.
 
-Also, CMake will look for the SeqAn include files in central locations such as ``/usr/local/include``. Again, depending on your installation this might *just work*. If not, you need to specify the location via the ``SEQAN_INCLUDE_PATH`` argument.
+Install SeqAn from package maintainer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


h-2:
Well the green stuff is not really correct now either :expressionless: 
The thing is that the module is never in a system-wide location. The config might be! So for system-wide detection the config is used, and for local detection "probably the module" and "possibly the config".
It's a :boom:'up thing.

rrahn:
This is all so 😞 ing

rrahn:
rebuild

rrahn:
@h-2 updated the docs

manual/source/Infrastructure/Use/FindSeqAnCMake.rst
```diff
@@ -77,26 +77,54 @@ First you should create a build directory, i.e. for cmake-builds everything happ
...
+      -DCMAKE_PREFIX_PATH="$HOME/local/seqan"
+
 
-If you instead did a full git checkout to your home-directory in the previous step, it might look like this:
+SeqAn from GitHub clone
+~~~~~~~~~~~~~~~~~~~~~~~
+
+In case you obtained SeqAn from a git clone, we still recommend to use the method described above and set the ``CMAKE_PREFIX_PATH`` to your SeqAn clone.
+But in addition you also have to specify the ``SEQAN_INCLUDE_PATH`` variable to find the SeqAn headers. 
+Assume you have cloned SeqAn into ``~/devel/seqan``, then your setup could look as the following:
```
h-2: I don't understand why this is different from the paragraph above. Why do they not need to set the INCLUDE_PATH?

manual/source/Infrastructure/Use/FindSeqAnCMake.rst
```diff
@@ -77,26 +77,54 @@ First you should create a build directory, i.e. for cmake-builds everything happ
...
+   
+   # cmake ../../my_project \
+      -DCMAKE_PREFIX_PATH="$HOME/devel/seqan/util/cmake" \
+      -DSEQAN_INCLUDE_PATH="$HOME/devel/seqan/include"
+
+Backwards compatibility
+~~~~~~~~~~~~~~~~~~~~~~~
 
+Before SeqAn 2.3 we used the module mode to setup SeqAn as an external project.
+To allow backwards compatibility we added a symlink from the ``FindSeqAn.cmake`` to ``seqan-config.cmake``.
```
h-2: It's not a symlink anymore, but just includes the other cmake file.

manual/source/Infrastructure/Use/FindSeqAnCMake.rst
rrahn: They have to set the include path. Read sentence at line 107. But in addition ...
Simply stating that the approach with selecting the PrefixPath is preferred over the module_path option for future compatibility.
There is also a code example.

manual/source/Infrastructure/Use/FindSeqAnCMake.rst
rrahn: Ok, but does it make any difference to the user? I will write call it redirect.

PR #2027, opened 2017-02-10
[FIX] GLIBCXX_PARALLEL slows down regex in arg parser

cpockrandt:


apps/fiona/fiona.cpp
```diff
@@ -79,11 +79,12 @@
 
     #if defined(STDLIB_GNU)
         #include <parallel/algorithm>
-        #define SORT ::__gnu_parallel::sort
+        #define PARALLELTAG Parallel()
     #else
-        #define SORT ::std::sort
+        #define PARALLELTAG Serial()
```
rrahn: This you don't need anymore with the seqan implementation. There is a parallel overload for non gnu_parallel version simply calling the sequential sort.
Just specifying the Parallel tag in the function call is sufficient. It will cover the same behaviour as before, but with less boilerplate.

apps/fiona/fiona.cpp
```diff
@@ -656,7 +657,7 @@ namespace seqan
             }
         
         // sort them descendingly by bucket size
-        SORT(begin(bktIdx, Standard()), end(bktIdx, Standard()), GreaterBucketSize<TDir>(dir));
+        sort(bktIdx, GreaterBucketSize<TDir>(dir), PARALLELTAG);
```
rrahn: Here just use ```Parallel()``` as tag.

apps/fiona/fiona.cpp
cpockrandt: I wasn't aware of that and didn't find it in the code in a quick way.

rrahn:
👍

PR #2028, opened 2017-02-10
[FIXES-#2026] Fixes gnk ctd problems.

### LINKED ISSUE #2026: CTD support broken by new function _typeToString
GenericKnimeNodes expects `int` instead of `integer`as a type. When we replaced the hardcoded types [here](https://github.com/seqan/seqan/blob/966da62f2ba61867359cad18f4db2dbaf431243e/include/seqan/arg_parse/arg_parse_ctd_support.h#L394) with _typeToString() the `int` was replaced by `integer` ([here](https://github.com/seqan/seqan/blob/966da62f2ba61867359cad18f4db2dbaf431243e/include/seqan/arg_parse/arg_parse_argument.h#L234) )

This causes a failure in generating KNIME nodes. 

And one more thing 
`version-check` parameter has a `bool` type but the restrictions are `1,ON,TRUE,T,YES,0,OFF,FALSE,F,NO`.

This is also causing a problem. We either have to make it a `string` type or restrict the "restrictions" to TRUE and FALSE. (See [`getRestrictions`](https://github.com/seqan/seqan/blob/966da62f2ba61867359cad18f4db2dbaf431243e/include/seqan/arg_parse/arg_parse_ctd_support.h#L164) )

rrahn:
fixes #2026

### DIFF
--- a/include/seqan/arg_parse/arg_parse_ctd_support.h
+++ b/include/seqan/arg_parse/arg_parse_ctd_support.h
@@ -173,7 +173,15 @@ _getRestrictions(std::vector<std::string> & restrictions, ArgParseArgument const
              valid != opt.validValues.end();
              ++valid)
         {
-            appendValue(restrictions, *valid);
+            switch (opt._argumentType)
+            {
+                case ArgParseArgument::ArgumentType::BOOL:
+                    if (*valid == "TRUE" || *valid == "FALSE")
+                        appendValue(restrictions, *valid);
+                    break;
+                default:
+                    appendValue(restrictions, *valid);
+            }
         }
     }
     else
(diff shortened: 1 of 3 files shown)


include/seqan/arg_parse/arg_parse_ctd_support.h
```diff
@@ -281,6 +289,17 @@ inline std::string _getManual(ArgumentParser const & me)
 }
 
 // ----------------------------------------------------------------------------
+// Function _toValidGnkTypeSpecifier()
+// ----------------------------------------------------------------------------
+
+inline std::string _toValidGnkTypeSpecifier(std::string type)
```
h-2: please don't copy, even if it's not performance sensitive code :nail_care:

PR #2030, opened 2017-02-10
[FIX] remove flag that was not being set by cmake anymore, anyhow

h-2:
supersedes #2000 which is created against develop


h-2:
@rrahn test failures unrelated

PR #2033, opened 2017-02-13
[FIX] Fixes gnk ctd problems.

### LINKED ISSUE #2026: CTD support broken by new function _typeToString

temehi:
This fixes #2026

### DIFF
--- a/include/seqan/arg_parse/arg_parse_argument.h
+++ b/include/seqan/arg_parse/arg_parse_argument.h
@@ -244,27 +244,27 @@ inline std::string _typeToString(ArgParseArgument const & me)
         break;
 
     case ArgParseArgument::INPUT_FILE:
-        typeName = "input_file";
+        typeName = "input-file";
         break;
 
     case ArgParseArgument::OUTPUT_FILE:
-        typeName = "output_file";
+        typeName = "output-file";
         break;
 
     case ArgParseArgument::INPUT_PREFIX:
-        typeName = "input_prefix";
+        typeName = "input-prefix";
         break;
 
     case ArgParseArgument::OUTPUT_PREFIX:
-        typeName = "output_prefix";
+        typeName = "output-prefix";
         break;
 
     case ArgParseArgument::INPUT_DIRECTORY:
-        typeName = "input_directory";
+        typeName = "input-directory";
         break;
 
     case ArgParseArgument::OUTPUT_DIRECTORY:
-        typeName = "output_directory";
+        typeName = "output-directory";
         break;
 
     default:
--- a/include/seqan/arg_parse/arg_parse_argument.h
+++ b/include/seqan/arg_parse/arg_parse_argument.h
@@ -244,27 +244,27 @@ inline std::string _typeToString(ArgParseArgument const & me)
         break;
 
     case ArgParseArgument::INPUT_FILE:
-        typeName = "input-file";
+        typeName = "input_file";
         break;
 
     case ArgParseArgument::OUTPUT_FILE:
-        typeName = "output-file";
+        typeName = "output_file";
         break;
 
     case ArgParseArgument::INPUT_PREFIX:
-        typeName = "input-prefix";
+        typeName = "input_prefix";
         break;
 
     case ArgParseArgument::OUTPUT_PREFIX:
-        typeName = "output-prefix";
+        typeName = "output_prefix";
         break;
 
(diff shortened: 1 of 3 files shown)


include/seqan/arg_parse/arg_parse_argument.h
```diff
@@ -244,27 +244,27 @@ inline std::string _typeToString(ArgParseArgument const & me)
...
 
     case ArgParseArgument::INPUT_DIRECTORY:
-        typeName = "input_directory";
+        typeName = "input-directory";
         break;
 
     case ArgParseArgument::OUTPUT_DIRECTORY:
-        typeName = "output_directory";
+        typeName = "output-directory";
         break;
```
h-2: don't these changes affect something other than the CTD stuff?
@smehringer  what is your opinion on this?

h-2: also, can ctd handle the directories?

temehi: @h-2 take a look at this commit https://github.com/seqan/seqan/commit/9ca8c89eac64fe7afd058471fd9bcdc174076841

It was this way before.

h-2: ok, but what about the directories, thats new, do you know if it works?

h-2:
@smehringer do you approve of these changes? Then I will merge!

include/seqan/arg_parse/arg_parse_ctd_support.h
```diff
@@ -289,15 +287,69 @@ inline std::string _getManual(ArgumentParser const & me)
...
+inline std::string _toValidGKNTypeSpecifier(ArgParseArgument const & me)
 {
-    if (type == "integer")
-        return "int";
-    return type;
+    std::string typeName = "";
+
+    switch (me._argumentType)
+    {
+    case ArgParseArgument::BOOL:
```
h-2: please indent the case labels properly!

include/seqan/arg_parse/arg_parse_ctd_support.h
```diff
@@ -289,15 +287,69 @@ inline std::string _getManual(ArgumentParser const & me)
...
+        break;
+
+    case ArgParseArgument::OUTPUT_DIRECTORY:
+        typeName = "output-prefix";
+        break;
+
+    default:
+        typeName = "unknown";
+        break;
+    }
```
h-2: please specify only those cases that differ and make `default` call the other function to handle unchanged values.

h-2:
You did not address my second comment.

h-2:
Ah, sorry. I looked in the wrong place. I see that you did address it! However, don't change the other function, if you are only changing whitespaces. Never mix whitespace changes with functional changes!

Also you can greatly simplifiy your function by just return the string literal instead of assign to the local variable first...

temehi:
@h-2. Sorry for my late response. 
Wouldn't that apply to the original function as well?

Should I just change the new one as you said and remove my remove the white space changes from the original function? Or should I also improve it?

h-2:
yes, it would apply to the original function, but please just change it in the new one, don't change the old function at all, if you are not changing the behaviour.
If you want to change the old function, please make sure that you have a clean commit history, i.e. a single commit for your functional changes, and single commit for formatting/whitespace changes.

PR #2058, opened 2017-04-16
CMake fixes

SoapGentoo:
**[FEATURE] Use GNUInstallDirs to allow changing target directories**

* `GNUInstallDirs` allows for changing all target directories, which makes parallel installations of SeqAn a lot easier.
* Simplify the `includedir` specification in the `.pc` file.
* Nest all applications' documentation in the seqan docdir, as this prevents pollution of the global docdir and keeps seqan apps together. This can be changed by using `CMAKE_INSTALL_DOCDIR`.
* Install man page into global mandir, as `man` will not find it otherwise.
* Install CMake support file and `.pc` file into arch-independent `CMAKE_INSTALL_DATAROOTDIR`.

**[FIX] Remove code generation flags**

* `-fstrict-aliasing` should not be added to `CXXFLAGS` unconditionally, as it is already implied by `-O2` and `-O3`. When compiling with `-O0`, no code generation affecting flags should be added to the compile line.
  See also:
    https://gcc.gnu.org/onlinedocs/gcc-6.3.0/gcc/Optimize-Options.html#index-O2-707
    
* `-Wstrict-aliasing` is completely unnecessary, as it is implied by -Wall.
  See also:
    https://gcc.gnu.org/onlinedocs/gcc-6.3.0/gcc/Warning-Options.html#index-Wall-304

### DIFF
--- a/apps/alf/CMakeLists.txt
+++ b/apps/alf/CMakeLists.txt
@@ -56,7 +56,7 @@ endif (NOT SEQAN_PREFIX_SHARE_DOC)
 
 # Install alf in ${PREFIX}/bin directory
 install (TARGETS alf
-         DESTINATION bin)
+         DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 # Install non-binary files for the package to "." for app builds and
 # ${PREFIX}/share/doc/alf for SeqAn release builds.
--- a/apps/bs_tools/CMakeLists.txt
+++ b/apps/bs_tools/CMakeLists.txt
@@ -99,7 +99,7 @@ endif (NOT SEQAN_PREFIX_SHARE_DOC)
 
 # Install bs_tools in ${PREFIX}/bin directory
 install (TARGETS bisar casbar four2three
-         DESTINATION bin)
+         DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 # Install non-binary files for the package to "." for app builds and
 # ${PREFIX}/share/doc/bs_tools for SeqAn release builds.
--- a/apps/dfi/CMakeLists.txt
+++ b/apps/dfi/CMakeLists.txt
@@ -56,7 +56,7 @@ endif (NOT SEQAN_PREFIX_SHARE_DOC)
 
 # Install dfi in ${PREFIX}/bin directory
 install (TARGETS dfi
-         DESTINATION bin)
+         DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 # Install non-binary files for the package to "." for app builds and
 # ${PREFIX}/share/doc/dfi for SeqAn release builds.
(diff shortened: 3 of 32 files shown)


h-2:
@SoapGentoo I will have a look at it next week!

SoapGentoo:
@h-2 thanks, would be great to get it in :smile:

SoapGentoo:
@h-2 any movement here? it's a pretty simple change

h-2:
@SoapGentoo sorry, I have been quite busy and I need to build packages to see that the effects are actually the same. I hope, I will get it done tomorrow!

In general: can you sign the clahub?

SoapGentoo:
@h-2 done

SoapGentoo:
@h-2 still no movement?

h-2:
@SoapGentoo Sorry, there are some hold-ups here and I just returned from vacations. I guarantee it will be reviewed in time for the next release, but can't promise it will happen in the next days.

SoapGentoo:
@h-2 here's an idea: what if I change the **default** `CMAKE_INSTALL_DOCDIR` to not nest it in `share/doc/seqan` but `share/doc`. 98% of users won't change the defaults, yet it allows us distributions to still nest everything under `share/doc/seqan`? Then both our use cases are covered.

h-2:
Sure, that's decent solution.

SoapGentoo:
@h-2 all done. For the record:

Before:
`/usr/local/share/doc/seqan/html/specialization_AdjacencyIterator.html`

After (with redefined default):
`/usr/local/share/doc/html/specialization_AdjacencyIterator.html`

After with specifying `-DCMAKE_INSTALL_DOCDIR=share/doc/seqan-2` (Gentoo layout):
`/usr/local/share/doc/seqan-2/html/specialization_AdjacencyIterator.html`

h-2:
@SoapGentoo Wait, but for the library it should always install into `/usr/local/share/doc/seqan`, not in `/usr/local/share/doc` ❓
I know it's not easy to pick things for the library and the apps apart, but the library is what is actually called seqan, it should be in seqan subfolders. The apps should have their own folders by default – and if packaged together could be set to also install into the seqan subfolders (although even then I would recommend putting their stuff into a signal `seqan-apps` subfolder and not the seqan folder).

h-2:
@SoapGentoo sorry for all the back and forth on this. We will do a new stable release quite soon so I would be interested in getting this merged... Can you set it up so that the library stuff always goes into a seqan subfolder by default and the apps go into their own folders by default? As long as it is configurable you can change it downstream to whatever anyway, or not?
Thanks!

SoapGentoo:
@h-2 sorry for the long wait, everything should be in place now.

SoapGentoo:
@h-2 the jenkins builds are failing and I cannot look at the logs.

h-2:
@SoapGentoo The Jenkins-windows stuff is still broken, we can ignore it.

rrahn:
@h-2 what is the state here?

h-2:
@SoapGentoo thank you for your work on this!

PR #2066, opened 2017-05-12
[FEATURE] add isBase function for traverser.

rrahn:
Extends the interface of the jst traverser to check whether the traverser is at the base or in some branch.


rrahn:
rebuild

Issue #1899, opened 2016-10-28
option for # of thread in bgzf stream


xenigmax:
In my opinion, we should not fix this. (16 threads always)
At least, we have to provide an option for changing this number.

https://github.com/seqan/seqan/blob/master/include/seqan/stream/iostream_bgzf.h#L496-L498

lkuchenb:
This can is actually quite an issue. Programs using SeqAn I/O that are run on a cluster will be kicked from the cluster by the grid engine if the user didn't happen to request a large enough  number of threads, without any way to control this as a developer or user.

PR #2074, opened 2017-05-31
Add BGZF_NUM_THREADS

lkuchenb:
See #1899 for my remarks on the potential trouble with spawning a hardcoded number of threads.

My understanding is that it is unlikely that a consistent mechanism for developers to control how many threads seqan will use for what will be implemented in 2.x.

This patch provides at least compile time control to devs writing applications against seqan for BAM I/O.

### DIFF
--- a/include/seqan/stream/iostream_bgzf.h
+++ b/include/seqan/stream/iostream_bgzf.h
@@ -20,6 +20,10 @@
 #ifndef INCLUDE_SEQAN_STREAM_IOSTREAM_BGZF_H_
 #define INCLUDE_SEQAN_STREAM_IOSTREAM_BGZF_H_
 
+#ifndef SEQAN_BGZF_NUM_THREADS
+#define SEQAN_BGZF_NUM_THREADS 16
+#endif
+
 namespace seqan {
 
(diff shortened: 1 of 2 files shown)


include/seqan/stream/iostream_bgzf.h
```diff
@@ -20,6 +20,10 @@
 #ifndef INCLUDE_SEQAN_STREAM_IOSTREAM_BGZF_H_
 #define INCLUDE_SEQAN_STREAM_IOSTREAM_BGZF_H_
 
+#ifndef BGZF_NUM_THREADS
```
h-2: Since it is a SeqAn specific MACRO, please prefix it with `SEQAN_`

h-2:
@lkuchenb I agree with the solution in general. @rrahn What do you think?

h-2:
@lkuchenb Please also add documention to https://github.com/seqan/seqan/blob/master/manual/source/Infrastructure/Use/CustomBuildSystem.rst#preprocessor-defines-affecting-seqan

lkuchenb:
- Added docs
- Changed name to `SEQAN_BGZF_NUM_THREADS`

PR #2076, opened 2017-06-01
[FEATURE] Added contrib package including vs15 contribs.

rrahn:


rrahn:
has no effect yet to the current build system. so I'll merge it immediately.

PR #2092, opened 2017-07-05
[FIX] Fixes bug in cigar-to-align helper function

rrahn:


cpockrandt:
Is there an issue to this PR? It's a bit difficult to review it when I have no clue what the problem was. ;)

rrahn:
The problem was, that the function should extract the alignment from the CIGAR string of BamRecord. But to extract the corresponding infix it needs the begin pos and the length of the infix. But instead the old code computed the end-position, leading to wrong infix sizes.

rrahn:
@cpockrandt ping

rrahn:
Tests are only failing due to timeout of one test, which us unrelated.

PR #2094, opened 2017-07-05
Fix/partial sum serial implementation

### LINKED ISSUE #2093: Timeouts for test_index_bifm
There's a bug when compiled with OpenMP (and gcc). I narrowed it down to the method

`partialSum()` in `parallel/parallel_algorithms.h` which is used for creating the Count table used for the LF mapping. For some reason the program spends > 400s in `do_spin`.

Anyway, there shouldn't be done any parallelization at all (when called by the FMindex). We usually never build indices on alphabet sizes larger than 10 (which will lead to 9 summations).

cpockrandt:
Fixes #2093

### DIFF
--- a/include/seqan/parallel/parallel_algorithms.h
+++ b/include/seqan/parallel/parallel_algorithms.h
@@ -160,7 +160,7 @@ partialSum(TTarget &target, TSource const &source, Tag<TParallelTag> parallelTag
 
     // STEP 1: compute sums of all subintervals (in parallel)
     //
-    SEQAN_OMP_PRAGMA(parallel for)
+    SEQAN_OMP_PRAGMA(parallel for num_threads(length(splitter)))
     for (int job = 0; job < (int)length(splitter) - 1; ++job)
         localSums[job + 1] = sum(infix(source, splitter[job], splitter[job + 1]), Serial());
 
@@ -171,7 +171,7 @@ partialSum(TTarget &target, TSource const &source, Tag<TParallelTag> parallelTag
 
     // STEP 3: compute partial sums of each subinterval starting from offset (in parallel)
     //
-    SEQAN_OMP_PRAGMA(parallel for)
+    SEQAN_OMP_PRAGMA(parallel for num_threads(length(splitter)))
     for (int job = 0; job < (int)length(splitter); ++job)
     {
         TConstIterator it = begin(source, Standard()) + splitter[job];
--- a/include/seqan/pipe/pipe_sampler.h
+++ b/include/seqan/pipe/pipe_sampler.h
@@ -618,7 +618,7 @@ namespace seqan
         TLimitsString const &limits = me.limits;
         int64_t seqCountPlusOne = length(me.limits);
 
-        SEQAN_OMP_PRAGMA(parallel for reduction(+:sum))
+        SEQAN_OMP_PRAGMA(parallel for reduction(+:sum) if (seqCountPlusOne > 99))
         for (int64_t i = 1; i < seqCountPlusOne; ++i)
         {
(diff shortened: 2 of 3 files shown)


h-2:
indexBiFM still timeouts on travis...

cpockrandt:
I know ... but even when I remove parallelization from the reverse function, there is still parallelization going on that I couldn't locate yet ...

h-2:
what is the state of this?

include/seqan/parallel/parallel_algorithms.h
```diff
@@ -140,9 +140,9 @@ sum(TSequence const &seq)
  * @see sum
  */
 
-template <typename TTarget, typename TSource, typename TParallelTag>
+template <typename TTarget, typename TSource>
 inline typename Value<TSource>::Type
-partialSum(TTarget &target, TSource const &source, Tag<TParallelTag> parallelTag)
+partialSum(TTarget &target, TSource const &source, Parallel)
```
rrahn: These changes would break the Serial implementation wouldn't they?
The thing is that the Splitter with the Serial tag would simply make it a single chunk leading to a sequential execution of the following for-loops. So that should not changed.

include/seqan/parallel/parallel_algorithms.h
```diff
@@ -189,6 +189,30 @@ partialSum(TTarget &target, TSource const &source, Tag<TParallelTag> parallelTag
...
+   TIterator dstIt = begin(target, Standard());
+   TValue sum = 0;
+   for (; it != end(source, Standard()); ++it, ++dstIt)
+   {
+       sum += *it;
+       *dstIt = sum;
+   }
+
+   return back(target);
+}
```
rrahn: Not necessary if you revert the changes above.

include/seqan/modifier/modifier_reverse.h
```diff
@@ -595,7 +595,7 @@ template < typename TText >
 inline SEQAN_FUNC_DISABLE_IF(Is<StlContainerConcept<TText> >)
 reverse(TText & text)
 {
-    reverse(text, Parallel());
+    reverse(text, Serial());
```
rrahn: Are you sure about these changes?

rrahn:
@cpockrandt ping

cpockrandt:
I started looking into this issue again. Still working on it but it seems that the generic `partialSum` implementation does not work efficiently in serial mode. I separated the Serial implementation into its own method which increased the running time significantly. This is only true for debug mode though, not if the code is optimized. Probably because for a for-loop with one iteration (one job), a separate thread is spawned each time which is avoided with optimization on.

Afaik the nightly tests are built in debug mode. But there's still a significant time difference. I guess it's time same problem as above but I didn't identify the code part yet.

cpockrandt:
In other words: even though `partialSum` is called in serial mode which will set the number of jobs to 1, it is significantly slower than simply removing the OpenMP macro in [line 174](https://github.com/cpockrandt/seqan/blob/7d7f556597cf8b1a71718c2b96383a2a9b8a2c85/include/seqan/parallel/parallel_algorithms.h#L174)

@rrahn What are your thoughts? Shall I make separate serial implementations like in the first commit?

rrahn:
Mhmm, have you tried restricting the number of threads of the for loop to the length of the splitter?
And no, we run the nightlies in optimised mode but with assertions enabled.

cpockrandt:
Thanks, great idea! That worked 😃 

Unfortunately I didn't identify the other bottlenecks yet.

cpockrandt:
Parallelization problems are fixed now. I also reduced the test case.

Issue #2067, opened 2017-05-12
bad_alloc with seqan_tcoffee


mikael-s:
Hi!

When launching `seqan_tcoffee` (version 1.13.7, seqan: 2.3.2) on a pair of similar sequences I encounter an uncaught `bad_alloc` exception.

The command line is `seqan_tcoffee  -s test.fasta -a dna` and the exception is raised almost immediately:
```
/home/salson/seqan-seqan-v2.3.2/include/seqan/basic/basic_exception.h:363 FAILED!  (Uncaught exception of type std::bad_alloc: std::bad_alloc)

stack trace:
  0          [0x45db8e]  seqan::ClassTest::fail() + 0xe
  1          [0x4500a8]  seqan_tcoffee()
  2    [0x7f6d2481cbb6]  /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5ebb6)
  3    [0x7f6d2481cc01]  /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5ec01)
  4    [0x7f6d2481ce19]  /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5ee19)
  5    [0x7f6d2481d339]  /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5f339)
  6          [0x49f073]  void seqan::resize<unsigned char, seqan::Tag<seqan::FullDPMatrix_> >(seqan::DPMatrix_<unsigned char, seqan::Tag<seqan::FullDPMatrix_> >&) + 0x133
  7          [0x4f9213]  seqan::Value<seqan::Score<int, seqan::Tag<seqan::Simple_> >, 0>::Type seqan::_computeAlignment<int, seqan::Tag<seqan::AffineGaps_>, seqan::String<seqan::TraceSegment_<unsigned long, un3
  8          [0x51231f]  void seqan::globalMsaAlignment<seqan::StringSet<seqan::String<seqan::SimpleType<unsigned char, seqan::Dna5_>, seqan::Alloc<void> >, seqan::Dependent<seqan::Tag<seqan::TagInsist_> > >, f
  9          [0x514488]  void customizedMsaAlignment<seqan::SimpleType<unsigned char, seqan::Dna5_>, seqan::Score<int, seqan::Tag<seqan::Simple_> > >(seqan::MsaOptions<seqan::SimpleType<unsigned char, seqan::D8
[...]

PR #2097, opened 2017-07-14
proper t-coffee error message if global alignment fails

joergi-w:
see issue #2067

### DIFF
--- a/include/seqan/graph_msa/graph_align_tcoffee_library.h
+++ b/include/seqan/graph_msa/graph_align_tcoffee_library.h
@@ -995,17 +995,7 @@ appendSegmentMatches(StringSet<TString, Dependent<TSpec> > const& str,
 
         // Alignment
         TSize from = length(matches);
-        TScoreValue myScore{};
-        try {
-            myScore = globalAlignment(matches, pairSet, score_type, ac, Gotoh());
-        }
-        catch (const std::bad_alloc & exception)
-        {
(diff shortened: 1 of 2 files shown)


h-2:
Hm, shouldn't you have a global exception handler somewehere that also catches other bad_allocs?

include/seqan/graph_msa/graph_align_tcoffee_library.h
```diff
@@ -995,7 +995,17 @@ appendSegmentMatches(StringSet<TString, Dependent<TSpec> > const& str,
...
-        TScoreValue myScore = globalAlignment(matches, pairSet, score_type, ac, Gotoh());
+        TScoreValue myScore{};
+        try {
+            myScore = globalAlignment(matches, pairSet, score_type, ac, Gotoh());
+        }
+        catch (const std::bad_alloc & exception)
+        {
+            std::cerr << "Allocation for globalAlignment failed. Use smaller data or try seeded alignment. \n"
+                      << exception.what() << std::endl;
+            exit(1);
```
rrahn: please use exit status: http://en.cppreference.com/w/cpp/utility/program/EXIT_status

joergi-w:
Thanks for the advice! I put the exception handler to a higher level, so it catches bad_allocs also for different program settings. And I use the EXIT_FAILURE macro if something goes wrong.

h-2:
rebuild

Issue #1944, opened 2016-11-21
Unexpected 'hasAlignments' returned from BamIO's jumpToRegion function


hannespetur:
Hello,

I think I might have noticed a bug in the seqan::jumpToRegion function, its API page is here: http://docs.seqan.de/seqan/master/specialization_BamFileIn.html#BamFileIn#jumpToRegion 
More specificly, sometimes the 'hasAlignments' parameter is unexpectecly set to true. Here is how it is defined:
> The hasAlignments parameter should be set to true if there are any alignments in [pos, posEnd).

And here is the implementation of the function: https://github.com/seqan/seqan/blob/master/include/seqan/bam_io/bam_index_bai.h#L294

In the implementation, it seems 'hasAlignments' is set to true as long as there are any reads in the specified contig, disregarding if their are in [pos, posEnd) or not.

You can reproduce this behavior by adding this test to `tests/bam_io/test_bam_index.h:79`
```
SEQAN_ASSERT(jumpToRegion(bamFile, found, 0, 20, 100, baiIndex));
SEQAN_ASSERT_NOT(found);
```
The second assertion will fail because 'found' is true, even though no alignments overlap [pos, posEnd] in `tests/bam_io/small.bam`. So it seems, at least to me, that either this is a bug in the implementation or 'hasAlignments' is poorly defined.

Best regards,
Hannes Pétur

hannespetur:
Thank you!

PR #2098, opened 2017-07-14
[FIX] fixed hasAlignements logic

temehi:
This fixes issue #1944


rrahn:
I think the 4.9 build can be ignored here and we can safely merge this.

PR #2113, opened 2017-09-13
[INTERNAL] Allow the contrib package to be found by looking up an environment variable.

rrahn:
Needed to get a cleaner test environment for the CI builds via Jenkins

### DIFF
--- a/util/cmake/SeqAnContribs.cmake
+++ b/util/cmake/SeqAnContribs.cmake
@@ -35,88 +35,68 @@
 # ============================================================================
 
 if (WIN32)
-  # For all contrib versions...
-  foreach (_SEQAN_CONTRIB_VERSION D20170601 D20160115)
+# For all contrib versions...
+foreach (_SEQAN_CONTRIB_VERSION D20170601 D20160115)
     set (_SEQAN_CONTRIB_DIR "seqan-contrib-${_SEQAN_CONTRIB_VERSION}")
 
-	  # Determine architecture for the precompiled contribs.
-	  if (CMAKE_SIZEOF_VOID_P EQUAL 8)
-		set (CONTRIB_ARCH "x64")
-	  else ()
-		set (CONTRIB_ARCH "x86")
-	  endif ()
-
-	  # Try to figure out where the user installed the contrib.  We expect
-	  # it to be either in C:\, or one of the Program Files dirs.
-	  #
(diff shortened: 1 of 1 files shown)


rrahn:
rebuild

rrahn:
Travis build error not related.

rrahn:
rebuild

h-2:
Anything I should review here or is this WIP?

rrahn:
!rebuild

util/cmake/SeqAnContribs.cmake
```diff
@@ -113,5 +119,27 @@ if (WIN32)
       endif (DEFINED SEQAN_CONTRIB_BASE)
 
     endforeach ()  # all contrib versions.
-endif (WIN32)
 
+    foreach (_BOOST_VERSION 1_65_1 1_65_0 1_64_0 1_63_0 1_62_0 1_61_0 1_60_0 1_59_0 1_58_0)
```
marehr: I think we use `find(boost)` or so. There should be a variable to set the base path for finding boost.

This should make this code completely obsolete.

util/cmake/SeqAnContribs.cmake
```diff
@@ -69,6 +69,12 @@ if (WIN32)
 		  set (SEQAN_CONTRIB_BASE "C:\\${_SEQAN_CONTRIB_DIR}-${CONTRIB_ARCH}")
 		endif ()
 	  endif (NOT DEFINED SEQAN_CONTRIB_BASE)
+    # Try to find in specific environment variable
```
marehr: You could iterate over all possible directory paths.

In this code we always do the same:
* If SEQAN_CONTRIB_BASE is not known, yet
   * Look if the new path is existing
   * If not repeat

This is basically a for loop over all directories.

That can be  simplified:
```cmake
set(_paths "$ENV{ProgramW6432}" "$ENV{ProgramFiles}" "C:\\" "$ENV{SEQAN_WIN_CONTRIB_DIRECTORY}")

foreach(path $_paths)
        set(path "$path/${_SEQAN_CONTRIB_DIR}-${CONTRIB_ARCH}"
        if (IS_DIRECTORY "$path")
            set (SEQAN_CONTRIB_BASE "$path")
            break ()
        endif ()
endforeach()
```
(Might be no valid cmake syntax)

rrahn:
!rebuild

PR #2116, opened 2017-09-19
[FIX] Fixes alignment of arguments in text-tool-doc if no argument description is present.

rrahn:
When multiple arguments are present, the tool doc displays the arguments `1..n` indented by `10` spaces, if no description to the list item is present.


smehringer:
LGTM

Issue #2112, opened 2017-09-07
readRecords from GffFileIn with ending "###" line fails with "seqan::UnexpectedEnd"


okulev:
Using Debian seqan2-dev 2.3.1+dfsg-4 and genome annotations tutorial snippet:

```
FragmentStore<> store;
GffFileIn gffFile(toCString(annFilename));
readRecords(store, gffFile);
```

on uncompressed ftp://ftp.ensembl.org/pub/release-90/gff3/homo_sapiens/Homo_sapiens.GRCh38.90.gff3.gz fails with the following message:

```
terminate called after throwing an instance of 'seqan::UnexpectedEnd'
  what():  Unexpected end of input.
Aborted
```

After removing the last line `###` from the GFF3 file, it is parsed successfully.

rrahn:
Mhmm, as far as I understand gff3 format, the `##` indicates a new section. 
Following this document here: http://gmod.org/wiki/GFF3
So if your file ends with these section indicator, the parser assumes more data. 
So in this case I wouldn't strictly say that the parser is wrong but the file format is not following the specifications.

okulev:
The GMOD document talks about sections only in FASTA section explanation, for FASTA sequence that follows `##FASTA` line.

The [specification](https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md) talks about pragma directives:

> Lines beginning with '##' are directives (sometimes called pragmas or meta-data) and provide meta-information about the document as a whole.

The `###` used by Ensembl is specifically listed as such directive:

> This directive (three # signs in a row) indicates that all forward references to feature IDs that have been seen to this point have been resolved. After seeing this directive, a program that is processing the file serially can close off any open objects that it has created and return them, thereby allowing iterative access to the file. Otherwise, software cannot know that a feature has been fully populated by its subfeatures until the end of the file has been reached. It is recommended that complex features, such as the canonical gene, be terminated with the ### notation.

rrahn:
Ok, thanks for pointing to this.
We'll take care of it.

rrahn:
This should be fixed.

PR #2126, opened 2017-10-10
[FIX] fixes GFF reading in the case of '###' terminated files

temehi:
This fixes issue #2112.
The fix also includes  possibility to ignore empty lines and right before directives/comments

### DIFF
--- a/include/seqan/gff_io/gff_io_base.h
+++ b/include/seqan/gff_io/gff_io_base.h
@@ -349,12 +349,13 @@ void readRecord(GffRecord & record, CharString & buffer, TFwdIterator & iter)
 {
     IsNewline isNewline;
 
-    // skip commented lines
+    // skip commented lines as well as ## directives
+    skipUntil(iter, NotFunctor<IsWhitespace>());  //skip empty lines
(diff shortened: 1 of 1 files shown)


include/seqan/gff_io/gff_io_base.h
```diff
@@ -436,6 +437,14 @@ void readRecord(GffRecord & record, CharString & buffer, TFwdIterator & iter)
             break;
         }
     }
+
+    // The last line might be a "### directive" specifically in GFF3
+    // Need to skip it to avoid another call of readRecords
+    skipUntil(iter, NotFunctor<IsWhitespace>());  //skip empty lines
+    while (!atEnd(iter) && value(iter) == '#')
```
rrahn: I am not fully aware of the specifications, but lines starting with `#` can always be ignored?

temehi: Yes they can be ignored safely.

rrahn: Good then :smile:

temehi:
@rrahn  🔔 ping!

PR #2118, opened 2017-09-27
Parallel DP

rrahn:
## Changes

- implements the basic wavefront dp 
- improves vectorisation of dp
- adds some data structures for the parallel module
- introduces execution policy
- supersedes #2106 
## TODO

- [x] API Documentation
- [ ] DP Manual
- [x] Increase test coverage
- [x] Add more sanity checks
- [x] Add localAlignmentScore for banded case.
- [x] Add automatic simd testing for different instruction sets.
- [x] Adapt copyright year

PR #2127, opened 2017-10-10
SeqAnSimdUtility: Use feature detection instead of compiler version checking

### LINKED ISSUE #2065: compiler crash with g++-5 and g++-6
The target test_align_simd_sse4 make the compilers crash: http://cdash.seqan.de/viewBuildError.php?buildid=222160

EDIT: cdash points to this build https://travis-ci.org/seqan/seqan/builds/228898434

marehr:
new:
* execute tests on host if the simd extension is supported by the cpu
* offer compiler flags as string and as list

fix:
* Correctly propagate include libraries to test targets for simd unit tests.

better diagnostics:
* add more compiler diagnostic

rewrote compiler detection:
* try_compile if seqans simd implementation is there
* first test seqan-simd backend
* detection will be executed once
* abort after first compiling failure
* simplify simd detection
* use flags for try and compile
* empty compiler flags will skip
* group clang blacklist rules together
* group licc blacklist rules together

use bug/crash detection for (seqan|ume)simd_compile_blacklist instead of version checking:
* detect icc16.0.{0,1} bug
* detect icc <17.0.2 bug
* detect clang4.0.0 bug
* detect missing intrinsics gcc 4.9
* gcc4.9 abi issue (avx2)
* gcc4.9 blacklist is covered by try-compile
* fix gcc4.9 issue (std::size) and new warning for clang to check binaries

other:
* dry redundant addition of testcases
* dry clone_target
* dry last static simd into loop
* tested source code should not result in a noop-binary
* doc improvement

----

Probably fixes #2065 (I forgot)


rrahn:
Ok, great. I think we can safely assume that the travis error was unrelated, but I will restart the job.
I will also have a look at the windows builds.

marehr:
Thank you!

rrahn:
The windows error is kind of annoying, as it is due to broken channel connection to the VM.
I will look into it next week, so we get it stable for CI and Nightlies.
Please check if it works on windows after I merged this.

h-2:
Are any of these new CMake files required for SeqAn as a library? If yes, are they also installed by the build system?
Please be aware that our cmake folder is **not shipped** to our users since quite a while now.

h-2:
Shouldn't `FindUmeSIMD.cmake` be provided by UmeSIMD? Did you try upstreaming this?

@marehr @rrahn 

(I am not trying to be annoying here, but if we don't ship the right things with the release, no-one will be able to use all that fancy parallelisation stuff)

marehr:
> Are any of these new CMake files required for SeqAn as a library? If yes, are they also installed by the build system?

There is no definite answer to that. If you add `-march=native` and your code compiles, than everything works out-of-the-box. SeqAnSimdUtility primarily mitigates known issues with compilers and linkers and try-compiles them or blacklists them by compiler version for unknown systems.

The use case as I see it: You want to build an app on a specific parallel computer and you want to execute it there. So you will compile your library with `-march=native` and the library will auto-select the correct available simd registers. Now either the code compiles or not. If not you will have to figure out why, the reason will most likely be a compiler bug. But, this person most likely does not want his build be suppressed or disabled or whatever, because he want his AVX512 generated code.

The binaries that are provided by us are known to work, because we know that our compilers produce correct binaries and work. And general compiled binaries will always work, because they don't us simd.

But, we could in general give out a warning to self-compiling users that simd-accelerated code needs recent compilers (for old compiler versions) to work and that even recent compilers can produce wrong code (as you can see for some clang 4.0 bugs).

But even then, they should always run their test cases to know if the produced results are correct or not (simd is so fragile and finicky), because there are some other nasty bugs where the code generation produces invalid or incomplete code. Those are currently detected by the compiler version for our test cases, but that's not reliable in general.

On the other side: SeqAnSimdUtility is (nearly) self-contained and could be used by our users to profit from our knowledge. We could provide this script for the user to automatically add and detect platforms for their test cases (sse4, avx2, avx512, ...).

----

TL,DR; For test cases we could provide this utility; For apps it makes no sense.

> Shouldn't `FindUmeSIMD.cmake` be provided by UmeSIMD? Did you try upstreaming this?
Was always on my list, but hadn't time to. (I will create a pull request)

h-2:
I have discussed this with @rrahn also. I was under the impression that all our SIMD stuff needs UME::SIMD, but since sse4 and avx2 will work without it, too, it's no problem. 

> Was always on my list, but hadn't time to. (I will create a pull request)

Yes, this would be good in any case, but it is not urgent.

mr-c:
So this is all at compile time?

What about Debian, where we compile a single binary for all AMD64 type processors? If the build machine is very recent will that result in un-executable code for older processors?

Might need to introduce a fat binary option.. https://twitter.com/biocrusoe/status/872827965712814080

marehr:
> What about Debian, where we compile a single binary for all AMD64 type processors? If the build machine is very recent will that result in un-executable code for older processors?

@mr-c No, you would have to compile our apps with `-march=native`, `-msse4`, `-mavx2`, etc. and since you don't do that (and we don't do that by default), it should absolutely have no impact on older systems.

But saying that, if you compile our library tests, it will also compile and test our SIMD implementation for sse4, avx2 and so on, but only if the compiler AND the cpu of the build system supports those extensions. This will appear in the build log, but again NO apps will be build with flags that are not supported by all AMD64 type processors. This only indicates that the building system is capable of using our SIMD features.

Since all this SIMD related stuff is so fragile and I experienced so many weird compiler bugs/crashes/liker problems, I'm eager to know if your test matrix will run through with these changes. If not we have to make changes, if it's unbearable/unmanageable we could add a kill-switch for the test-cases, because they have absolutely no impact on our apps. 

Could I clarify your concerns?

mr-c:
@h-2 Thanks for the clarification, that was helpful.

It would be nice if the seqan-apps were compiled in a way to allow Debian users with compliant CPUs to take advantage of the extensive SIMD work.

Perhaps https://github.com/clearlinux/make-fmv-patch could be of assistance here..

marehr:
@mr-c We want to do something like that, I think we wanted to use:
https://gcc.gnu.org/onlinedocs/gcc-6.4.0/gcc/Function-Multiversioning.html

But, first we want to implement and test SIMD stuff more extensively, before defaulting it.

But, we have one pending pr #2118 that will bring us SIMD accelerated global/local alignment. :tada:
And after that we definitely want our users to profit from it.

PR #2130, opened 2017-10-12
Fix small issue in SeqAnSimdUtility

marehr:
* SEQAN_SIMD might not be detected, because of some left over platform flags (e.g. `-mavx2`, `-mavx512f`) from a previous run.
* Force compiler to use avx512 instructions in try-compile

### DIFF
--- a/util/cmake/SeqAnSimdUtility.cmake
+++ b/util/cmake/SeqAnSimdUtility.cmake
@@ -185,18 +185,22 @@ set(SEQAN_SIMD_AVX512_KNL_SOURCE
 "#include <cstdint>
 #include <immintrin.h>
 #include <iostream>
+#include <random>
 
 int main() {
+  std::random_device r;
+  std::default_random_engine e(r());
+  std::uniform_int_distribution<int64_t> d(1, 10);
+
   alignas(64) uint64_t s[]{9,9,9,9,9,9,9,9};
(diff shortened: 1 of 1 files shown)


util/cmake/SeqAnSimdUtility.cmake
```diff
@@ -185,18 +185,22 @@ set(SEQAN_SIMD_AVX512_KNL_SOURCE
 "#include <cstdint>
 #include <immintrin.h>
 #include <iostream>
+#include <random>
 
 int main() {
+  std::random_device r;
```
rrahn: I strongly advice against using random stuff in tests and build stuff. 
Why does it need to be randomly generated?

marehr: Uhm, this produces a binary that tests if some errors occur when using sse4/avx2/avx512. If I don't include some randomness, the compiler will optimize everything away and tests nothing.

h-2: use the volatile keyword to prevent a variable from being optimised away!

marehr: did not work

rrahn: Ok I understand the problem now. 
Let's keep it this way

PR #2131, opened 2017-10-12
 [FIX] always include pthread with whole-archive

h-2:
supersedes #2006 
(this time against develop)

### DIFF
--- a/tests/rna_io/CMakeLists.txt
+++ b/tests/rna_io/CMakeLists.txt
@@ -15,7 +15,9 @@ message (STATUS "Configuring tests/rna_io")
 # ----------------------------------------------------------------------------
 
 # Search SeqAn and select dependencies.
-find_package (SeqAn REQUIRED)
+if (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
+    find_package (SeqAn REQUIRED)
+endif ()
 
 # ----------------------------------------------------------------------------
 # Build Setup
--- a/tests/simd/CMakeLists.txt
+++ b/tests/simd/CMakeLists.txt
@@ -15,7 +15,10 @@ message (STATUS "Configuring tests/simd")
 # ----------------------------------------------------------------------------
 
 # Search SeqAn and select dependencies.
-find_package (SeqAn REQUIRED)
+if (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
+    find_package (SeqAn REQUIRED)
+endif ()
+
 include (SeqAnSimdUtility)
 
 # ----------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


h-2:
@rrahn travis and osx tests now succeed!

tests/simd/CMakeLists.txt
```diff
@@ -15,7 +15,10 @@ message (STATUS "Configuring tests/simd")
 # ----------------------------------------------------------------------------
 
 # Search SeqAn and select dependencies.
-find_package (SeqAn REQUIRED)
+if (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
```
rrahn: When is the SIMD test not build under `DEVELOP`

tests/rna_io/CMakeLists.txt
```diff
@@ -15,7 +15,9 @@ message (STATUS "Configuring tests/rna_io")
 # ----------------------------------------------------------------------------
 
 # Search SeqAn and select dependencies.
-find_package (SeqAn REQUIRED)
+if (NOT "${SEQAN_BUILD_SYSTEM}" STREQUAL "DEVELOP")
```
rrahn: why is this only for these two tests necessary?
And when are tests not build in `DEVELOP` mode?

util/cmake/seqan-config.cmake
```diff
@@ -298,12 +298,9 @@ if ((${CMAKE_SYSTEM_NAME} STREQUAL "Linux") OR (${CMAKE_SYSTEM_NAME} STREQUAL "k
   set (SEQAN_LIBRARIES ${SEQAN_LIBRARIES} rt)
 endif ()
 
+# some OSes don't link pthread fully when building statically so we explicitly include whole archive
 if (UNIX AND NOT APPLE)
-  if ((CMAKE_CXX_FLAGS MATCHES "-static") OR (SEQAN_CXX_FLAGS MATCHES "-static") OR (CMAKE_EXE_LINKER_FLAGS MATCHES "-static"))
     set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--whole-archive -lpthread -Wl,--no-whole-archive")
-  else ()
-    set (SEQAN_LIBRARIES ${SEQAN_LIBRARIES} pthread)
-  endif ()
```
rrahn: so we always set whole archive independent of static flag, meaning we always have it statically linked also in `DEVELOP` mode?

tests/rna_io/CMakeLists.txt
h-2: > why is this only for these two tests necessary?

It's necessary for all tests, but these two deviated from the rest by doing it differently.

> And when are tests not build in DEVELOP mode?

Normally not at all, but the important thing is that it is DEVELOP mode that makes the dependencies be detected first. I guess we could remove these checks from the tests altogether, but that would be a different PR. This PR brings the two tests in line with the rest.

tests/simd/CMakeLists.txt
h-2: It is built, we are just making it detect dependencies only once.

util/cmake/seqan-config.cmake
h-2: No, whether or not it is "whole" is independent of static/dynamic. Yes, we are always making it "whole" now, because the way of detecting static is broken, but it doesn't imply that it is always static. In fact when linking dynamically it doesn't make a difference so it only applies to static linking really.

rrahn: ok then

PR #2142, opened 2017-10-16
fix issues on non-intel/amd platforms

h-2:
WIP, do not merge

### DIFF
--- a/apps/yara/tests/run_tests.py
+++ b/apps/yara/tests/run_tests.py
@@ -33,6 +33,12 @@ sam_transforms = [
 def main(source_base, binary_base):
     """Main entry point of the script."""
 
+    # gold standard binary files created on little endian
+    if sys.byteorder != 'little':
+        print 'Skipping tests for Yara on big endian'
+        print '====================================='
+        return 0
+
     print 'Executing test for Yara'
     print '=============================='
     print
--- a/include/seqan/bam_io/bam_alignment_record.h
+++ b/include/seqan/bam_io/bam_alignment_record.h
@@ -117,12 +117,12 @@ struct BamTypeChar
     {
         VALUE =
             (IsSameType<TValue, char>::VALUE)?              'A':
-            (IsSameType<TValue, signed char>::VALUE)?       'c':
-            (IsSameType<TValue, unsigned char>::VALUE)?     'C':
-            (IsSameType<TValue, short>::VALUE)?             's':
-            (IsSameType<TValue, unsigned short>::VALUE)?    'S':
-            (IsSameType<TValue, int>::VALUE)?               'i':
-            (IsSameType<TValue, unsigned int>::VALUE)?      'I':
+            (IsSameType<TValue, int8_t>::VALUE)?            'c':
+            (IsSameType<TValue, uint8_t>::VALUE)?           'C':
+            (IsSameType<TValue, int16_t>::VALUE)?           's':
+            (IsSameType<TValue, uint16_t>::VALUE)?          'S':
+            (IsSameType<TValue, int32_t>::VALUE)?           'i':
+            (IsSameType<TValue, uint32_t>::VALUE)?          'I':
             (IsSameType<TValue, float>::VALUE)?             'f':
 //          (IsSameType<TValue, double>::VALUE)?            'd':
             (IsSequence<TValue>::VALUE)?                    'Z':
@@ -131,14 +131,14 @@ struct BamTypeChar
 };
 
 // List of primitive BAM types (ordered by expected usage frequency)
-typedef TagList<int,
-        TagList<unsigned int,
+typedef TagList<int32_t,
+        TagList<uint32_t,
         TagList<float,
-        TagList<short,
-        TagList<unsigned short,
+        TagList<int16_t,
+        TagList<uint16_t,
         TagList<char,
-        TagList<unsigned char,
-        TagList<signed char
+        TagList<uint8_t,
+        TagList<int8_t
 //      TagList<double
         > > > > > > > > BamTagTypes;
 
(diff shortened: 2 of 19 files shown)


apps/yara/tests/run_tests.py
```diff
@@ -72,9 +79,9 @@ def main(source_base, binary_base):
         conf = app_tests.TestConf(
             program=path_to_indexer,
             args=[ph.inFile('input/%s-genome.fa' % organism),
-                  '-o', ph.outFile('%s-genome' % organism)],
+                '-o', ph.outFile('%s-genome' % organism)],
             to_diff=[(ph.inFile('gold/%s-genome.%s' % (organism, ext)),
-                     ph.outFile('%s-genome.%s' % (organism, ext)), 'md5') for ext in exts])
+                      ph.outFile('%s-genome.%s' % (organism, ext)), 'md5') for ext in exts])
```
rrahn: superfluous white space correction

include/seqan/bam_io/bam_alignment_record.h
```diff
@@ -607,6 +607,37 @@ getAlignmentLengthInRef(BamAlignmentRecord const & record)
     return l;
 }
 
+// ----------------------------------------------------------------------------
+// Function appendRawPod()
+// ----------------------------------------------------------------------------
+
+#if SEQAN_BIG_ENDIAN
+inline void
+ensure_little_endian(BamAlignmentRecordCore & r)
```
rrahn: Really? Please don't start mixing CamelCase and Snake case in SeqAn 2.x

rrahn: I also suggest calling it `makeLittleEndian` to be more STLish.

include/seqan/bam_io/bam_alignment_record.h
```diff
@@ -607,6 +607,37 @@ getAlignmentLengthInRef(BamAlignmentRecord const & record)
...
+{
+    ensure_little_endian(r.rID);
+    ensure_little_endian(r.beginPos);
+    // _l_qname unchanged because 8bit
+    // mapQ unchanged because 8bit
+    r.bin       = htole16(r.bin);
```
rrahn: That is kind of weird to call the specific system function here.
I know that these are bit fields pointing to `uint32_t` types, but maybe you could add a internal function that fixes the byte size and let `makeLittleEndian` in the default implementation call the internals.
```cpp
namespace detail
{
template <typename TInt>
auto makeLittleEndian(TInt const value, ByteSize<4> const &)
{
    return htole32(value);
}
template <typename TInt>
auto makeLittleEndian(TInt const value, ByteSize<2> const &)
{
    return htole16(value);
}
...

template <typename T>
auto makeLittleEndian(T const value)
{
    return makeLittleEndian(value, ByteSize<sizeof(T)>{});
}
}
```

You can also keep passing it by reference.

include/seqan/platform.h
```diff
@@ -462,4 +464,77 @@ typedef int8_t __int8;     // nolint
...
+    uint64_t tmp = htole64(*reinterpret_cast<uint64_t*>(&in));
+    char *out = reinterpret_cast<char*>(&in);
+    *out = *reinterpret_cast<char*>(&tmp);
+}
+#endif // SEQAN_BIG_ENDIAN
+
+} // namespace seqan
+
+// DEFAULT PAGESIZE FOR MMAP
+#ifndef SEQAN_DEFAULT_PAGESIZE
```
rrahn: Can you please document why 64Kbyte?

tests/bam_io/test_bam_sam_conversion.h
```diff
@@ -132,7 +132,9 @@ SEQAN_DEFINE_TEST(test_assign_tags_bam_to_sam_type_f)
     arrayCopy(DATA, DATA + 7, &bamTags[0]);
     CharString samTags;
     assignTagsBamToSam(samTags, bamTags);
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386) || defined(_M_IX86) // floaing point madness
```
rrahn: please give a more specific debug information.

include/seqan/bam_io/bam_alignment_record.h
h-2: > Really? Please don't start mixing CamelCase and Snake case in SeqAn 2.x

Wow, I totally didn't see this.

include/seqan/bam_io/bam_alignment_record.h
h-2: I don't understand what you mean, maybe we can discuss this?

include/seqan/platform.h
h-2: It's a natural constant encoded by the universe like π 😉

tests/bam_io/test_bam_sam_conversion.h
h-2: ok

include/seqan/platform.h
h-2: I have added a comment just below you remark!

include/seqan/bam_io/bam_alignment_record.h
h-2: This won't work, because you cannot take the address of bitfields. There is no way around manually reassigning this here. One could use `enforceLittleEndian` instead of `htole16` like this:
```cpp
auto tmp1 = r.bin;
enforceLittleEndian(tmp1);
r.bin = tmp1;
```
But it would very likely prevent optimisations here. `htole16` is actually standardised so it may not be pretty, but I think it is ok to use it in this one scenario.

h-2:
@rrahn can you re-review?

PR #2150, opened 2017-11-01
[FIX] ambiguous calls to std::empty/seqan::empty to allow building with -std=c++1z/-std=c++17

marehr:


h-2:
Looks good, but let's wait for travis!
Could you (ina separate PR) replace the gcc-6 entry in travis by a gcc-7 entry that includes `-std=c++17`? Because right now the changes of this PR are not checked...

Thanks!

marehr:
I tried building our test cases with `-std=c++17` locally and I had some more issues, so this will be the first patch for a series of other fixes.

Primarily those did not run:
```
	 73 - test_test_journaled_string_tree (Not Run)
	 75 - test_journaled_string_tree_traverser (Not Run)
	 76 - test_journaled_string_tree_find (Not Run)
```

h-2:
ok, do you want me to go ahead and merge this or will you push other commits onto this branch?

marehr:
I would prefer the merge, because the stuff I'm currently working in seqan3 on is depending on this.

I just wanted to give you a heads on that its not enabling c++17 for the whole library, but for most parts.

PR #2154, opened 2017-11-02
compiler warnings on gcc7

### LINKED ISSUE #2143: compiler warnings on gcc7
See cdash.

marehr:
Resolves #2143.

### DIFF
--- a/include/seqan/basic/allocator_interface.h
+++ b/include/seqan/basic/allocator_interface.h
@@ -220,44 +220,6 @@ allocate(T const &,
 #endif
 }
 
-template <typename T, typename TValue, typename TSize, typename TUsage>
-inline void
-allocate(T &,
-         TValue * & data,
-         TSize count,
-         Tag<TUsage> const &)
-{
-//  data = (TValue *) operator new(count * sizeof(TValue));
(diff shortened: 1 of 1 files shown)


h-2:
Aren't the two modified functions exactly the same? If yes, can you merge them into one before adding new code complexity twice?

marehr:
I didn't add those functions, they differ in the signature. One being const and another being non-const. 

Am 3. November 2017 16:45:29 MEZ schrieb Hannes Hauswedell <notifications@github.com>:
>Aren't the two modified functions exactly the same? If yes, can you
>merge them into one before adding new code complexity twice?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/seqan/seqan/pull/2154#issuecomment-341742180

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

h-2:
> I didn't add those functions

Yeah, I know, but as they get fatter it is good to evaluate if both overloads are necessary. I think removing the non-const one should just work, but maybe you have to re-wire one signature to the other...

PR #2155, opened 2017-11-02
Disable UME::SIMD for alignments

marehr:
Our alignment code does not build with UME::SIMD right now as it is.

But since it compiles with the new simd-alignment module rene is currently working on, I simply deactivated the code path for UME::SIMD, so that the test-cases won't fail.


h-2:
Hm, In don't understand it. Currently none of the tests fails in the nightlies so what's the problem?

marehr:
We don't test ume::simd in our nightlies 😅 if we would, you would see a failing build. 

Am 3. November 2017 16:43:04 MEZ schrieb Hannes Hauswedell <notifications@github.com>:
>Hm, In don't understand it. Currently none of the tests fails in the
>nightlies so what's the problem?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/seqan/seqan/pull/2155#issuecomment-341741447

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #2158, opened 2017-11-10
Add seqan::min function for simd vectors

marehr:
Add seqan::min function for simd vectors.

### DIFF
--- a/tests/simd/test_simd_vector.h
+++ b/tests/simd/test_simd_vector.h
@@ -193,7 +193,7 @@ SEQAN_DEFINE_TEST(test_simd_types)
     static_assert(LENGTH<SimdVector<uint32_t, 4>::Type>::VALUE == 4, "128bit register fits 4 int's");
     SimdVector<uint32_t, 4>::Type a128 = {0, 1, 2, 3};
     for (uint32_t i = 0; i < 4; ++i) {
-        // std::cout << i << ": " << a128[i] << " = " << i << std::endl;
+        // std::cout << "DEBUG: " << i << ": " << a128[i] << " = " << i << std::endl;
         SEQAN_ASSERT_EQ(a128[i], i);
     }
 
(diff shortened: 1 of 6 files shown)


tests/simd/test_simd_vector.h
```diff
@@ -441,6 +441,27 @@ SEQAN_TYPED_TEST(SimdVectorTestCommon, Max)
...
+
+    TSimdVector a{0u}, b{0u};
+    fillVectors(a, b);
+
+    auto c = min(a, b);
+
+    for (auto i = 0; i < length; ++i)
+    {
+        TValue a_i = -3 + i * 3, b_i = length - i;
+        // std::cout << i << " / " << length << ": " << (int)c[i] << " = min (" << (int)a[i] << ", " << (int)b[i] << ")" << std::endl;
```
h-2: can this line be removed?

marehr: I have this line in all test functions.

I can remove them in all those functions, but it was quite handy to have them to comment them out, if something was wrong. It is only a test case.

h-2: Hm, but it is not clear for someone who doesn't know the code. Commented out code can mean anything: "this broke and I couldn't fix it so I just removed it", "this is debug and has no effect", "I forgot to re-add this and it is necessary" ....
So I would remove it in 99% of the test cases. If you have very complex code that is super-hard to debug, and you wrote some debugging functionality, you could leave it inside, but then I would instead:
```cpp
#if 0 // debug(marehr): The following code block does ....
    foobar();
    bax();
    i + 3;
#endif
```
So it is absolutely clear that it was intentionally kept, by whom, and why.

marehr: @h-2, a `std::cout`, which does not change any state and is in a test case, is clearly for debugging.
I read your concern and I agree with you in general, but in this case it adds zero complexity, but is really handy to have when something fails.

If this is your only remark, could please merge it?

rrahn: I suggest meet half way and simply add a DEBUG info at the beginning of the std::cout output: 
like so: `// std::cout << "DEBUG:" << ... << std::endl;`

PR #2163, opened 2017-11-15
Develop

gurgese:
modifyed the name of the function called in the first example for finding substrings from " findAll" to "find"


h-2:
This only affects documentation? Why does it contain 19 commits?

gurgese:
I made a mistake choosing my branch.
I can close this and make another pull, otherwise you can directly make this correction :)

PR #2174, opened 2017-12-01
[FIX] testcase for bam_file can't build if zlib is missing

marehr:


rrahn:
Since you started this evil task. Can you please check what is broken there?

marehr:
!rebuild

rrahn:
So it seems that the compilation with msvc finally works but that there is a weird bug in the RnaRecord structure.
Can you have a look at this as well? I know ... but you opened it 😈 !!!

marehr:
!rebuild

PR #2183, opened 2017-12-13
fix issue with google test

### LINKED ISSUE #2182: seqan strings do not work with gtest
I am trying to integrate a CMake file for SeqAn and google's gtest in order to prepare for a larger project.

I am getting the following error (abbreviated) when I try to make the tests.

error: ambiguous overload for ‘operator<<’ (operand types are ‘std::ostream {aka std::basic_ostream<char>}’ and ‘const seqan::String<seqan::SimpleType<unsigned char, seqan::Dna_>, seqan::Alloc<> >’)
   *os << value;

I am trying this out on my GitHub page (https://github.com/pnrobinson/GTT), in case anybody wants to see the gory details.

I realise that seqan has its own unit testing framework, but I would love to use gtest -- is there any workaround?
thanks Peter

---

reported by @pnrobinson

marehr:
fixes #2182

### DIFF
--- a/include/seqan/sequence/string_base.h
+++ b/include/seqan/sequence/string_base.h
@@ -2034,6 +2034,30 @@ operator<<(TStream & target,
     return target;
 }
 
+// A specialization needed to avoid an ambiguous call combined with googletest.
+//
+// We use `std::basic_ostream<char>` instead of `std::ostream`, because we would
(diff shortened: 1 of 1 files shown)


h-2:
This is a bug on gtest's side and it also affects SeqAn3. It is possible to workaround it, but I am not happy about it.

PR #2185, opened 2017-12-14
[ENHANCEMENT,TEST] directory argument/option now also accept trailing whitespaces

### LINKED ISSUE #1988: INPUT_DIRECTORY and OUTPUT_DIRECTORY don't accept trailing /
The arg parser now explicitly supports directories. However, if I specify `.lambda` as directory extension, this does not include `.lambda/`.
This is unintuitive and some shells automatically add this with auto-completion.

smehringer:
Fixes #1988 

@h-2 Can you maybe try this on real use cases before merging? It seams to work fine but just to be sure since I think no app (in the seqan repo) uses this functionality.

### DIFF
--- a/include/seqan/arg_parse/arg_parse_argument.h
+++ b/include/seqan/arg_parse/arg_parse_argument.h
@@ -438,6 +438,28 @@ inline bool isOutputFileArgument(ArgParseArgument const & me)
 }
 
 // ----------------------------------------------------------------------------
+// Function isDirectoryArgument()
+// ----------------------------------------------------------------------------
+
+/*!
+ * @fn ArgParseArgument#isDirectoryArgument
+ * @headerfile <seqan/arg_parse.h>
+ * @brief Returns whether the argument is a directorz argument.
+ *
+ * @signature bool isDirectoryArgument(arg);
+ *
+ * @param[in] arg The ArgParseArgument to query.
+ *
+ * @return bool <tt>true</tt> if it is a directory argument, <tt>false</tt> otherwise.
+ */
+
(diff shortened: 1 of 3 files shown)


include/seqan/arg_parse/arg_parse_argument.h
```diff
@@ -923,7 +945,15 @@ inline void _checkValue(ArgParseArgument const & me)
 {
     unsigned i = 0;
     for (std::vector<std::string>::const_iterator it = me.value.begin(); it != me.value.end(); ++it, ++i)
-        _checkValue(me, *it, i);
+    {
+        auto val = *it;
+
+        if (isDirectoryArgument(me)) // strip trailing slash for directories
+            if (val[length(val) - 1] == '/')
+                val = val.substr(0, length(val) - 1);
```
h-2: resize instead of reassign maybe?

include/seqan/arg_parse/arg_parse_argument.h
```diff
@@ -1138,6 +1168,10 @@ inline std::string getFileExtension(ArgParseArgument const & me, unsigned pos =
     if (value.empty())
         return "";
 
+    if (isDirectoryArgument(me)) // strip trailing slash for directories
+        if (value[length(value) - 1] == '/')
+            value = value.substr(0, length(value) - 1);
```
h-2: same here

include/seqan/arg_parse/arg_parse_argument.h
smehringer: Would be more efficient. I'll try it

PR #2189, opened 2017-12-15
[FIX#2186] yara-mapper seg-faults with the intel compiler

### LINKED ISSUE #2186: yara-mapper seg-faults with the intel compiler
I could figure out that this was caused by 8be002628c4047bcbdb10c6b823a63ea0a540a49.

It seems that icpc produces wrong code for defaulted copy constructors AND `#pragma pack(push, 1)`

An explicit copy constructor works.

marehr:
fixes #2186

### DIFF
--- a/include/seqan/basic/pair_packed.h
+++ b/include/seqan/basic/pair_packed.h
@@ -78,9 +78,8 @@ struct Pair<T1, T2, Pack>
     // ------------------------------------------------------------------------
     // Members
     // ------------------------------------------------------------------------
-
-    T1 i1;
-    T2 i2;
+    T1 i1{};
+    T2 i2{};
 
     // ------------------------------------------------------------------------
     // Constructors
@@ -88,8 +87,10 @@ struct Pair<T1, T2, Pack>
 
     // Pair() = default; does not work on gcc4.9, it issues warnings if T1/T2
     // have no proper default constructor. >=gcc5.0 reports no warnings.
+    // Caused by yara_indexer build, demo_tutorial_indices_base and
+    // demo_tutorial_index_iterators_index_bidirectional_search.
 #if defined(COMPILER_GCC) && (__GNUC__ <= 4)
-    Pair() : i1(), i2() {};
+    Pair() : i1(T1()), i2(T2()) {};
 #else
     Pair() = default;
 #endif
--- a/include/seqan/basic/pair_packed.h
+++ b/include/seqan/basic/pair_packed.h
@@ -88,8 +88,20 @@ struct Pair<T1, T2, Pack>
 
     // Pair() = default; does not work on gcc4.9, it issues warnings if T1/T2
     // have no proper default constructor. >=gcc5.0 reports no warnings.
+#if defined(COMPILER_GCC) && (__GNUC__ <= 4)
     Pair() : i1(), i2() {};
+#else
+    Pair() = default;
+#endif
+
+    // NOTE(marehr) intel compiler bug in 17.x and 18.x: defaulted copy-constructor
+    // in classes with `#pragma pack(push, 1)` seg-faults. This leads to a
+    // seg-fault in yara-mapper (app test case yara).
+#if defined(COMPILER_LINTEL) || defined(COMPILER_WINTEL)
+    Pair(Pair const & p) : i1(p.i1), i2(p.i2) {};
+#else
     Pair(Pair const &) = default;
+#endif
     Pair(Pair &&) = default;
     ~Pair() = default;
     Pair & operator=(Pair const &) = default;


include/seqan/basic/pair_packed.h
```diff
@@ -88,8 +88,20 @@ struct Pair<T1, T2, Pack>
 
     // Pair() = default; does not work on gcc4.9, it issues warnings if T1/T2
     // have no proper default constructor. >=gcc5.0 reports no warnings.
+#if defined(COMPILER_GCC) && (__GNUC__ <= 4)
     Pair() : i1(), i2() {};
+#else
+    Pair() = default;
+#endif
```
h-2: These two mean different things. In one case the members will always be default-initialised, in the other case they are not. This has a potentially large impact on performance and/or correctness of code.
I see that this also changed in 8be002628c4047bcbdb10c6b823a63ea0a540a49 ...
Which of the behaviours is intended?

include/seqan/basic/pair_packed.h
```diff
@@ -88,8 +88,20 @@ struct Pair<T1, T2, Pack>
...
+
+    // NOTE(marehr) intel compiler bug in 17.x and 18.x: defaulted copy-constructor
+    // in classes with `#pragma pack(push, 1)` seg-faults. This leads to a
+    // seg-fault in yara-mapper (app test case yara).
+#if defined(COMPILER_LINTEL) || defined(COMPILER_WINTEL)
```
h-2: Have you checked that the move constructor isn't also affected? Sounds kinda weird...

include/seqan/basic/pair_packed.h
marehr: Good catch, forgot to re-add the default initialization.

The original behaviour was
```c++
T1 i1;
T2 i2;
Pair() : i1(), i2() {};
```
After that I changed it to
```c++
T1 i1{};
T2 i2{};
Pair() = default;
```
That triggered warnings on gcc.

Thus changing it back to

```c++
T1 i1;
T2 i2;
Pair() : i1(), i2() {};
```

Now I thought that it might be better, if we ever drop a compiler, to make it explicit for newcomers.


```c++
T1 i1{};
T2 i2{};
#if ...
Pair() : i1(), i2() {};
#else
Pair() = default;
#endif
```

I might be wrong, but this code should not change the original behaviour, right? And should not have any performance impacts, right? 

Thank you!

h-2:
```c++
T1 i1{};
T2 i2{};
#if ...
Pair() : i1(), i2() {};
#else
Pair() = default;
#endif
```

is the same as 

```c++
T1 i1{};
T2 i2{};
Pair() = default;
```

for `#else` which is the normal case which will result in warning, or not? Did you try just using the member initialisers and removing all constructors so this is a simple aggregate type?

include/seqan/basic/pair_packed.h
marehr: I'm currently trying to construct a minimal test case. To figure that out. Fact is that we don't use the move constructor.

include/seqan/basic/pair_packed.h
marehr: I removed the other copy constructors, but some other code depended on them.

h-2: > I removed the other copy constructors, but some other code depended on them.

You need to define all, or remove all. If all are removed, all are created by the compiler.

marehr: I know, I did mean this copy-constructor:

```c++
     template <typename T1_, typename T2_, typename TSpec__>
     // TODO(holtgrew): explicit?
     Pair(Pair<T1_, T2_, TSpec__> const &_p) :
         i1(getValueI1(_p)), i2(getValueI2(_p))
     {}

Comment (marehr) [reply include/seqan/basic/pair_packed.h]: Here are some example errors from gcc-4.9:

```c++
seqan-src/include/seqan/basic/pair_packed.h:94:5: warning: missing initializer for member ‘seqan::Tuple<seqan::SimpleType<unsigned char, seqan::Dna_>, 7u, seqan::BitPacked<> >::i’ [-Wmissing-field-initializers]
     Pair() = default;
     ^

seqan-src/include/seqan/pipe/pipe_sampler.h:507:27: note: synthesized method ‘seqan::Pair<T1, T2, seqan::Tag<seqan::Pack_> >::Pair() [with T1 = seqan::Pair<unsigned char, unsigned int, seqan::Tag<seqan::Pack_> >; T2 = seqan::Tuple<seqan::SimpleType<unsigned char, seqan::Dna_>, 7u, seqan::BitPacked<> >]’ first required here 
             limits(_limits) {}
                           ^

seqan-src/include/seqan/basic/pair_packed.h:94:5: warning: missing initializer for member ‘seqan::Tuple<long unsigned int, 7u, seqan::Tag<seqan::Pack_> >::i’ [-Wmissing-field-initializers]
     Pair() = default;
     ^

seqan-src/include/seqan/pipe/pipe_sampler.h:118:25: note: synthesized method ‘seqan::Pair<T1, T2, seqan::Tag<seqan::Pack_> >::Pair() [with T1 = long unsigned int; T2 = seqan::Tuple<long unsigned int, 7u, seqan::Tag<seqan::Pack_> >]’ first required here 
             tmpRef(&tmp2) {}
                         ^

```

marehr: Funny enough, if I use bracket `{}` initialization, I get the same warning. But then using `()` I get no error.

```c++
seqan-src/include/seqan/basic/pair_packed.h:92:23: warning: missing initializer for member ‘seqan::Tuple<seqan::SimpleType<unsigned char, seqan::Dna_>, 7u, seqan::BitPacked<> >::i’ [-Wmissing-field-initializers]
     Pair() : i1{}, i2{} {};
                       ^
```

I thought more about it, shouldn't the current implementation without member default initialization correspond to the old version. For me it looks like, that the `{}` initialization was buggy in gcc-4.9 and that they fixed it after that.

h-2: but what about just removing all of the constructors? Let's discuss this on monday.

marehr: Yep, you are right :)

BTW without any constructor I get:

```console
seqan-src/include/seqan/pipe/pipe_base.h:857:16: error: no matching function for call to ‘seqan::Pair<unsigned char, unsigned int, seqan::Tag<seqan::Pack_> >::Pair(int, int)’
         me.pos = TPair(0, 0);
                ^

seqan-src/include/seqan/index/index_skew7_multi.h:209:78: error: no matching function for call to ‘seqan::Pair<long unsigned int, long unsigned int, seqan::Tag<seqan::Pack_> >::Pair(long unsigned int, const long unsigned int&)’
             return TResult(off[seqOfs % 7][getValueI1(x_i1)] - seqOfs/7, x.i2);
                                                                              ^
```

h-2: Yes, if you make it an aggregate type, you would need to do `TPair{0,0}`. But requiring that would be a breaking change so let's not do it.

marehr:
!rebuild

include/seqan/basic/pair_packed.h
marehr: I tried 3 days, but I don't get it completely reduced.

Issue #2194, opened 2017-12-18
windows builds broken on nightlies


h-2:
also due to #2142

PR #2195, opened 2017-12-18
Fix/revert breaking deprecation

h-2:
These changes as part of the deprecation PR were actually breaking and not easy to fix so I propse reverting them.

### DIFF
--- a/apps/searchjoin/db.h
+++ b/apps/searchjoin/db.h
@@ -130,25 +130,25 @@ namespace seqan
 template <>
 struct Fibre<TDbDnaSaSmall, FibreSA>
 {
-    typedef String<Pair<unsigned int, unsigned char, BitPacked<24, 8> >, StringSpec<TDbDnaSaSmall>::Type> Type;
+    typedef String<Pair<unsigned int, unsigned char, BitPacked<24, 8> >, DefaultIndexStringSpec<TDbDnaSaSmall>::Type> Type;
 };
 
 template <>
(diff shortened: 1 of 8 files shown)


h-2:
requirement for seqan/lambda#109

apps/searchjoin/db.h
```diff
@@ -130,25 +130,25 @@ namespace seqan
 template <>
 struct Fibre<TDbDnaSaSmall, FibreSA>
 {
-    typedef String<Pair<unsigned int, unsigned char, BitPacked<24, 8> >, StringSpec<TDbDnaSaSmall>::Type> Type;
+    typedef String<Pair<unsigned int, unsigned char, BitPacked<24, 8> >, DefaultIndexStringSpec<TDbDnaSaSmall>::Type> Type;
```
rrahn: `DefaultIndexStringSpec` was previously defined?

h-2: Yes, it was previously defined and marked deprecated in by @smehringer's PR. However since she also replaced the internal use of the function it also meant that downstream users who overloaded this metafunction no longer actually influenced the code in SeqAn. This made it a breaking change.
Additionally, due to definition/instantiation/specialization times of the templates it is not possible to just use `StringSpec` instead.

Since I couldn't fix it in a reasonable amount of time in my app, I concluded most of our users wouldn't be able to either.

rrahn: ok

rrahn:
windows builds are fixed by #2194

PR #2198, opened 2017-12-20
[FIX] Fixes CMP0054 warning in travis builds.

rrahn:


h-2:
Shouldn't we set this globally instead of just on travis? It would fix the issues on FreeBSD nightlies, too & likely prevent the issue from appearing should Debian update to cmake-3.9 (before updating to 3.10).

rrahn:
No, this is not our fault but we need to set the policy. But you are right it is better to set it globally.

rrahn:
Ok from MPOV this can be merged.

Issue #2193, opened 2017-12-18
update platform support for 2.4.0


h-2:
* [x] drop icpc-16 on all platforms due to #2142 breakage and bad c++14 support in general
  * [x] update icpc support after 17 and 18 were added to nightlies, see #2199 
  * [x] drop gcc-4.9 on on mac
  * [x] update MSVC support info after adding VS2017, see also #2200

marehr:
So that means we drop gcc 4.9?

rrahn:
On mac, only

marehr:
😞

h-2:
regarding icpc16 this needs to be changed still:
https://github.com/seqan/seqan/blob/develop/util/cmake/SeqAnBuildSystem.cmake#L551
And the whole WINTEL thing needs testing!!

marehr:
CMake seems to support WINTEL redistribution since half a year

https://github.com/Kitware/CMake/blob/2868019b2f93f2eb50de901d8a58ceb9b5790ae5/Modules/InstallRequiredSystemLibraries.cmake#L62-L90

> And the whole WINTEL thing needs testing!!
We test them nightly, what do you mean?

h-2:
> CMake seems to support WINTEL redistribution since half a year

Looks good, can we make use of that?

> We test them nightly, what do you mean?

We test 16 (which we don't support anymore), we don't test 17 and 18.

PR #2206, opened 2018-01-03
drop icpc 16 support in seqan2.4

marehr:
One task of #2193


h-2:
I think you forgot this thing:
https://github.com/seqan/seqan/blob/develop/util/cmake/SeqAnBuildSystem.cmake#L551

PR #2209, opened 2018-01-05
[INTERNAL] Fixes unused attribute and adds c++17 flag for gcc on travis

rrahn:


include/seqan/platform.h
```diff
@@ -401,7 +401,11 @@ typedef int8_t __int8;     // nolint
 
 // A macro to eliminate warnings on GCC and Clang
 #if defined(COMPILER_GCC) || defined(COMPILER_CLANG) || defined(COMPILER_LINTEL)
-#define SEQAN_UNUSED __attribute__((unused))
```
marehr: This might be a problem, because you can place `__attribute__` at different places/positions than `[[maybe_unused]]`

util/travis/linux-cibuild.sh
```diff
@@ -18,6 +18,11 @@ if [ "$(echo ${CXX} | cut -c1-5)" = "clang" ]; then
   export CXXFLAGS="${CXXFLAGS} -Qunused-arguments -DSEQAN_IGNORE_MISSING_OPENMP=1"
 fi
 
+# compile with c++-17 if g++-7 is used.
+if [ "$(echo ${CXX})" = "g++-7" ]; then
```
marehr: that means g++-7 will ALWAYS build with -std=c++17? I would want that g++-7 also tests c++14

rrahn: Sure, but adding another build to Travis will make it more often reduce the parallel builds to sequential execution and wait times will be to long. I am preparing in parallel ci for Unix on Jenkins. As soon we have this we can revert this build here

marehr: true

include/seqan/platform.h
marehr: Just mentioning:

```c++
using type = type __attribute__(); // valid, [] invalid
using type [maybe_unused] = type; // valid

Comment (rrahn) [reply include/seqan/platform.h]: I know, but the only place that could cause this trouble was in the declaration of the exception handler, which should be fixed by this PR as well. But we have to wait for the builds to be sure.

File: include/seqan/platform.h [hunk 3]
```diff
@@ -401,7 +401,11 @@ typedef int8_t __int8;     // nolint
...
+#if __cplusplus < 201703
+#define SEQAN_UNUSED [[gnu::unused]]
+#else
+#define SEQAN_UNUSED [[maybe_unused]]
+#endif // __cplusplus < 201703
 #else
 #define SEQAN_UNUSED
 #endif
```
h-2: The if clauses are wrong. because the macro is never expanded on MSVC. You need to check for ` __cplusplus < 201703` first and then if it is not the case check for the compilers and if possible to define the gnu attribute.

util/travis/linux-cibuild.sh
h-2: In the nightlies we have g++7 with std=c++14, too, but for travis it's ok to only test one combination from my POV. (as long as c++17 is checked at all)

h-2:
BTW: Do you have fixes for the issues prepared? Otherwise all travis builds will fail after this.

rrahn:
Not yet. I’ll wait for Travis to finish and see what the main issues are. Then I submit the fixes for them

include/seqan/platform.h
rrahn: Good catch. 🔍

PR #2211, opened 2018-01-05
add SEQAN_IF_CONSTEXPR

h-2:
- [x] wait for #2209 then rebase
  - [ ] other obvious candidates where we can use this?

### DIFF
--- a/include/seqan/find/find_score.h
+++ b/include/seqan/find/find_score.h
@@ -311,7 +311,7 @@ inline void _patternInit (Pattern<TNeedle, DPSearch<TScore, TSpec, TFindBeginPat
         x += score_gap;
     }
 
-    if (IsSameType<TSpec, FindPrefix>::VALUE)
+    SEQAN_IF_CONSTEXPR (IsSameType<TSpec, FindPrefix>::VALUE)
     {//compute data_maxscore
         me.data_maxscore = 0;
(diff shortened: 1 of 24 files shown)


rrahn:
seems you broke all builds 🥇

include/seqan/find/find_score.h
```diff
@@ -375,12 +375,15 @@ _findScoreSimpleProportional(TFinder & finder, Pattern<TNeedle, DPSearch<TScore,
     TSize haystack_length = length(container(hostIterator(finder)));
 
     //limit search width for prefix search
-    if (IsSameType<TSpec, FindPrefix>::VALUE && (score_gap < 0))
+    SEQAN_IF_CONSTEXPR (IsSameType<TSpec, FindPrefix>::VALUE)
     {
-        TSize maxlen = prefix_begin_position + length(needle(me)) + ((scoreLimit(me) - me.data_maxscore) / score_gap) + 1;
-        if (haystack_length > maxlen)
+        if (score_gap < 0))
```
rrahn: Here is your error 🔍

h-2:
>  it should rather have positive effects on the performance but I strongly advice doing some micro-benchmarks. 

Well, I think there is no way this can have negative effects. It reduces code generation and branching. In the worst case it will be the same as not using it.

h-2:
rebased on current develop with c++17 CI

Issue #2166, opened 2017-11-22
[BUG] Segmentation fault on implicit cast of StringSet object to String-of-Strings object.


smehringer:
I found some strange behavior working with StringSets by accident.

The following cpp program segfaults:
```cpp
#include <seqan/sequence.h>

using namespace seqan;

void foo(String<DnaString const & v)
{
    // doesn't matter..
}

int main()
{
    StringSet<DnaString> set;
    appendValue(set, "acgtac");
    appendValue(set, "acgtac");

    foo(set);

    return 0;
}
``` 

![stacktrace](https://user-images.githubusercontent.com/7955878/33135211-b171579c-cf99-11e7-9036-e97623858553.png)
(sorry for the crappy picture but my virtual Linux machine has no internet access)

It segfaults when trying to call `getObjectId(set)` which has some weird function body ([see here](https://github.com/seqan/seqan/blob/966da62f2ba61867359cad18f4db2dbaf431243e/include/seqan/sequence/sequence_interface.h#L286)):
```cpp
    return &*end(me, Standard());
```

h-2:
I don't think it should do an implicit cast at all. Implicit casts are the devil's work 👿

rrahn:
Actually it cannot do an implicit cast as StringSet encapsulates much more stuff than a simple vector of strings. Therefor, the user should be forced to do this explicitly, which will inevitable induce a deep copy of the data. I believe no one want this (probably not even explicitly).

joergi-w:
I performed tests with the code above and I do not find the segmentation fault on both Mac and Linux, using gcc versions 4.9 / 5 / 6 / 7. Interestingly, in the foo function one can still apply functions for StringSet (e.g. lengthSum) as well as functions for String (e.g. endPosition). Because results can be unexpected (toCString for instance gives a const string containing a memory address), I agree that we should prevent a cast between String and String set.

PR #2212, opened 2018-01-08
[FIXED-#2166] String and StringSet conversion

joergi-w:


h-2:
Why did you keep the StringSet constructors as explicit and did not apply the same check as for Strings?

joergi-w:
The `StringSet` constructors do not cause all the trouble we saw with the explicit `String` constructors, so I considered the stricter as the better solution here. Should I change them to the `enable_if` version anyway?

h-2:
Yes, best be consistent.

joergi-w:
The enable_if does not make sense here, because the String argument in the StringSet constructor is already restricted to TString, which is the String Type in the StringSet. So I just removed the explicit keyword.

PR #2216, opened 2018-01-10
[INTERNAL] deactivate some apps on travis

h-2:
to bring down build and test time

### DIFF
--- a/apps/pair_align/CMakeLists.txt
+++ b/apps/pair_align/CMakeLists.txt
@@ -12,6 +12,11 @@ message (STATUS "Configuring apps/pair_align")
 
 set (SEQAN_APP_VERSION "1.3.7")
 
+if (SEQAN_TRAVIS_BUILD)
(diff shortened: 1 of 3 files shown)


apps/pair_align/CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ message (STATUS "Configuring apps/pair_align")
 
 set (SEQAN_APP_VERSION "1.3.7")
 
+if (SEQAN_TRAVIS_BUILD)
```
rrahn: Is it wise to always disable it? 
Maybe if we can detect the compiler version used, we can disable it based on the used `CXX` environment variable.
All other builds usually run through.

rrahn:
Actually the windows CI did work this time, and actually there is an error 👍

apps/pair_align/CMakeLists.txt
h-2: But bringing down the build times in general is a good idea. Also we might add new tests that then again require to deactivate something else..

PR #2219, opened 2018-01-11
[FIX] Fixes BAM jumpToRegion functionality.

### LINKED ISSUE #2213: [BUG] PR #2098 breaks BAM jumpToRegion()
I'm using `jumpToRegion` in my code for the internship and noticed that it failed on the newest develop. I did a git bisect and it seems that #2098 breaks the behavior.
It cannot "find reads in region .." anymore although there definitely are some.

@temehi can we maybe look into this together when I show you the data (GIAB NA12878)?

smehringer:
Fixes #2213 

Some explanation, because this bug seems to exist for quite a while:

Goal: Jump to the beginning of a region `[regionStart, regionEnd]`

In the current master branch (seqan 2.3) the `jumpToRegion()` jumps to highest candidate offset (candidate offset of bam index blocks) whose position is still smaller than `regionStart`. Reaching this index block, it sets the bam file stream to the first record it finds. This is most of the time not even near the actual `regionStart` and the information of `hasAlignments` was also wrong, since it has no knowledge about the actual region `[regionStart, regionEnd]`.

In the current develop branch, were the issue in master was reported by #1944, the `hasAlignment` parameter was partially fixed by PR #2098 to only be set to true: IF the very first record, that appears in the right most index block still smaller than `regionStart`, already lies inside the region. This only works if the region `[regionStart, regionEnd]` starts at the beginning of an bam index block which is often not the case.

What this PR therefore introduces, is that having found the highest candidate offset still smaller than `regionStart`, we now search with an additional while loop through the sorted records and set the bam file stream position to the first alignment overlapping the region (and `hasAlignments` to true).

A small benchmark on NA12878 chromosome 21 reveals the following read counts:
(The seqan cpp program jumps to "the start" of the region, and if `hasAlignments` is true it extracts all reads until `regionEnd` is reached, samtools was used to compare the behaviour) 
```console
    310828 samtools.sam
    314697 seqan_current_fix.sam
   2119086 seqan_master.sam
      7229 seqan_develop.sam
```

The now small difference in the current fix to samtools is that we only jump to the first read that overlaps the region, and cannot check for the following reads if the same holds. This has to be done by the user. I added a note in the documentation.

### DIFF
--- a/include/seqan/bam_io/bam_index_bai.h
+++ b/include/seqan/bam_io/bam_index_bai.h
@@ -162,24 +162,29 @@ public:
  * @fn BamFileIn#jumpToRegion
  * @brief Seek in BamFileIn using an index.
  *
- * You provide a region <tt>[pos, posEnd)</tt> on the reference <tt>refID</tt> that you want to jump to and the function
- * jumps to the first alignment in this region, if any.
+ * You provide a region <tt>[regionStart, regionEnd]</tt> on the reference <tt>refID</tt>
+ * that you want to jump to and the function jumps to the first alignment (bam record)
+ * whose start position lies inside this region, if any.
+ * Note: The current inmplementation only consideres the read start positions.
+ *       Therefore, we do guarantee that reads overlapping the region are included.
+ *       To account for this limitation you may want to choose the start of your
+ *       your region with an appropiate offset (e.g. start - length_of_read).
  *
- * @signature bool jumpToRegion(bamFileIn, hasAlignments, refID, pos, posEnd, index);
+ * @signature bool jumpToRegion(bamFileIn, hasAlignments, refID, regionStart, regionEnd, index);
  *
  * @param[in,out] bamFileIn     The @link BamFileIn @endlink to jump with.
- * @param[out]    hasAlignments A <tt>bool</tt> that is set true if the region <tt>[pos, posEnd)</tt> has any
- *                              alignments.
+ * @param[out]    hasAlignments A <tt>bool</tt> that is set true if the region <tt>[regionStart, regionEnd]</tt> has any
+ *                              at least one overlapping alignment (bam record).
  * @param[in]     refID         The reference id to jump to (<tt>int32_t</tt>).
- * @param[in]     pos           The begin of the region to jump to (<tt>int32_t</tt>).
- * @param[in]     posEnd        The end of the region to jump to (<tt>int32_t</tt>).
+ * @param[in]     regionStart   The begin of the region to jump to (<tt>int32_t</tt>).
+ * @param[in]     regionEnd     The end of the region to jump to (<tt>int32_t</tt>).
  * @param[in]     index         The @link BamIndex @endlink to use for the jumping.
  *
  * @return bool true if seeking was successful, false if not.
  *
  * @section Remarks
  *
- * This function fails if <tt>refID</tt>/<tt>pos</tt> are invalid.
+ * This function fails if <tt>refID</tt>/<tt>regionStart</tt> are invalid.
  */
 
 static inline void
@@ -202,8 +207,8 @@ inline bool
 jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
              bool & hasAlignments,
              int32_t refId,
-             int32_t pos,
-             int32_t posEnd,
+             int32_t regionStart,
+             int32_t regionEnd,
              BamIndex<Bai> const & index)
 {
     if (!isEqual(format(bamFile), Bam()))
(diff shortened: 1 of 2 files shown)


h-2:
Please don't forget to add this prominently to the changelog, because other apps might rely on the previous (wrong) behaviour!

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -298,22 +299,47 @@ jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
...
+    // alignment at the start of the offset lies inside the region.
+    // Therefore we must advance further or until the region ends.
+    setPosition(bamFile, offset);
+
+    while (!atEnd(bamFile))
+    {
+        auto seek_pos = position(bamFile); // store current pos to reset bamFile if necessary
+        readRecord(record, bamFile);
+
+        int32_t recordEndPos = record.beginPos + getAlignmentLengthInRef(record);
```
temehi: Shouldn't the endPosition be` begin + length - 1`?

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -298,22 +299,47 @@ jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
...
+                hasAlignments = true;
+                setPosition(bamFile, seek_pos);
+                break;
+            }
+        }
+        else
+        {
+            break;
+        }
+    }
```
temehi: I would remove the comment below it's redundant and has no meaning here.

temehi:
> The now small difference in the current fix to samtools is that we only jump to the first read that overlaps the region, and cannot check for the following reads if the same holds. 

I didn't quite understand this, what do you mean by the following reads?

include/seqan/bam_io/bam_index_bai.h
smehringer: True.

include/seqan/bam_io/bam_index_bai.h
smehringer: I just copied the line from before but now that you mention it.. yes you are right.

smehringer:
@temehi 
> I didn't quite understand this, what do you mean by the following reads?

It means that samtools checks each read if it actually overlaps with the region but we just ump to the start and make no further checks.
Imagine this:

```
                          regionStart                                regionEnd
--------------------------------|----------------------------------------|----------------

read1 pos 16     ===========
read2 pos 20          =============
read3 pos 21            ======
```

The `jumpToRegion()` now jumps to read2, since it is the first read overlapping regionStart. My benchmark then extracts all reads until regionEnd. So my set of reads includes read3, which samtools doesn't. I can just add a check but this is not part of the `jumpToRegion()` functionality but user specific.

temehi:
Now I understand. But isn't it a requirement to sort alignments by position before using bam_index?

temehi:
sorry disregard my previous comment

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -298,22 +299,47 @@ jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
...
+            if (recordEndPos > regionStart)
+            {
+                hasAlignments = true;
+                setPosition(bamFile, seek_pos);
+                break;
+            }
+        }
+        else
+        {
+            break;
```
rrahn: This looks kind of cumbersome.
Can you change this to:
```cpp
if (record.beginPos >= regionEnd)
    break;   // Begins after target region.

if (recordEndPos > regionStart)  // Found candidate.
{
    hasAlignments = true;
    setPosition(bamFile, seek_pos);
    break;
}
```

rrahn:
@smehringer can you also make a corresponding entry in the changelog file.
And second, I actually might miss something about our jumpToRegion design. It now just picks the first read that overlaps with the begin of the region. Now, in your benchmark, you just iterate over the reads until you find the first one, whose begin position is behind the region end? But then your benchmark does implement the find logic, right, and you can also check for read3 in your example, while doing so. 
I mean, the logic of jumpToRegion does not need to filter out the misplaced reads. That is part of your 'tool', like `samtools view ...`, right? Would that actually fix the difference in the reported hits?

smehringer:
@rrahn Please do a final review.

(Just for completion of this issue) As decided in the discussion earlier, we will limit the `jumpToRegion()` functionality to merely use the records start positions. A call of `jumpToRegion(bamFile, hasAlignments, refId, start, end, bamIndex)` will now set the `bamFile` stream iterator to the first alignment record, whose start position is greater or equal to `start`. Read extraction functionality for overlapping reads must be handled by the user with a manual offset or can be seen as a future feature request which is not part of this issue.

temehi:
lgtm! 👍

rrahn:
travis failed on g++7. But there was also a compile issue. I restarted the job.

rrahn:
Alright. Seems to be ok now.

PR #2226, opened 2018-01-16
[FIX] Don't go down the wavelet tree twice when retrieving hits

cpockrandt:
Saves some time when calling ``getOccurrences()``.

Before, the value (i.e. character) in the bwt at a given position was retrieved by going down the tree and then the rank was computed up to the given position using that value (which leads to going down the tree twice). Now the tree is only processed once and the value/character is retrieved while computing the rank.


h-2:
CI fails....

cpockrandt:
@h-2 Not anymore 👼

cpockrandt:
Oops 🤣

PR #2233, opened 2018-01-23
[FIX] Fixes read bam check for empty quality string.

### LINKED ISSUE #2230: "Unknown/missing" QUAL field value in a BAM file is read in incorrectly; works correctly for SAM
In a SAM file, an asterisk ("*") in the QUAL field (11th column) indicates that quality information for the read is missing or unknown.  When reading such a record with `readRecord()`, the `qual` member of the `BamAlignmentRecord` is set to an empty `CharString`, just as the docs state.  But when the equivalent record in a BAM file (where an "unknown/missing" QUAL field is represented as a series of 0xFF byte values) is read in, the `qual` member is instead set to a number of space characters equal in length to the sequence length.  This obviously causes problems round-tripping, etc.

The attached MWE simply reads a SAM or BAM file from stdin, and writes all of its records verbatim to stdout.  The attached SAM file x.sam has a single 11-base-pair sequence, and the attached BAM file is just the equivalent in BAM format (produced by `samtools view -b < x.sam > x.bam`).  Running the code with either file on stdin should produce identical output (provided BAM support has been configured using `SEQAN_HAS_ZLIB`), but the BAM input produces 11 spaces in the rightmost column instead of a "*" character.  (I also output the `qual` member to stderr to determine whether the bug was in `readRecord()` or `writeRecord()`: the spaces appear here already, so the problem must be in `readRecord()`.)

The problem occurs with SeqAn version 2.3.1 as installed by the Ubuntu package libseqan2-dev version 2.3.1+dfsg-3ubuntu1.  I scrolled through 3 pages of bugs with the search terms "is:issue BAM" and didn't see any (open or closed) similar issues.
[seqan_sam_bam_bug_report.zip](https://github.com/seqan/seqan/files/1643554/seqan_sam_bam_bug_report.zip)

smehringer:
Fixes #2230


rrahn:
Travis failed.

smehringer:
Seems like we indirectly tested the wrong behaviour in some demo files from the tutorials. I changed to according `stdout` files.

PR #2236, opened 2018-01-23
[FIX] Fixes some spurious segfaults, due to unprotected condition broadcasting in unit tests.

rrahn:


tests/parallel/test_parallel_enumerable_thread_local.h
```diff
@@ -38,7 +38,7 @@
 
 namespace test_align_parallel
 {
-struct alignas(128) TestValue
+struct TestValue
```
h-2: why no alignment anymore?

rrahn: Well, it just doesn't make sense. Also not in the context tested.

PR #2238, opened 2018-01-24
Fixes some SIMD compile issues using gcc5

rrahn:
- detects internal compiler error of gcc5 raised in comparison operator of 64 byte vector.
 - fixes type truncation when using createVector with different vector host type and value type.


rrahn:
Error is due to memory issue on travis.

PR #2241, opened 2018-01-26
Fix/icpc align simd test

### LINKED ISSUE #2199: icpc-18 issues
* [x] warnings during cmake:
>  Intel compiler >=17.0.5 reevaluate if [test_align_simd_*] can be compiled.

  * [x] test_test_align_simd segfaults
  * [x] app_test_yara (probably the same as #2186)
  * [x] check icpc-17 again.

rrahn:
Fixes #2199
Reenables the simd align test for the intel compiler.
It seems there is some compiler issue with the use of the ZipView.
So the error was actually not related to the simd implementation.

### DIFF
--- a/tests/simd/test_simd_vector.h
+++ b/tests/simd/test_simd_vector.h
@@ -76,7 +76,9 @@ inline void test_matrix_transpose()
 //    std::cout << std::endl;
 //    for(int i=0;i<DIM;++i)
 //        print(std::cout, tmp[i]) << std::endl;
-
+#if defined(__x86_64__)
+    _mm_empty();  // Fixes icpc warning #13203: No EMMS instruction before call to function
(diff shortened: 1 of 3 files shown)


tests/simd/test_simd_vector.h
```diff
@@ -76,7 +76,9 @@ inline void test_matrix_transpose()
 //    std::cout << std::endl;
 //    for(int i=0;i<DIM;++i)
 //        print(std::cout, tmp[i]) << std::endl;
-
+#if defined(__x86_64__)
```
h-2: Please also check `defined(__amd64__)` as some platforms don't define `__x86_64__`.

rrahn: 👍

PR #2244, opened 2018-01-26
[FIX] Fixes correct detection of end-of-file marker and some synchronisation issue.

### LINKED ISSUE #1927: A rare deadlock case in bam file loading
I tried to read 100G bam file, and it couldn't finish the job.
I think it's about event management of multiple thread in bgzf stream since CPU usage became 0.

Since # of threads is fixed in bgzf stream(#1899), I manually changed this to 1 thread / 1 job.
And now it's not working with relatively small file( < 10GB).

I'm working on this.

rrahn:
Fixes #1927 
Do not merge yet.
I need to look into something first.
But you can have a first review on the changes.

### DIFF
--- a/include/seqan/stream/iostream_bgzf.h
+++ b/include/seqan/stream/iostream_bgzf.h
@@ -339,6 +339,7 @@ public:
         std::mutex              cs;
         std::condition_variable readyEvent;
         bool                    ready;
+        bool                    bgzfEofMarker;
 
         DecompressionJob() :
             inputBuffer(BGZF_MAX_BLOCK_SIZE, 0),
@@ -347,7 +348,8 @@ public:
             size(0),
             cs(),
             readyEvent(),
-            ready(true)
+            ready(true),
+            bgzfEofMarker(false)
         {}
 
         // TODO(rrahn): Do we need a copy constructor for the decompression job.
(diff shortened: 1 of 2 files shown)


h-2:
> Do not merge yet.

Is this still valid?

rrahn:
Sorry, yes it is valid. Can now be merged.

include/seqan/stream/iostream_bgzf.h
```diff
@@ -444,6 +448,14 @@ class basic_unbgzf_streambuf :
                             (char*)&job.inputBuffer[0] + BGZF_BLOCK_HEADER_LENGTH,
                             tailLen);
 
+                        // Check if end-of-file marker is set
+                        if (memcmp(reinterpret_cast<uint8_t const *>(&job.inputBuffer[0]),
+                                   reinterpret_cast<uint8_t const *>(&BGZF_END_OF_FILE_MARKER[0]),
+                                   28) == 0)
```
h-2: Do we now check *every block* for the EOF marker?

rrahn: This is a safety guard.
The specification says, that applications that execute in append mode, can adde EOF marker, but then add new blocks to it. So in that particular case it would be insufficient to stop just here. In addition the followed uncompressed block must have zero length to indicate the true end of the file. Detecting the EOF marker followed by an empty block would be safe indicator for the end of the file. So we could avoid the check here, but I am not sure if this will have a measurable effect on  Bam-IO. I will try a microbenchmark to see if it has any effect.

rrahn:
elapsed time averaged over 4 runs streaming a 14 GB bam file:
* w/ fix 69.75s
 * w/o fix 69.75 s

PR #2246, opened 2018-01-27
Fix/clang 4 simd

rrahn:
clang-4.0.1 still does not support avx512_skx, so we adapt the developer info and disable the tests on our nightlies.


h-2:
Actually this breaks clang-3.5 in the CI? @rrahn Can you investigate?

rrahn:
No the error is rather not related. But I hope #2248 will solve this spurious failing test

PR #2247, opened 2018-01-27
Revert "[FIX] SIMD: Fixes createVector truncated value type issue."

rrahn:
This reverts commit 6df610cdf93d82efd631790d3dc799ab956d17f7.

It broke UME Simd and I could not reproduce the error after reverting it.


h-2:
wow, it got marked as merged automatically

PR #2254, opened 2018-01-31
[FIX] Fixes compiler warning on gcc-8.

rrahn:


rrahn:
@h-2 It seems I did not make a PR for this?

Issue #2245, opened 2018-01-26
New platform issues


h-2:
:sob: 

https://buildd.debian.org/status/package.php?p=seqan2&suite=experimental

- [x] test_align_parallel_data_structures [ppc64el]
- [x] test_demo_tutorial_file_io_overview_solution3 [ppc64el, arm64, s390x, powerpc, ppc64]
- [x] test_demo_tutorial_sam_and_bam_io_example7 [ppc64el, arm64, s390x, powerpc, ppc64]
-  ~~enable UME simd in nightly builds on larix.~~

@rrahn can you look at the first one, I will have a look at the other two.

rrahn:
Not tonight but maybe tomorrow. Got some other issues with some change in simd related code. :unamused:

h-2:
> Not tonight but maybe tomorrow. 

Yeah, me too, definetely not before sunday.

h-2:
#2233 introduced errors 2 & 3, likely an issue of signed/unsigned char over/underflows. Possibly just using `0xFF` instead of `'\xFF'` will already fix this.

rrahn:
@h-2 can you please include UME simd 0.8.1 in the nightlies on larix.
I will do it for mac

h-2:
Wow, would you guess the result of this program:
```cpp
#include <iostream>
int main()
{
    std::cout << ((32 - '!') == 0xFF) << '\n';
    std::cout << ((32 - '!') == '\xff') << '\n';

    std::cout << unsigned(0xFF) << '\n';
    std::cout << unsigned('\xff') << '\n';
    std::cout << unsigned(char(0xFF)) << '\n';
    std::cout << unsigned(char('\xff')) << '\n';

    std::cout << (char(32 - '!') == 0xFF) << '\n';
    std::cout << (char(32 - '!') == '\xff') << '\n';
}
```

Once with just g++ and once with `-funsigned-char`?

<details> 
  <summary>signed char</summary>

```
0
1
255
4294967295
4294967295
4294967295
0
1
```
</details>

<details> 
  <summary>unsigned char</summary>

```
0
0
255
255
255
255
1
1
```
</details>

That in combination with `typeof(char + char) == int` can really ruin the day.

h-2:
fixed by #2248 and #2249

h-2:
@mr-c Can you give -rc1 a try? Hopefully all new platform issues are resolved now.

PS: Link to to Debian Med was added!

h-2:
[...]

PR #2255, opened 2018-01-31
[FIX,TEST] deactivate some more tests on alpha

h-2:
see #2141 #2245


tests/index/test_index_bifm.cpp
```diff
@@ -198,6 +199,7 @@ SEQAN_TYPED_TEST(BidirectionalFMIndexTest, SearchInStringSet)
         }
     }
 }
+#endif
```
rrahn: Can you please add a `// __alpha__` as closure comment for the `#endif`?

PR #2260, opened 2018-02-01
Release/adjust copyright

h-2:
@rrahn please merge if all tests pass

PR #2261, opened 2018-02-01
Release/adjust copyright

rrahn:
Supersedes #2260


rrahn:
Since the other build went through I will merge it.
I just modified the script and rerun it basically.

Issue #2258, opened 2018-02-01
Vertex deletion can give index-out-of-range crash in connectedComponents()


wtwhite:
Deleting a vertex that is not the last (highest-ID) vertex and then calling `connectedComponents()` causes a crash.  This occurs when a `UnionFind` object tries to access an element past the end of its internal `_values` member, a `String`.  The attached MWE turns on `SEQAN_ENABLE_DEBUG` to force an assertion failure; without this, sometimes you see a segfault, sometimes nothing.

The root cause seems to be incorrectly assuming that the set of vertex IDs of the vertices in the graph is exactly equal to the range 0 to `numVertices(g)-1`:

1. In `resizeVertexMap()` on line 295 of union_find.h, which is called near the start of `connectedComponents()`, only `numVertices(g)` size is allocated for the member `_values`.
2. As groups of vertices are combined later on, it is assumed that `_values[id]` is valid for any vertex ID `id`, in particular on line 329, inside `findSet()`.  If non-last vertices have been deleted, then there can be valid vertex IDs with values >= `numVertices(g)`, and this causes the assertion failure.

I think the best thing to do would be to change this overload of `resizeVertexMap()` to forward to the `resizeVertexMap()` that corresponds to the underlying Graph's type -- these functions (which seem to be implemented in property_map_generic.h) use `getIdUpperBound()` instead of `numVertices()` to decide how big to make the `String`.  That should fix this specific bug, but there could be other instances of assuming "dense vertex IDs" lurking in union_find.h.

I'll try to get this working myself and give a patch if I'm successful (could take me a while though).

**MWE:** [test_removeVertex_bug.zip](https://github.com/seqan/seqan/files/1685727/test_removeVertex_bug.zip)

PR #2266, opened 2018-02-02
Fix connected components issue2258

wtwhite:
The `_values` member was being resized to the number of vertices currently in the graph, which works if vertices have never been deleted.  But if vertices have ever been deleted, then there might exist vertices in the graph with higher IDs, and trying to look these up in `_values` later on would cause an assertion failure or segfault.

After this fix, the MWE in issue #2258 correctly reports 1 component in the graph instead of crashing.


h-2:
@smehringer Can you have a look?

h-2:
@wtwhite Thanks for the PR!

h-2:
Reassigning to @joergi-w as he will take over maintainership of graphs in SeqAn!

include/seqan/misc/union_find.h
```diff
@@ -298,7 +298,8 @@ resizeVertexMap(UnionFind<TValue> & unionFind,
                 Graph<TSpec> const & g)
 {
     clear(unionFind);
-    return resize(unionFind, numVertices(g));
+    resizeVertexMap(unionFind._values, g, -1);
+    return length(unionFind);		// Note: Other resizeVertexMap() overloads return void...
```
joergi-w: True. I might change this function to `void` in the future for consistency and the returned value is not used throughout the SeqAn library anyway. Thank you for pointing that out.

wtwhite: You're welcome :)

Issue #2257, opened 2018-02-01
jumpToRegion has trouble finding a read


hannespetur:
Hello SeqAn team,

Thank you for all your hard work. I am writing this issue because I believe I have discovered a bug `seqan::jumpToRegion(...)` in the latest development files of SeqAn. To reproduce my problem, consider the following `test.sam` file with only one read at position 100:

```
@HD	VN:1.3	SO:coordinate
@SQ	SN:ref	LN:100000
schrodinger_read	0	ref	100	8	18M	=	31	40	CATCATCATCATCATCAT	*
```

Convert to BAM and index it:
```
samtools view test.sam -b -o test.bam
samtools index test.bam
```

I am using (a slightly modified version of the) [BAM indices tutorial](http://seqan.readthedocs.io/en/master/Tutorial/InputOutput/SamAndBamIO.html#using-bam-indices) which reads my `test.bam` and `test.bam.bai` files. By changing the `int beginPos = 9, endPos = 30;` to other positions revealed some strange behavior...
 * If `beginPos = 1, endPos = 99999` then `hasAlignments` is `false` and the file stream points to the end of the file. I expected `hasAlignments == true` and the file stream points to the read.
 * If `beginPos = 2, endPos = 99999` then `hasAlignments` is `false` and the file stream points to the end of the file. I expected `hasAlignments == true` and the file stream points to the read.
 * If `beginPos = 16384, endPos = 99999` then `hasAlignments` is `false` and the file stream points to the end of the file. This is expected.
 * If `beginPos = 16385, endPos = 99999` then `hasAlignments` is `false` **BUT** now the file stream points to the read. I expected it would have the same behavior as `beginPos = 16384`.

It seems that if a read is in the very first BAM index bin (which is of size 16384) then searching that bin does not work.

I might even have found another bug while investigating this, because if I move the read to position 20000, recreate the BAM+BAI and run again with the same regions then the file stream points correctly to the read in all scenarios but hasAlignments is always `false`. I expected `hasAlignments == true` in all cases.

All the best,
(Icelandic) Hannes

rrahn:
Hi Hannes, 

many thanks for your thorough analysis. We'll investigate this.

h-2:
@smehringer it looks like you "verschlimmbessert" the situation :laughing:

PR #2273, opened 2018-02-05
[FIX/FEATURE] Fixes jumpToRegion. New viewRecord function.

smehringer:
Related to #2257 . Do not merge yet.

The jumpToRegion() functionality was cleaned up since it did unnecessarily use the BAM binning index although it did not support overlaps or complete read extraction. It now only uses the linear index to jump to the first read that STARTS in the specified region.

The function `viewRecords` shall implement `samtools view`. Given a region `(start, end)` all reads that OVERLAP this region by at least one base, excluding start and end, are extracted from the BAM file.

The current code needs more testing because the region boundaries might not be exactly like samtools. As far as I can tell, samtools excludes the 1-based region boundaries from the overlap. The code tries to imitate the exact same, expecting 1-based coordinates to be consistent. Long read benchmarks are also needed.

But I wanted to put up the code for reviews about the design and possible errors and to make sure that this PR will be part of 2.4.1 as soon as possible.


include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,151 @@ class BamIndex<Bai>
...
 
+// ----------------------------------------------------------------------------
+// Function viewRecords()
+// ----------------------------------------------------------------------------
+
+/*!
+ * @fn BamFileIn#viewRecords
+ * @brief Extract all reads overlapping a region from a BamFileIn using an index.
+ *
+ * You provide a region <tt>[regionStart, regionEnd]</tt> on the reference <tt>refID</tt>
```
hannespetur: For me the function seems to extract from the region `(regionStart, regionEnd]` but samtools gives me reads from the region `[regionStart, regionEnd]`. For example, with a read of just 2 bp at position 200 (1-based) then both viewRecords and samtools view find the when `regionStart = 200; regionEnd = 200` but only samtools finds the read when  `regionStart = 201; regionEnd = 201`. So there is probably a very minor off-by-one error somewhere...

smehringer: Darn. After my benchmarks I thought samtools extraxts reads from `(reagionStart, regionEnd)` but maybe I got confused by the BED 0-based 1-based conversion. Does samtools have any documentation on this? I couldn't find a statement.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,151 @@ class BamIndex<Bai>
...
+ *
+ * This function fails if <tt>refID</tt>/<tt>regionStart</tt> are invalid.
+ */
+
+//! See SAM specifications https://samtools.github.io/hts-specs/SAMv1.pdf page 16
+static inline void
+_baiReg2bins(std::vector<uint16_t> & list, uint32_t beg, uint32_t end)
+{
+    SEQAN_ASSERT_LEQ(beg, end);
+    SEQAN_ASSERT_LEQ(end, 536870912); // BAM format only allows reference of size < 2^29
```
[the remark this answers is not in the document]
smehringer: Thanks :)

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,151 @@ class BamIndex<Bai>
...
+    // -------------------------------------------------------------------------
+    uint64_t linearMinOffset;
+    if (empty(bamIndex._linearIndices[refId]))
+    {
+        linearMinOffset = 0; // if no linear index is given, value 0 never limits chunk seeking
+    }
+    else
+    {
+        unsigned windowIdx = regionStart >> 14; // Linear index consists of 16kb windows.
+        linearMinOffset = bamIndex._linearIndices[refId][windowIdx];
```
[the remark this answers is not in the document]
smehringer: Hmmm. This is strange. Since `regionStart` is within the length of the sequence (I suppose you didn't give abnormal bounds) I thought this could never happen.. because the index stores an offsett for ever 16kb window of the reference.. and if we do not store empty windows the access to `linearIndice` is not trivial.

include/seqan/bam_io/bam_index_bai.h
hannespetur: I couldn't find it clearly specified either but in my tests samtools view always gave me an inclusive region. I tried both using (1-based) `samtools view test.bam RNAME:BEGIN-END` and a (0-based) BED file with the `-L <BED>` option.

P.S. The `-L <BED>` option in samtools does not use the BAI index, but instead reads the entire BAM... :/

include/seqan/bam_io/bam_index_bai.h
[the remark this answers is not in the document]
hannespetur: It seems it happens when I specify a region that is in a index window/bin that is after the last read's window. For example, if I have a BAM file with one 100 bp read at position 60000 then querying [20000-29000] works fine (no reads found) but querying [80000-89000] throws an assertion failure. I think we just need a little check for this case.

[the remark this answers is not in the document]
smehringer: What is the length of your reference, specified in the BAM header, for this example?

include/seqan/bam_io/bam_index_bai.h
smehringer: Ops.. I tricked myself. :roll_eyes: 
I read in the BED file but didn't convert to 1-based coordinates again before I used `viewRecords`.
Thanks for the benchmarking!

include/seqan/bam_io/bam_index_bai.h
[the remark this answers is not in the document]
hannespetur: `LN:100000` but it does not seem to have any effect if I change it. `LN:1` and `LN:10000000` also gives my identical results.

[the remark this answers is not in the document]
smehringer: Thanks for trying it. Then I guess the linear index ends either when the reference length is reached or there are no more records(/end of file is reached). I'll add a check.

smehringer:
@hannespetur I updated the code. Can you run your tests again? :smiley:

include/seqan/bam_io/bam_index_bai.h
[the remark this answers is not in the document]
hannespetur: Thanks, I will give it a try when it is ready. And sorry that I wasn't very clear before, but what I meant was that the linear index length is not depended on the reference length (`LN`) at all. The assertion failure only happens when you specify a region that is behind the last record on that contig/chromosome (does not need to be the end of the file). If the last record is much further than the reference length the linear index also reaches as far as the last record.

For example, I was still able to find the read at position 60000 even when I had `LN:1` in the BAM header. In that case the linear index reaches much further than the specified reference length.

hannespetur:
@smehringer Thank you, yes I will do it now.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+inline bool
+_getMinFileOffset(uint64_t & minFileOffset,
+                  BamIndex<Bai> const & bamIndex,
+                  int32_t refId,
+                  int32_t position)
+{
+    unsigned windowIdx = position >> 14; // Linear index consists of 16kb windows.
+
+    if (windowIdx >= length(bamIndex._linearIndices[refId]))
+	return false;
```
hannespetur: Replace TAB with 8 spaces. 😄

hannespetur:
@smehringer LGTM (well, except for that one super tiny TAB->spaces remark). All my tests all pass now!

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+
+    minFileOffset = bamIndex._linearIndices[refId][windowIdx];
+
+    return false;
```
hannespetur: => `return true;`

hannespetur:
@smehringer  ..I spoke too soon. I found another problem.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+}
+
+template <typename TContainer, typename TSpec>
+inline
+SEQAN_FUNC_ENABLE_IF(And<IsSameType<typename Value<TContainer>::Type, BamAlignmentRecord>,
+                     Not<IsSameType<typename Value<TContainer>::Type, TContainer>>>, bool)
+viewRecords(TContainer & resultContainer,
+            FormattedFile<Bam, Input, TSpec> & bamFile,
+            BamIndex<Bai> const & bamIndex,
+            int32_t refId,
```
rrahn: all non-mutable elements should be passed `const`-qualified

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+    // the given region to be also one-based (To be consistent with samtools).
+    // Since SeqAn works 0-based we need to substract 1.
+    SEQAN_ASSERT_GEQ_MSG(regionStart, 1, "Expect 1-based region interval which excludes 0 as a start pos.");
+    --regionStart;
+    --regionEnd;
+
+    // Retrieve the candidate bin identifiers for [regionStart, regionEnd).
+    // -------------------------------------------------------------------------
+    std::vector<uint16_t> candidateBins;
+    _baiReg2bins(candidateBins, regionStart, regionEnd);
```
rrahn: Careful, I think in _baiReg2bins you still expect 1-based positions, don't you?
At least for `regionEnd`?

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+ */
+
+//! See SAM specifications https://samtools.github.io/hts-specs/SAMv1.pdf page 16
+static inline void
+_baiReg2bins(std::vector<uint16_t> & list, uint32_t beg, uint32_t end)
+{
+    SEQAN_ASSERT_LEQ(beg, end);
+    SEQAN_ASSERT_LEQ(end, 536870912u); // BAM format only allows reference of size < 2^29
+
+    --end; // zero based index
```
rrahn: I think this should be removed.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+
+    // Sanity checks
+    // -------------------------------------------------------------------------
+    if (!isEqual(format(bamFile), Bam()))
+        return false;
+
+    if (refId < 0)
+        return false;  // Cannot seek to invalid reference.
+
+    if (static_cast<unsigned>(refId) >= length(bamIndex._binIndices))
```
rrahn: maybe better use `decltype(length(bamIndex._binIndices))`

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+    for (uint16_t bin : candidateBins)
+    {
+        TMapIter mIt = bamIndex._binIndices[refId].find(bin);
+
+        if (mIt == bamIndex._binIndices[refId].end()) // empty bins are not stored
+            continue;
+
+        for (TChunkIter chunkIt = begin(mIt->second.chunkBegEnds, Rooted()); !atEnd(chunkIt); goNext(chunkIt))
+        {
+            if (chunkIt->i2 < linearMinOffset)
```
rrahn: Maybe a comment about `i2` being the end file offset of the current chunk.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+
+            while(position(bamFile) != static_cast<int>(chunkIt->i2)) // this includes end of file
+            {
+                readRecord(record, bamFile);
+
+                if (record.beginPos > regionEnd)
+                    break; // chunk of consecutive reads is already out of bounds
+
+                // max(1, length) in if clause because if read is unmapped, the endPos calculation is wrong.
+                // (Could be also done ny checking if (unmapped) but might e less efficient than max.
```
rrahn: typo: by

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+                if (record.beginPos <= regionEnd &&
+                    static_cast<int>(record.beginPos + std::max(1u, getAlignmentLengthInRef(record)) - 1) >= regionStart)
+                    appendValue(resultContainer, record);
```
rrahn: I am not sure what the default behaviour is, but better, to specify `Generous()` as last parameter.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -180,28 +335,13 @@ class BamIndex<Bai>
  * @param[in]     regionEnd     The end of the region to jump to (<tt>int32_t</tt>).
  * @param[in]     index         The @link BamIndex @endlink to use for the jumping.
  *
- * @return bool true if seeking was successful, false if not.
+ * @return bool false if an error occurred while seeking.
```
rrahn: ..., otherwise `true`

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+            continue;
+
+        for (TChunkIter chunkIt = begin(mIt->second.chunkBegEnds, Rooted()); !atEnd(chunkIt); goNext(chunkIt))
+        {
+            if (chunkIt->i2 < linearMinOffset)
+                continue; // use linearMinOffset to discard unneccessary chunks
+
+            setPosition(bamFile, chunkIt->i1); // jump to candidate chunk
+
+            while(position(bamFile) != static_cast<int>(chunkIt->i2)) // this includes end of file
```
rrahn: Please use `int32_t` in the static asserts.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -330,9 +389,9 @@ jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
         readRecord(record, bamFile);
 
         if (record.beginPos >= regionEnd)
-            break;
+            break; // no records can be found anymore iin coordinate sorted file
```
rrahn: typo: iin

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+ *
+ * This function fails if <tt>refID</tt>/<tt>regionStart</tt> are invalid.
+ */
+
+//! See SAM specifications https://samtools.github.io/hts-specs/SAMv1.pdf page 16
+static inline void
+_baiReg2bins(std::vector<uint16_t> & list, uint32_t beg, uint32_t end)
+{
+    SEQAN_ASSERT_LEQ(beg, end);
+    SEQAN_ASSERT_LEQ(end, 536870912u); // BAM format only allows reference of size < 2^29
```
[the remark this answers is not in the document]
rrahn: The specification says `... does not support reference chromosome sequences longer than 2^29 − 1`.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -211,117 +351,36 @@ jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
...
-    // Compute offset in BGZF file.
-    // ------------------------------------------------------------------------
-    uint64_t offset = std::numeric_limits<uint64_t>::max();
+    if (regionStart > regionEnd)
+        return false; // invalid region
 
-    // Retrieve the candidate bin identifiers for [regionStart, regionEnd).
-    String<uint16_t> candidateBins;
-    _baiReg2bins(candidateBins, regionStart, regionEnd);
+    if (regionEnd > 536870912)
```
rrahn: I think it must be `>=`

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -211,117 +351,36 @@ jumpToRegion(FormattedFile<Bam, Input, TSpec> & bamFile,
...
-
-        if (record.beginPos >= regionEnd)
-            break;  // Cannot find overlapping any more.
-    }
-
-    if (offset == std::numeric_limits<uint64_t>::max())
-        return true; // Finding no overlapping alignment is not an error, hasAlignments is false.
+    // -------------------------------------------------------------------------
+    uint64_t linearMinOffset;
+    if (!_getMinFileOffset(linearMinOffset, index, refId, regionStart))
```
rrahn: I think the linear index uses 0-based index for the windows.
So to make it consistent with viewRecords, I would assume that regionStart is passed as 1-based position. 
That should be then here corrected.

hannespetur:
@rrahn Well, it seems like I overlooked a few problems in this PR 😃 And sure. I will find time to add some tests soon.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +154,160 @@ class BamIndex<Bai>
...
+ * @signature bool viewRecords(bamFileIn, bamIndex, refID, regionStart, regionEnd, index);
+ *
+ * @param[in,out] resultContainer Container of @link BamAlignmentRecord @endlink 's to store the alignments.
+ * @param[in,out] bamFileIn       The @link BamFileIn @endlink to jump with.
+ * @param[in]     bamIndex        The @link BamIndex @endlink to use for the jumping.
+ * @param[in]     refID           The reference id to jump to (<tt>int32_t</tt>).
+ * @param[in]     regionStart     The begin of the region to jump to (<tt>int32_t</tt>).
+ * @param[in]     regionEnd       The end of the region to jump to (<tt>int32_t</tt>).
+ *
+ * @return bool true if seeking was successful, false if an error occurred or no
```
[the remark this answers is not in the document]
smehringer: Hm yeah lets discuss 😄 
I agree that returning a bool might not be the perfect solution, but 
1) isn't the SeqAn2 "style" an in-out-parameter instead of returning it? Even `readRecord()` gets the record passed as in-out instead of returning it.
 2) Wouldn't I force the user to use a specific container (e.g. `String<BamAlignmentRecord>`) that he eventually wants to change (e.g. vector, array or list) thereby copying?
3) I deliberately documented to **append** the reads to the container such that the user can call  `viewRecords(container, ..region1..); viewRecords(container, ..region2..)` and has all records of both regions in the container. I thought this might be useful. And it might be more intuitive than appending a return value, no?

Just my thoughts on this before we change it. 
(PS: If we do return a container I'd prefer a std::vector..)

include/seqan/bam_io/bam_index_bai.h
smehringer: No and yes. The function expects 0-based coordinates but was computing a half open interval (reason for `--end` ). Since we want to mimic samtools and samtools computes a closed interval we will do the same now. So thanks for pointing it out.

smehringer:
@rrahn final review please.
@hannespetur can you sign the clahub? (I signed it so it must be you it's angry about)

hannespetur:
Clahub is now as signed as an int.

hannespetur:
But I think we cannot have the test the 0-based interval usage, since it triggers an assertion failure and thus fails the test.

smehringer:
ops, I thought tests are built in release mode..

smehringer:
@rrahn Could you do review this again please?

tests/bam_io/test_bam_index.h
```diff
@@ -75,39 +75,145 @@ SEQAN_DEFINE_TEST(test_bam_io_bam_index_open)
...
+    SEQAN_ASSERT_EQ(back(records).beginPos, 1534);
+    clear(records);
+
+    viewRecords(records, bamFile, baiFile, 1, 300, 400);
+    SEQAN_ASSERT_EQ_MSG(length(records), 196u, "196 reads expected in region seq2:300-400");
+    SEQAN_ASSERT_EQ(front(records).beginPos, 265);
+    SEQAN_ASSERT_EQ(back(records).beginPos, 398);
+    clear(records);
+}
+
```
smehringer: newline detected by @rrahn, your private investigator you can trust :male_detective:

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -180,159 +341,77 @@ class BamIndex<Bai>
  * @param[in]     regionEnd     The end of the region to jump to (<tt>int32_t</tt>).
  * @param[in]     index         The @link BamIndex @endlink to use for the jumping.
  *
- * @return bool true if seeking was successful, false if not.
+ * @return bool false if an error occurred while seeking, otherwise true.
```
rrahn: I usually like to emphasise `false` and `true`

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +155,164 @@ class BamIndex<Bai>
...
+        return; // invalid region
+
+    // Adjust region
+    // -------------------------------------------------------------------------
+    // We want to mimic samtools view. Since SAM/BAM files are 1-based we expect
+    // the given region to be also 1-based, to be consistent with samtools.
+    // Since SeqAn works 0-based we need to subtract 1.
+    SEQAN_ASSERT_GEQ_MSG(regionStart, 1u, "Expect 1-based region interval which excludes 0 as a start pos.");
+
+    if (regionStart != 0) // avoid overflow in case of wrong usage of user
```
rrahn: I would assert this as well. Otherwise [0,1] would be a valid interval by the user but might give you unexpected results.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -180,159 +341,77 @@ class BamIndex<Bai>
...
+    // Adjust region
+    // -------------------------------------------------------------------------
+    // We want to be consistent with viewRecords and thus expect 1-based coordinates.
+    // Since SeqAn works 0-based we need to subtract 1.
+    SEQAN_ASSERT_GEQ_MSG(regionStart, 1u, "Expect 1-based region interval which excludes 0 as a start pos.");
 
-        if (record.beginPos >= regionEnd)
-            break;  // Cannot find overlapping any more.
-    }
+    if (regionStart != 0) // avoid overflow in case of wrong usage of user
```
rrahn: same as above.

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -61,6 +61,7 @@
 #ifndef INCLUDE_SEQAN_BAM_IO_BAM_INDEX_BAI_H_
 #define INCLUDE_SEQAN_BAM_IO_BAM_INDEX_BAI_H_
 
+
```
rrahn: 😱

include/seqan/bam_io/bam_index_bai.h
```diff
@@ -154,6 +155,183 @@ class BamIndex<Bai>
...
+        SEQAN_THROW(std::logic_error("BAM index bin directory does not match the given reference identifier 'refId'. "
+                   "Maybe your BAM index is corrupted or refers to a different BAM file."));
+        return;
+    }
+
+    if (regionEnd > static_cast<uint32_t>(contigLengths(context(bamFile))[refId]))
+    {
+        SEQAN_THROW(std::logic_error("The 'regionEnd' parameter point to a position greater than the length of the reference."));
+        // correct end positions that exceed the reference length for release build.
+        regionEnd = static_cast<uint32_t>(contigLengths(context(bamFile))[refId]);
```
rrahn: Well, you can remove the statements after the throw statement as they will never be executed.

include/seqan/bam_io/bam_index_bai.h
smehringer: DAMMIT

smehringer: damn git auto merge

PR #2277, opened 2018-02-08
[FIXES-#2268] further pthread fixes on old ubuntu

### LINKED ISSUE #2268: Commit b09e87e0 causes link error "undefined reference to `pthread_create'" on Linux
On Linux, SeqAn seems to need libpthread for multithreaded BAM file I/O (at least).  But since commit b09e87e0 (I think; I haven't tried git bisect to verify), using any SeqAn code that needs libpthread leads to the linker error "undefined reference to `pthread_create'".

TTBOMK, the reason is: Prior to commit b09e87e0, libpthread was included by appending `pthread` to `SEQAN_LIBRARIES` in CMakeLists.txt, which has the effect of *appending* `-lpthread` to the command line sent to the linker.  Since that commit, instead of appending to `SEQAN_LIBRARIES`, some linker options including `-lpthread` are appended to `CMAKE_EXE_LINKER_FLAGS`, which has the effect of *prepending* `-lpthread` to the command line sent to the linker -- and since the linker (on Linux anyway) only extracts library object code for symbols referenced by object files *that appeared earlier on the linker command line*, this means that `pthread_create` is never found, causing the link failure.  (FWIW this linker behaviour seems crazy to me.  But perhaps there's some justification if you dig deep enough...)

What puzzles me is why this build error apparently hasn't caused problems with SeqAn's own CI tests.  I'm running a very standard Ubuntu Linux 4.8.0-59-generic with g++ 6.3.0, so I don't think it's some peculiarity of my system.  I'm on the `develop` branch.

Note that prior to this commit, the "new" behaviour of using `CMAKE_EXE_LINKER_FLAGS` instead of `SEQAN_LIBRARIES` to include libpthread could already occur under other conditions (when building statically, it would seem), and it was possibly broken for that also.

**Workaround:** Manually specify `SEQAN_LIBRARIES` when running CMake.  For me this amounted to adding `-DSEQAN_LIBRARIES="pthread;rt;/usr/lib/x86_64-linux-gnu/libz.so"` to the CMake command line.

h-2:
fixes #2268

### DIFF
--- a/apps/mason2/CMakeLists.txt
+++ b/apps/mason2/CMakeLists.txt
@@ -103,14 +103,14 @@ add_executable (mason_splicing mason_splicing.cpp)
 add_executable (mason_tests mason_tests.cpp)
 
 # Add dependencies found by find_package (SeqAn).
-target_link_libraries (mason_frag_sequencing ${SEQAN_LIBRARIES} mason_sim)
-target_link_libraries (mason_genome ${SEQAN_LIBRARIES} mason_sim)
-target_link_libraries (mason_materializer ${SEQAN_LIBRARIES} mason_sim)
-target_link_libraries (mason_methylation ${SEQAN_LIBRARIES} mason_sim)
-target_link_libraries (mason_simulator ${SEQAN_LIBRARIES} mason_sim)
-target_link_libraries (mason_splicing ${SEQAN_LIBRARIES} mason_sim)
-target_link_libraries (mason_tests ${SEQAN_LIBRARIES} mason_sim)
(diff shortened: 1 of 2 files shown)


h-2:
Wow, apparently it does make it difference and triggers a problem with double-include of pthread in mason (which I am fixing now).

h-2:
(Double-linking pthread actually being a problem is only due to old eglibc on old Ubuntu/travis, but oh well)

h-2:
!rebuild

PR #2282, opened 2018-02-11
More cmake fixes

SoapGentoo:
@h-2 I've found two niggles whilst packaging SeqAn 2.4.0 for Gentoo: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=3155495650f555369e58b58f54fc6766e9691f20

1. Fix pthread linking. Using `CMAKE_EXE_LINKER_FLAGS` is the wrong variable for specifying `-lpthread` due to the fact that it comes before the object files. One of the reasons this is a problem for Gentoo is that we build with `-Wl,--as-needed` globally to reduce overlinking, which then causes build failures, as you're specifying a dependency before the linker encounters a consumer of said dependency.
2. Do not install individual `LICENSE` files. A software is licensed because you clearly state what license it falls under, not because you install a `LICENSE` file with it. Imagine if every package in a distro installed the GPL/BSD/MIT licenses, you would needlessly be adding redundant files.

### DIFF
--- a/apps/alf/CMakeLists.txt
+++ b/apps/alf/CMakeLists.txt
@@ -60,8 +60,7 @@ install (TARGETS alf
 
 # Install non-binary files for the package to "." for app builds and
 # ${PREFIX}/share/doc/alf for SeqAn release builds.
-install (FILES LICENSE
-               README
+install (FILES README
          DESTINATION ${SEQAN_PREFIX_SHARE_DOC})
 install (FILES example/small.fasta
          DESTINATION ${SEQAN_PREFIX_SHARE_DOC}/example)
--- a/apps/bs_tools/CMakeLists.txt
+++ b/apps/bs_tools/CMakeLists.txt
@@ -103,8 +103,7 @@ install (TARGETS bisar casbar four2three
 
 # Install non-binary files for the package to "." for app builds and
 # ${PREFIX}/share/doc/bs_tools for SeqAn release builds.
-install (FILES LICENSE
-               README
+install (FILES README
          DESTINATION ${SEQAN_PREFIX_SHARE_DOC})
 #install (FILES example/fasta1.fa
 #               example/fasta2.fa
--- a/apps/dfi/CMakeLists.txt
+++ b/apps/dfi/CMakeLists.txt
@@ -60,8 +60,7 @@ install (TARGETS dfi
 
 # Install non-binary files for the package to "." for app builds and
 # ${PREFIX}/share/doc/dfi for SeqAn release builds.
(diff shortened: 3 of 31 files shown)


h-2:
Thanks for the PR!

I have already addressed 1. in #2277 but I wasn't aware of `CMAKE_THREAD_LIBS_INIT`, I'll check it out.

Concerning the LICENSE files we need to discuss this, because we also build individual packages  (and e.g. GPL requires the license file to be included); also the apps have different copyright holders sometimes and we need to document this somewhere.

h-2:
> but I wasn't aware of CMAKE_THREAD_LIBS_INIT, I'll check it out.

I don't think we can use this, because it doesn't set whole-archive.

SoapGentoo:
@h-2 the GPL only requires the text to be included **with distribution, not with installation**. Concerning `-lpthread`, isn't it possible to choose a different approach here? The general way to build mostly static binaries is to compile them against an ancient glibc and leave `libc.so`/`libpthread.so` etc dynamically linked. Compiling `libpthread.a` into a binary bloats up size and adds no value against using an ancient glibc (which also produces smaller binaries).

SoapGentoo:
Also, you can always consider something along the lines of

```
if (${CMAKE_THREAD_LIBS_INIT})
    set (SEQAN_LIBRARIES ${SEQAN_LIBRARIES} "-Wl,--whole-archive ${CMAKE_THREAD_LIBS_INIT} -Wl,--no-whole-archive")
endif ()
```

which is a bit less aggressive to dietlibc/uclibc/musl.

SoapGentoo:
Final thought: In general `-Wl,--whole-archive` is only needed for symbols that aren't used directly, which is the case if you add `-lpthread` to `LDFLAGS` and not later on. If you're building with `LDFLAGS="-static"`, have you tried my patch? Given that `-lpthread` is at the end, `-Wl,--whole-archive` should not be necessary anymore. The only time I ever required `-Wl,--whole-archive` is for global static initialisers in static libraries, that don't have a symbol dependency path from some other executable/target.

h-2:
We have very little resources for SeqAn2 right now so I am reluctant to do any major build system changes that cause more headache later on or require us to change the entire deployment process (building against older versions, not being fully static, stripping out and re-adding LICENSE files to individual tarballs...).

I will look at #2277 again and try to get that in, because that has already been locally tested and is less invasive.

We will re-evalute the whole thing for SeqAn3 and consider your suggestions for dealing with pthread there. Beyond that I am afraid you might need to maintain local patches against SeqAn2 if the current status doesn't work for you.

PR #2285, opened 2018-02-12
[INTERNAL] Fix ctd output: add ctdVersion tag in header.

### LINKED ISSUE #2280: generated CTD missing required ctdVersion field
https://github.com/WorkflowConversion/CTDSchema/blob/master/CTD.xsd#L64

See also https://github.com/WorkflowConversion/CTDConverter/issues/35

  - [ ] ctdVersion not written
  - [ ] insegt: `UserWarning: Restriction [gff] of a single value found for parameter [fusion-output].`
  - [ ] flexbar: `Invalid default value(s) provided for parameter barcode-min-overlap of type <type 'int'>: ''`

smehringer:
Fixes #2280


rrahn:
lgtm! But I will also add a xmllinter test to check if the exported cat file fulfils the cdt shema.
Then we can rebase this PR and see if everything works correct.

PR #2286, opened 2018-02-14
Infrastructure adds automatic testing of correct ctd export of our tools

rrahn:
Adds automatic testing for the CTD exporter of the SeqAn apps, in order to detect downstream errors as by #2280 earlier in the pipeline.
Especially, if the CTD schema is updated.
This will break the builds but #2285 will fix it.
So after merging this, #2285  can be rebased and should fix all errors.

### DIFF
--- a/util/cmake/SeqAnBuildSystem.cmake
+++ b/util/cmake/SeqAnBuildSystem.cmake
@@ -358,6 +358,27 @@ macro (seqan_build_system_init)
         find_package(BZip2)
         find_package(Boost)
         find_package(SeqAn CONFIG REQUIRED)
+        find_package(LibXml2)
+    endif ()
+
+    option(CTD_TEST_ENABLED "Checks if correct ctd export of a tool can be tested automatically." OFF)
+
+    if (LIBXML2_FOUND)
+        find_program (XMLLINT_EXECUTABLE xmllint DOC "The xmllint tool available in the xmllib package.")
+        if (XMLLINT_EXECUTABLE-NOTFOUND)
+            message (STATUS "Could not find xmllint. Not configuring ctd tests.")
+        else ()
+            set(CTD_SCHEMA_ROOT "${PROJECT_BINARY_DIR}/vendor/ctd_schema")
+            execute_process(COMMAND git clone https://github.com/WorkflowConversion/CTDSchema.git ${CTD_SCHEMA_ROOT}
+                            OUTPUT_QUIET
+                            ERROR_QUIET)
+
+            if (NOT EXISTS "${CTD_SCHEMA_ROOT}/CTD.xsd")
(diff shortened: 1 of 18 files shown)


smehringer:
@rrahn Did you check that after fix #2285 no other errors occur?

rrahn:
Well not completely for all tests, but in a single application I did check if the ctdVersion is exported the tests will be successful.

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -358,6 +358,27 @@ macro (seqan_build_system_init)
...
+        if (XMLLINT_EXECUTABLE-NOTFOUND)
+            message (STATUS "Could not find xmllint. Not configuring ctd tests.")
+        else ()
+            set(CTD_SCHEMA_ROOT "${PROJECT_BINARY_DIR}/vendor/ctd_schema")
+            execute_process(COMMAND git clone https://github.com/WorkflowConversion/CTDSchema.git ${CTD_SCHEMA_ROOT}
+                            OUTPUT_QUIET
+                            ERROR_QUIET)
+
+            if (NOT EXISTS "${CTD_SCHEMA_ROOT}/CTD.xsd")
+                message (STATUS "Could not find CTD.xsd in <${CTD_SCHEMA_ROOT}>. Not confoguing ctd tests.")
```
smehringer: typo `confoguing`

smehringer:
LGTM (although I am very inexperienced with cmake)

util/cmake/SeqAnBuildSystem.cmake
```diff
@@ -358,6 +358,27 @@ macro (seqan_build_system_init)
...
+
+    option(CTD_TEST_ENABLED "Checks if correct ctd export of a tool can be tested automatically." OFF)
+
+    if (LIBXML2_FOUND)
+        find_program (XMLLINT_EXECUTABLE xmllint DOC "The xmllint tool available in the xmllib package.")
+        if (XMLLINT_EXECUTABLE-NOTFOUND)
+            message (STATUS "Could not find xmllint. Not configuring ctd tests.")
+        else ()
+            set(CTD_SCHEMA_ROOT "${PROJECT_BINARY_DIR}/vendor/ctd_schema")
+            execute_process(COMMAND git clone https://github.com/WorkflowConversion/CTDSchema.git ${CTD_SCHEMA_ROOT}
```
h-2: maybe use `ExternalProject_Add(` instead?

rrahn: Well yes, but it actually wants to build then the source. I didn't try but it did not seem to be the right choice here. I couldn't find an explicit documentation about how to disable the build step.

h-2:
Merging as-is.

PR #2295, opened 2018-03-17
Make alignment score calculation more generalisable

SoapZA:
Hi @h-2 
we'd like to merge one of the patches we've been carrying around downstream that makes alignment score calculation slightly more precise.

### DIFF
--- a/include/seqan/align/evaluate_alignment.h
+++ b/include/seqan/align/evaluate_alignment.h
@@ -239,12 +239,12 @@ TScoreVal computeAlignmentStats(AlignmentStats & stats,
             if (!isGapOpen0)
             {
                 stats.numGapOpens += 1;
-                stats.alignmentScore += scoreGapOpen(scoringScheme);
+                stats.alignmentScore += scoreGapOpenVertical(scoringScheme);
             }
             else
             {
                 stats.numGapExtensions += 1;
-                stats.alignmentScore += scoreGapExtend(scoringScheme);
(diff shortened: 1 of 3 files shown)


SoapZA:
@rrahn those templates *require* the additional arguments, hence I don't see how you could turn them into variadic templates? The reason I turned the Horizontal and Vertical Gap Open/Extend schemes into variadic templates is because their arguments aren't strictly necessary.

rrahn:
@SoapZA Basically you are right, and `setScore` even has a trailing template parameter.
So you can leave this as is. 
For the others you could add a `static_assert` to check that size of the parameter pack is exactly two, where it is required. Btw. There are more interfaces in the `score_base.h` that could be adapted.
But then again, why was the change necessary in the first place? The main reason that comes into my mind is because of a different design pattern (which I also think spares a lot of code). But, then all interfaces should be adapted such that they follow the rules of the same design pattern. Hence all getter-functions related to the score class should be adapted accordingly. Or just leave all interfaces as was :D. We only make minor patch releases for SeqAn 2 anymore, so we switched the active development to SeqAn 3 now and it is not necessary to start changing design decisions made pre-c++11. In SeqAn 3 we start with a whole new design concept based on c++17 and higher, so it will be zero backwards compatible to SeqAn 2.

SoapZA:
We have removed SeqAn as a dependency and do not need this patch merged anymore

PR #2298, opened 2018-03-20
[FIX] Fixes crash when pattern is larger than contig seq for omp-enabled

s22chan:
This fixes the crash where the region end index underflows and the child fork thinks there's still more results to report back in https://github.com/seqan/seqan/blob/584ae4fbff8312cfe31b3e0aea651edfb1b580ef/apps/razers3/razers_parallel.h#L736-L738

now it will err log 
```
ERROR: windowFindBegin() failed in thread 0
```
and report no hits. The logic can probably be further simplified for https://github.com/seqan/seqan/blob/584ae4fbff8312cfe31b3e0aea651edfb1b580ef/apps/razers3/razers_parallel.h#L664-L667

to have
```cpp
hasMore = windowFindBegin(...);
windowsDone += hasMore ? 0 : 1;
```

but I'm unsure as to whether the err has utility.

### DIFF
--- a/include/seqan/index/find_pigeonhole.h
+++ b/include/seqan/index/find_pigeonhole.h
@@ -1115,7 +1115,7 @@ windowFindNext(
     // iterate over all non-repeat regions within the window
     for (; finder.curPos < windowEnd; )
     {
-        THstkPos nonRepeatEnd = finder.endPos - length(pattern.shape) + 1;
+        THstkPos nonRepeatEnd = finder.endPos - _min(finder.endPos, length(pattern.shape) + 1);
         THstkPos localEnd = _min(windowEnd, nonRepeatEnd);
 
         // filter a non-repeat region within the window


marehr:
!rebuild

marehr:
@s22chan Thank you for your pull request.

Razers3 is currently unmaintained and we have to figure out what the underlying issue is here.

Do you know how a non-parallel execution differs in this aspect? You changed one file that seems to be unrelated to a parallel invocation, does that mean that this bug is also available in a sequential run? 

I'm not from the domain, but is it a valid use case to have shorter contig sequences than the pattern? It seems to me that the author did not take this use case into consideration and that there might be more affected places that might fail.

Would it be okay to filter-out contig sequences that are shorter than the pattern in a preprocess step? How would this change the output? Sorry again, I have no experience with razers3.

s22chan:
I'm not sure if I can make the execution path clear... as you can see from my above comment, the _mapSingleReadsParallelToContig function does not error out for windowFindBegin and logs a warning instead, because it relies on the subsequent code to fill a "no results" return value from the fork (that can be short circuited as per comments above again).

If you look at the serial version at https://github.com/seqan/seqan/blob/f1ec3e8f9e91b8c6bb7c02cdaef1fe7577f94a19/apps/razers3/razers_window.h#L133 you will see it only runs the match if a window size for the pattern can be found.

marehr:
Sry for not coming back earlier.

> Filtering ahead of time could be done without impact either, not sure where's a good place for that since this function is the filter in the serial code.

We discussed this internally and we would prefer if this case is handled out of the loop. The normal use case is that the window is big enough, s.t. the underflow will not happen and thus it would incur an unnecessary performance penalty to do a `min` in every iteration.

A potential place would be the caller of `windowFindNext` which should check it.
I would do a remark in a comment of `windowFindNext` that the function assumes longer windows than the pattern length and would add an `assert` in `windowFindNext` at the beginning of the function, to catch those errors during debug runs.

PR #2302, opened 2018-04-01
[FIX] Bidirectional FM Index: goDownString pushing too many states on…

### LINKED ISSUE #2300: Probable bug in bidirectional iterators
The implementation of goDown seems to be incorrect when the following conditions are met 

- The iterator uses an underlying bidirectional FM index 

- The thrid argument is Rev() (that is, we are going down the suffix tree, or equivalently expanding the pattern to the right)

- The second parameter is a string of many characters (as opposed to one character, or no second parameter

Consider this simple program , sorry for the formating

```c++
#include <vector>
#include <fstream>
#include <iostream>
#include <string>
#include <queue>
#include <algorithm>
#include <string.h>

#include <seqan/basic.h>
#include <seqan/sequence.h>
#include <seqan/stream.h>
#include <seqan/index.h>
using namespace std;
using namespace seqan;

typedef FastFMIndexConfig<void, uint32_t> TFastConfig;
typedef Index<String<Dna5>, BidirectionalIndex<FMIndex<void, TFastConfig> > > TIndex;
typedef Iter<TIndex, VSTree<TopDown<ParentLinks<> > > > MyiterT;

int main (int argc, const char* argv[]) {
    TIndex index;
    open(index, "XTEST");
   
    MyiterT it(index);
    
    goDown(it, Dna5String("AAAAT"), Rev());
    cerr << representative(it) << endl;
    goDown(it, Dna5String("CG"), Rev());
    cerr << representative(it) << endl;
    goUp(it);
    cerr << representative(it) << endl;
    goUp(it);
    cerr  << representative(it) << endl;
    goUp(it);
    cerr << representative(it) << endl;

    return 0;
}
```

The output in my case is 

AAAAT
AAAATCG
AAAAT
AAAAT

The fourth string should be AAA since it is the next node up in the suffix tree, and this is the problem.

This problem does not happen when goingDown by a single letter, or with the Fwd() tag as the third parameter. ~~Furtheremore, is seems that goUp can go several edges back, based on history as evidenced by the third line (two letters in this case). Is this also a bug, since the same does not happen with unidirectional iterators ?~~

cpockrandt:
…to ParentLinks iterator

fixes #2300

### DIFF
--- a/include/seqan/index/index_esa_stree.h
+++ b/include/seqan/index/index_esa_stree.h
@@ -1999,7 +1999,7 @@ SEQAN_CONCEPT_IMPL((Index<TText, IndexEsa<TSpec> > const), (StringTreeConcept));
 /*!
  * @fn TopDownHistoryIterator#goUp
  * @headerfile <seqan/index.h>
- * @brief Iterates up one edge to the parent in a tree/trie. After using goDown(iterator, string [, lcp]) on a suffix or
+ * @brief Iterates up one or multiple edges to the parent in a tree/trie. E.g., after using goDown(iterator, string [, lcp]) on a suffix or
  * prefix trie (e.g. IndexSa or FMIndex) and thus going down multiple edges at once, goUp(iterator) will go up as many
  * edges as the corresponding call of goDown(iterator, string[, lcp]) has moved down.
  *
--- a/include/seqan/index/index_esa_stree.h
+++ b/include/seqan/index/index_esa_stree.h
@@ -1999,7 +1999,9 @@ SEQAN_CONCEPT_IMPL((Index<TText, IndexEsa<TSpec> > const), (StringTreeConcept));
 /*!
  * @fn TopDownHistoryIterator#goUp
  * @headerfile <seqan/index.h>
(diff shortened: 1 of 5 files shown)


smehringer:
@cpockrandt @eseiler what's the status here? Is this only waiting for h-2's review?

include/seqan/index/index_esa_stree.h
```diff
@@ -1999,7 +1999,9 @@ SEQAN_CONCEPT_IMPL((Index<TText, IndexEsa<TSpec> > const), (StringTreeConcept));
 /*!
  * @fn TopDownHistoryIterator#goUp
  * @headerfile <seqan/index.h>
- * @brief Iterates up one edge to the parent in a tree/trie.
+ * @brief Iterates up one edge to the parent in a tree/trie. After using goDown(iterator, string [, lcp]) on a suffix or
```
smehringer:
```suggestion
 * @brief Iterates up one or multiple edges to the parent in a tree/trie. E.g., after using goDown(iterator, string [, lcp]) on a suffix or
```

PR #2305, opened 2018-04-17
Yara - fix seed anchor

temehi:
- Fixed seed and anchor checking (restores full-sensitivity)
- Removed deprecated option --all

PR #2309, opened 2018-05-07
Align secondary matches in yara

lkuchenb:
This is a new feature for yara, allowing the user to optionally have secondary (co- and suboptimal) matches aligned and the corresponding CIGARS added to the output. Previously, secondary matches were annotated only with their edit distance.

This patch

 - Adds a user parameter `-as`, `--align-suboptimal` to enable the feature
 - Adds a safe guard to prevent the parameter to be combined with `-sa omit`
 - Computes the alignments if requested
 - Outputs the CIGARS
   - If the user selected `-sa record`, the CIGARS are simply added to the records
   - If the user selected `-sa tag`, the `XA` tag is set in the same format as `bwa mem`: `XA:Z:{ref}:{strand}{beginPos}:{cigar}:{edit distance}`
   - If the user selected *not* to calculate the alignments, but set `-sa tag`, the standard yara `XA` tag is added: `XA:Z:{ref}:{beginPos}:{endPos}:{strand}:{edit distance}`.

I find it quite confusing, that the output option is currently called

    -sa, --secondary-alignments STRING
          Specify whether to output secondary alignments in the XA tag of the primary alignment, as separate secondary
          records, or to omit them. One of tag, record, and omit. Default: tag.

because alignments were never calculated up to now. This becomes even more confusing now that there is the option to actually calculate and output them. I would therefore suggest an interface change and change this to

    -sm, --secondary-matches STRING
          Specify whether to output secondary matches in the XA tag of the primary alignment, as separate secondary
          records, or to omit them. One of tag, record, and omit. Default: tag.

and to keep a hidden option `-sa` with the same effect to maintain backwards compatibility. This is not (yet) part of this patch.

Please review the attached code, I'm also happy to discuss the interface changes and choice of output format.


lkuchenb:
@esiragusa Could you maybe review the changes and give some feedback?

esiragusa:
@lkuchenb Thanks for the PR. I am going through it finally.

Just to clarify, I explicitly avoided to align secondary mappings because that is always a waste of computation and that is prone to be misused. I couldn't come up with any use case requiring a priori bulk alignment of all mappings. Usually one wants to count mappings per read or look at one single alignment per read once it has been disambiguated. In the latter case, one would compute single alignments after disambiguation. But as you already did the work, let's merge this anyway. 

Also, the SAM terminology of alignment vs. map (that is match) seems wrong to me. The two terms are used interchangeably, however Alignment > Map. The [SAM spec](https://samtools.github.io/hts-specs/SAMv1.pdf) refers only to secondary alignments but for consistency we could speak of secondary records and secondary mappings/matches. As you probably noticed, the htslib implementation (newer samtools) now forbids `*` in the CIGAR field so that simpler mappings cannot be represented anymore in SAM, even though SAM = Sequence Alignment/Map format...

Okay, now some general comments for your PR:
- [ ] It is fine to rename `--secondary-alignments` as `--secondary-matches` (or `--secondary-mappings`).
- [ ] Please rename suboptimal as secondary,  both the command-line option ```--align-suboptimal``` as ```--align-secondary``` and all new symbols in the source code.
- [ ] Before merging, please remember to bump the minor app version [here](https://github.com/seqan/seqan/blob/master/apps/yara/CMakeLists.txt#L49).
- [ ] Also, please find a way to cover secondary alignments in the app test.

As for the implementation, here is a diagram to visualize matches management in Yara:

```
             matchesByCoord +------------> matchesSetByCoord
                   +               +
                   |               +
Permutation View   |               +
                 ++|        Bucket by SeqID
Sorted by errors   |               +
                   |               +                             +------> optimalMatchesSet
                   v               +                             |
             matchesByErrors +-----------> matchesSetByErrors ++++++ Segment Views
                   +                                             |
                   |                                             +------> suboptimalMatchesSet
  Position View  ++|              1:1
                   |               +
                   v               +
             primaryMatches <------------> cigarSet
```

So far Yara aligns only one match per read. Each element in `cigarSet` correspond to exactly one match in `primaryMatches`. The aligner goes over `primaryMatches` and fills `cigarSet`.

Note that all matches = co-optimal matches U suboptimal matches (Yara maintains prefix/suffix views on matches sorted by errors). A primary match is then picked among all matches (in single-end mode a random co-optimal match is picked). If you want to align also secondary matches, then you have to process all matches, not just suboptimal matches.

To align all matches, my first thought would be to have `cigarSet <--> matchesByErrors` and let the aligner go over `matchesByErrors`, so that the aligner fills either one match or all matches. Then the writer would have to bucket cigarSet by read to access alignments. I am assuming that ```matchesByErrors = concat(matchesSetByErrors)```. I hope I am not missing anything here.

I see that you introduced `primaryCigars` and `primaryCigarPositions` as a view of `cigarSet` for primary matches. You let the aligner go over `concat(suboptimalMatchesSet)` and fill `cigarSet`, but that concat doesn't contain all matches, right? Then you find `primaryCigars` among `cigarSet`, but that won't contain any primary cigars. Hmm... I am puzzled. Have you changed `suboptimalMatchesSet`? 

If I am getting it right, `suboptimalCigars` buckets the whole `cigarSet` by read. You added `suboptimalCigarsGlue` and `suboptimalCigarsGluePositions` as a workaround to have `suboptimalCigars[readID]`, correct?  Now, if you define `suboptimalCigars` as a `Segment StringSet` over `cigarSet`, you will call `value(suboptimalCigars, readID)` [here](https://github.com/seqan/seqan/blob/master/include/seqan/sequence/string_set_segment.h#L486). That converts `TPos pos` using the stored positions of type `TPositions` and takes an infix over the host `cigarSet` (which is another `Segment StringSet`). Your `TPos` is a number and `TPositions` should be a `String<Pair<>>` if [this](https://github.com/seqan/seqan/blob/master/include/seqan/sequence/string_set_segment.h#L131) metafunction being is used. That should work as expected, unless the metafunction is not picked because an ugly const subclassing `struct StringSetPositions<StringSet<THost, TSpec> const>` is missing?

lkuchenb:
@esiragusa From what I can tell the data structure `suboptimalMatchesSet` contains also the primary matches. When writing the secondary matches, you search for the primary one and skip it ([here](https://github.com/seqan/seqan/blob/develop/apps/yara/mapper_writer.h#L404)). Am I missing something here?

esiragusa:
Okay that makes sense now 🎉 I felt like I was looking at the wrong diff :trollface:

So suboptimalMatchesSet keeps all matches within best+x strata (it trims elements only from the back). The name suboptimalMatchesSet is misleading - sorry about that. If you want, you can rename it as outputMatchesSet or validMatchesSet or similar.

As for managing positions of primaryMatches within suboptimalMatchesSet buckets, I believe I could have avoided finding again the positions in the writer. Feel free to refactor that if you like. Two independent comments on that:
- Now that we have primaryCigarPositions, is it possible (and convenient) to recycle it as primaryPositions both for matches and cigars?
- primaryMatches indexes point to (flat) matchesByErrors. If I am correct, we could just derive them by `positions(primaryMatches) - stringSetLimits(suboptimalMatchesSet)` (elementwise, note that stringSetLimits should have length +1). Is it the case?


Does a `struct StringSetPositions<StringSet<THost, TSpec> const>` overload solves the issue with the glue view? If that works, there should be no major questions left!

lkuchenb:
@esiragusa My main problem with `StringSet<StringSet, Segment>` is exactly that it requires `String<Pair<>>` as positions. So this won't work without glue:

```
// Bucket the suboptimal cigars equivalently to the set of suboptimal matches
assign(stringSetLimits(me.suboptimalCigars), stringSetLimits(me.suboptimalMatchesSet), Exact());
```
Of course one could start and actually calculate all the necessary pairs, I just found the glue version more efficient and easier to comprehend, except for the unnecessary layer.

StringSets with `THost=StringSet` offer two infix interfaces, one directly to the host and one element-wise infix:
```
int main(int argc, char *argv[])
{
    typedef StringSet<CharString, Owner<ConcatDirect<> > > TStringSet;
    typedef Size<TStringSet>::Type                         TSize;
    typedef Pair<TSize, TSize>                             TPair;

    TStringSet sSet;
    appendValue(sSet, "foo");
    appendValue(sSet, "bar");
    appendValue(sSet, "baz");

    // Infix Interface (1): Direct host infix
    std::cout << infix(sSet, TPair(0,1), TPair(1,2)) << std::endl;

    // Infix interface (2): Element-wise infix
    auto sSetInfix = infix(sSet, 0, 2);
    for (auto x : sSetInfix)
        std::cout << "[ " << x << " ]";
    std::cout << std::endl;
    return 0;
}
```
which produces
```
ooba
[ foo ][ bar ]
```
Your `StringSet<StringSet, Segment>` class supports only type (1). Maybe a clean solution would be to add a template parameter that can be overridden to support (2).

esiragusa:
Okay let's take a step back.

I needed to remove elements in place and in parallel from the back of bucketed matches, that is why I though of the Segment StringSet as a Concat StringSet whose strings can be resized within the limits of their fixed capacity. I wanted Segment StringSet to support the following interface:

```c++
typedef StringSet<CharString, Segment<CharString>> TSegmentStringSet;

CharString s;
TSegmentStringSet sSet;
setHost(sSet, s);
assign(s, "foobarbaz");
appendInfix(sSet, 0, 3, Exact());
appendInfix(sSet, 3, 5, Exact());
appendInfix(sSet, 6, 9, Exact());

// Infix interface (2): Element-wise infix <-- OK
auto sSetInfix = infix(sSet, 0, 2);
for (auto x : sSetInfix)
    std::cout << "[ " << x << " ]";
std::cout << std::endl;
```

Infix interface (1) works too, at least its meaningful version (1a):

```c++
typedef typename StringSetPosition<TSegmentStringSet>::Type TPair;

// Infix interface (1a): Direct host infix <-- OK
std::cout << infix(sSet, TPair(0,1), TPair(0,3)) << std::endl;

// Infix interface (1b): Direct host infix <-- BROKEN, BUT DOES IT MAKE SENSE?
std::cout << infix(sSet, TPair(0,1), TPair(1,2)) << std::endl;
```

I thought Segment StringSet was cool, so I reused it to manage seeds in reads, that is infixes within an existing Concat StringSet.

```c++
typedef StringSet<CharString, Owner<ConcatDirect<>>> TConcatStringSet;

TConcatStringSet cSet;
appendValue(cSet, "foo");
appendValue(cSet, "bar");
appendValue(cSet, "baz");

typedef StringSet<TConcatStringSet, Segment<TConcatStringSet>> TConcatStringSetBucket;
typedef typename StringSetPosition<TConcatStringSet>::Type TPair;

TConcatStringSetBucket cBucket;
setHost(cBucket, cSet);
appendInfix(cBucket, TPair(0,1), TPair(0,2), Exact());
appendInfix(cBucket, TPair(1,1), TPair(1,3), Exact());
appendInfix(cBucket, TPair(2,1), TPair(2,3), Exact());

for (auto x : cBucket)
    std::cout << "[ " << x << " ]";
std::cout << std::endl;
```
So Segment StringSet works well when its host is a String or a Concat StringSet.

Now, what we want to do is to bucket a Segment StringSet (with a String of CigarElements as host). If we don't plan to reuse existing generic algorithms on it, we can proceed without the container interface. We can simply manage bucket's limits ourselves and use interface (2) in the writer:

```c++
_writeMappedRead(...)
{
...
TMatches const & matches = me.matchesSet[readId];
TCigars const & cigars = infix(me.cigarsSet,
                               me.cigarsSet[me.cigarsLimits[readId]],
                               me.cigarsSet[me.cigarsLimits[readId + 1]]);
...
_writeSecondary(me, matches, cigars, ...);
}
```
That is not very neat, however it removes the glueing workaround.

The above is exactly what the Concat StringSet does (but we don't need the rest of the StringSet interface). The Concat StringSet is supposed to own and work on a String, but we can let it bucket our container for us in this way:

```c++
typedef StringSet<CharString, Segment<CharString>> TSegmentStringSet;
typedef StringSet<TSegmentStringSet, Owner<ConcatDirect<>>> TSegmentStringSetBucket;

CharString s;
TSegmentStringSetBucket sBucket;

// Expose the concat.
TSegmentStringSet & sSet = concat(sBucket);
setHost(sSet, s);
assign(s, "foobarbaz");
appendInfix(sSet, 0, 3, Exact());
appendInfix(sSet, 3, 6, Exact());
appendInfix(sSet, 6, 9, Exact());

// Just fill the bucket limits yourself.
appendValue(stringSetLimits(sBucket), 2);
appendValue(stringSetLimits(sBucket), 3);

for (auto x : front(sBucket))
    std::cout << "[ " << x << " ]";
std::cout << std::endl;
```

Segment StringSet balks because it relies on StringSetPosition, while it would have to work on the outer Position only. Anyway, here we need the Concat functionality.

@h-2 @rrahn @marehr The StringSet concept probably needs some thinking. Is there a way to have contiguous/non-contiguous bucketed views of containers in seqan3? Can you do that with ranges/views?

lkuchenb:
So bottom line, why not use a standard `StringSet<T, Owner<ConcatDirect<> > >` instead of a `StringSet<T, Segment<T> >` with glue. Makes sense, since the buckets are consecutive, I changed the code, it would be great if you could have a look to check if you see any problems.

 - [X] It is fine to rename --secondary-alignments as --secondary-matches (or --secondary-mappings).
 - [ ]  Please rename suboptimal as secondary, both the command-line option --align-suboptimal as -- align-secondary and all new symbols in the source code.<BR>_I changed the interface - about the data structures I used suboptimal because you use `suboptimalMatchesSet` - and the alignments correspond to exactly that. Do you want me to rename your data structure as well?_
- [X] Before merging, please remember to bump the minor app version here.
- [ ] Also, please find a way to cover secondary alignments in the app test.<BR>_Secondary matches are not covered in general by the existing app tests. I will look into it._

lkuchenb:
@esiragusa Also added app tests now. Would be nice if you could re-check the PR.

apps/yara/mapper.h
```diff
@@ -317,6 +324,9 @@ struct Mapper
 
     typename Traits::TCigar             cigars;
```
esiragusa: Now we can rename suboptimalMatchesSet --> matchesSet, cigars --> cigarString, cigarSet --> cigars, suboptimalCigars --> cigarsSet.

You can copy these definitions for cigar-based variables:
```
    typename Traits::TCigarString       cigarString;
    typename Traits::TCigarsSet         cigarsSet;
    typename Traits::TCigars &          cigars;
```
Note that ```cigars``` is a reference now.

apps/yara/mapper.h
```diff
@@ -223,6 +225,11 @@ struct MapperTraits
     typedef String<CigarElement<> >                                 TCigar;
     typedef StringSet<TCigar, Segment<TCigar> >                     TCigarSet;
     typedef StringSetLimits<TCigarSet>::Type                        TCigarLimits;
+    typedef Position<TCigarSet>::Type                               TCigarSetPos;
```
esiragusa: To avoid further confusion, please rename cigar-based types as follows:

```
    typedef String<CigarElement<> >                                 TCigarString;
    typedef StringSet<TCigarString, Segment<TCigarString> >         TCigars;
    typedef StringSetLimits<TCigars>::Type                          TCigarsLimits;
    typedef Position<TCigars>::Type                                 TCigarsPos;
    typedef String<TCigarsPos>                                      TCigarsPositions;
    typedef ModifiedString<TCigars, ModPos<TCigarsPositions> >      TCigarsView;
    typedef StringSet<TCigars, Owner<ConcatDirect<> > >             TCigarsSet;
```

esiragusa:
Damn the comments got shuffled...

apps/yara/mapper.h
```diff
@@ -1161,26 +1171,101 @@ inline void _verifyMatchesImpl(Mapper<TSpec, TConfig> & me, PairedEnd)
...
+        assign(stringSetLimits(me.suboptimalCigars), stringSetLimits(me.suboptimalMatchesSet), Exact());
+
+        // Find primary cigars
+        resize(me.primaryCigarPositions, length(me.primaryMatchesPositions));
+        iterate(me.primaryMatches, [&](TMatchesViewViewIt it)
+            {
+                unsigned readId = position(it);
+                if (!isValid(*it))
+                {
+                    me.primaryCigarPositions[readId] = -1u;
```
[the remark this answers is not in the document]
lkuchenb: Helps debugging if that happens. I just felt it was good practice not to leave values uninitialized.

lkuchenb:
@esiragusa I added changes addressing the issues you mentioned in your revie

esiragusa:
All right @lkuchenb. Thanks a lot for making the extra changes. 🙏

I still see packPositions in the diff. Please remove that (if it is unused) and then you are good to go.

lkuchenb:
Strange, I am sure I removed it for the third time now - anyway, gone now 👍

lkuchenb:
@esiragusa Status here is still 'changes requested'

lkuchenb:
@rrahn This should be good to merge. Touches yara only. I'll PR a separate SeqAn Changelog update afterwards, which will cover #2305, #2316 and this PR.

PR #2316, opened 2018-06-17
Yara: Add option to specify strata by absolute errors

lkuchenb:
We have a use case for this where we map against multiple alternative references which are clustered beforehand, allowing a fixed number of errors N during clustering. To guarantee that we do not miss hits against then hidden members of the reference clusters, we need to map best + N errors.

@esiragusa What do you think about merging this?

### DIFF
--- a/apps/yara/mapper.cpp
+++ b/apps/yara/mapper.cpp
@@ -174,6 +174,14 @@ void setupArgumentParser(ArgumentParser & parser, Options const & options)
     setMaxValue(parser, "strata-rate", "10");
     setDefaultValue(parser, "strata-rate", 100.0 * options.strataRate);
 
+    addOption(parser, ArgParseOption("sc", "strata-count", "Consider suboptimal alignments within this absolute number \
+                                      of errors from the optimal alignment. Increase this threshold to increase \
+                                      the number of alternative alignments at the expense of runtime.",
+                                                          ArgParseOption::INTEGER));
+    setMinValue(parser, "strata-count", "0");
+    setMaxValue(parser, "strata-count", "127");
+    setDefaultValue(parser, "strata-count", 0);
+
     addOption(parser, ArgParseOption("y", "sensitivity", "Sensitivity with respect to edit distance. \
                                                           Full sensitivity guarantees to find all considered alignments \
                                                           but increases runtime, low sensitivity decreases runtime by \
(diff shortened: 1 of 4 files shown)


esiragusa:
lgtm 👍

apps/yara/mapper.cpp
```diff
@@ -174,6 +174,13 @@ void setupArgumentParser(ArgumentParser & parser, Options const & options)
     setMaxValue(parser, "strata-rate", "10");
     setDefaultValue(parser, "strata-rate", 100.0 * options.strataRate);
 
+    addOption(parser, ArgParseOption("sc", "strata-count", "Consider suboptimal alignments within this absolute number \
+                                      of errors from the optimal alignment. Increase this threshold to increase \
+                                      the number of alternative alignments at the expense of runtime.",
+                                                          ArgParseOption::INTEGER));
+    setMinValue(parser, "strata-count", "0");
+    setDefaultValue(parser, "strata-count", 0);
+
```
esiragusa: setMaxValue too?

apps/yara/misc_options.h
```diff
@@ -282,8 +282,9 @@ inline TReadSeqSize getReadIndels(TOptions const & options, TReadSeqSize readSeq
 template <typename TMatch, typename TOptions, typename TReadSeqSize>
 inline TReadSeqSize getReadStrata(TOptions const & options, TReadSeqSize readSeqLength)
 {
-    return std::min((TReadSeqSize)(readSeqLength * options.strataRate),
-                    (TReadSeqSize)MemberLimits<TMatch, Errors>::VALUE);
+    return options.strataCount != -1u ? options.strataCount :
+        std::min((TReadSeqSize)(readSeqLength * options.strataRate),
+                 (TReadSeqSize)MemberLimits<TMatch, Errors>::VALUE);
```
esiragusa: And maybe use min too to avoid overflow.

lkuchenb:
Added checks and a maximum value. I guess the only potential spot for overflows is here

https://github.com/seqan/seqan/blob/develop/apps/yara/bits_matches.h#L744

where the sum is computed, but that should be covered as both terms of the sum are restricted to 7 bits.

esiragusa:
Okay @lkuchenb 👍  Feel free to merge.

lkuchenb:
@rrahn @h-2 ping

rrahn:
lgtm

rrahn:
@lkuchenb could you please add an entry in the `CHANGELOG.rst` about the changes you made.

PR #2326, opened 2018-07-17
directly building test cases with g++ fails

marehr:
The underlying function of a test case is templatized and thus needs an explicit call `test_name<true>` to invoke.

This only happens when `SEQAN_ENABLE_TESTING` is not set.

### DIFF
--- a/include/seqan/basic/debug_test_system.h
+++ b/include/seqan/basic/debug_test_system.h
@@ -2365,7 +2365,7 @@ inline void fail()
 #define SEQAN_END_TESTSUITE \
     return 0;                                   \
     }
-#define SEQAN_CALL_TEST(test_name) do { SEQAN_TEST_ ## test_name(); } while (false)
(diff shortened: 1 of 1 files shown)


smehringer:
LGTM 
@marehr do you think we can merge this?

marehr:
I don't understand the question ;)  This is mergeable from my POV.

smehringer:
Well its been a while since you opened this PR so I was just checking. as h-2 is not reviewing seqan2 pull request anymore is there anyone else who could take a look since I'm not confident about cmake code? Otherwise if you say this is an easy and save enough change, I'll merge this :tada:

smehringer:
Is this related to any issue? Or how did you notice this?

marehr:
Simply building it  with g++ without setting any macros; just include paths. I wanted to build a test case with a vanilla g++ compiler.

Issue #2319, opened 2018-06-28
Bug in __getMinFileOffset()


serosko:
Related to #2273.
(using commit 19f556177d47f9c78c34b5bdc1dc49131791b181).
When using jumpToRegion() with BAI files created by Picard tools or samtools-0.1.7, __getMinFileOffset() return 0 for the file offset, because those indices will contain 0 for bins without any alignments (in contrast to indices created by newer versions of samtools, which contain the value of the next smaller bin instead of 0). The difference has been described/discussed in the SAM tools mailing list [https://sourceforge.net/p/samtools/mailman/message/25923233/](https://sourceforge.net/p/samtools/mailman/message/25923233/).
This behaviour leads to an error when subsequently trying to read the next record because we jumped to the beginning of the file.
A possible fix is scanning the linear index for the next lowest bin containing a value > 0 if a 0 would be returned and then, after jumping, calling read readRecord() until the desired refID has been reached or passed.

marehr:
Hi @serosko,

thank you for your bug report. I think we have to discuss this in the team. @smehringer who extensively worked on this is unfortunately on a really long vacation. We might have to postpone this issue until she is back.

I cursorily read the discussion on the mailing list and it seems that the default behaviour is indeed to do a linear scan to find the next lowest bin. This value presets a lower position of the linear scan to speed the search up. In recent version, it seems that this value is always pointing to the correct position and that this linear scan only takes `O(1)` time to find it which makes it effectively a jump.

It seems that in version 0.1.8 the change happened and the same field now contains the precomputed lower position for the linear scan to the next bin. They argue that this does not change behaviour, which I agree if you assume that this is a linear scan.

If you look at the [release date](https://github.com/samtools/samtools/releases?after=0.1.9) of samtools-0.1.7 (Nov 16, 2009), it is nearly 10years ago.

So is it really worthwhile to not assume this as a jump?

Is there a more recent version of Picard which you could use and which fixes this problem?
I think there was a specification document of the sam format, maybe we should look again if there was any mentioning of a linear scan.

serosko:
Hi, thank you for your reply.
[...]

PR #2327, opened 2018-07-17
Fix behavior of jumpToRegion with linear indices containing 0.

serosko:
Adressing issue #2319.
__getMinFileOffset() would return 0 on older BAI versions, causing
crashes when trying to call readRecord after jumping to the given
offset.


rrahn:
@smehringer can you have a look into it?

rrahn:
@smehringer could you have a look at it?

Issue #2331, opened 2018-08-06
Pairwise alignment gives wrong result for floating point score types


joergi-w:
Consider the tutorial's example for an alignment with affine gap costs. Let us use `double` instead of `int` as data type for score:
```cpp
#include <iostream>
#include <seqan/align.h>

using namespace seqan;

int main()
{
  typedef String<Dna> TSequence;
  typedef Align<TSequence, ArrayGaps> TAlign;
  
  TSequence seq1 = "AAATGACGGATTG";
  TSequence seq2 = "AGTCGGATCTACTG";
  
  TAlign align;
  resize(rows(align), 2);
  assignSource(row(align, 0), seq1);
  assignSource(row(align, 1), seq2);
  
  // here we use double now
  double score = globalAlignment(align, Score<double, Simple>(4., -2., -2., -4.), AffineGaps());
  std::cout << "Score: " << score << std::endl;
  std::cout << align << std::endl;
  
  return 0;
}
```
The result is not 16, as we would expect, but 18, and the alignment looks much different. 
In addition, if we look at the alignment and calculate the score manually, we see that the result 18 is wrong: 
8x 4 (match) + 1x -2 (mismatch) + 3x -4 (gap open) + 6x -2 (gap extension) = 32 -2 -12 -12 = 6.
```
Score: 18
      0     .    :    .    
        AAAT-GACGGAT----TG
             | |||||    ||
        ----AGTCGGATCTACTG
```
int-version for comparison
```
Score: 16
      0     .    :    .   
        AAATGACGGAT----TG
        |   | |||||    ||
        A---GTCGGATCTACTG
```

PR #2332, opened 2018-08-06
[FIX-#2331] correct infinity value in dp alignment

joergi-w:
Fix for #2331: The problem was a wrong alignment result and a wrong score, whenever TScoreValue is a float type in dp alignment computation. 

The issue was caused by the border initialization, which was aimed to be -inf/2 and implemented as std::numeric_limits::min / 2. This works fine for integer types, but for floating point types `std::numeric_limits::min` gives the closest positive value to 0. When adding gap extension cost to ≈0 in the first column and row, this leads in general to a better score in the max-function than the intended gap open cost, so all matrix values become wrong. 

I changed the implementation to `std::numeric_limits::lowest`, which does not change anything for integer types, but for floating point types it represents the smallest possible value as intended.

### DIFF
--- a/tests/align/test_alignment_dp_cell.h
+++ b/tests/align/test_alignment_dp_cell.h
@@ -77,15 +77,19 @@ void testDPCellDefaultInfinity(TGapCosts const &)
 {
     using namespace seqan;
 
-    typedef DPCell_<int, TGapCosts> TDPCell;
-    typedef DPCell_<int, TGapCosts> const TDPConstCell;
+    int    result_int       = DPCellDefaultInfinity<DPCell_<int, TGapCosts>>::VALUE;
+    int    result_int_const = DPCellDefaultInfinity<DPCell_<int, TGapCosts> const>::VALUE;
+    float  result_flt       = DPCellDefaultInfinity<DPCell_<float, TGapCosts>>::VALUE;
+    float  result_flt_const = DPCellDefaultInfinity<DPCell_<float, TGapCosts> const>::VALUE;
+    double result_dbl       = DPCellDefaultInfinity<DPCell_<double, TGapCosts>>::VALUE;
+    double result_dbl_const = DPCellDefaultInfinity<DPCell_<double, TGapCosts> const>::VALUE;
 
-    int result1 = DPCellDefaultInfinity<TDPCell>::VALUE;
-    int result2 = DPCellDefaultInfinity<TDPConstCell>::VALUE;
-
-    int test = std::numeric_limits<int>::min() / 2;
-    SEQAN_ASSERT_EQ(result1, test);
-    SEQAN_ASSERT_EQ(result2, test);
+    SEQAN_ASSERT_EQ      (result_int,       std::numeric_limits<int>::lowest() / 2);
+    SEQAN_ASSERT_EQ      (result_int_const, std::numeric_limits<int>::lowest() / 2);
+    SEQAN_ASSERT_IN_DELTA(result_flt,       std::numeric_limits<float>::lowest() / 2.0f, 1e-3);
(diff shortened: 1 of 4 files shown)


joergi-w:
@rrahn A few lines above my change there is defined 
```cpp
template <typename T>
const int DPCellDefaultInfinity<T>::VALUE = std::numeric_limits<int>::min();
```
Even if the alignment works now, should we replace the `int` type by a template here?

joergi-w:
Omg! This actually is the issue why LaRA produced crazy results all the time...

rrahn:
>@rrahn A few lines above my change there is defined
template <typename T>
const int DPCellDefaultInfinity<T>::VALUE = std::numeric_limits<int>::min();
Even if the alignment works now, should we replace the int type by a template here?

Where did you find this?

tests/align/test_align_global_alignment_specialized.h
```diff
@@ -315,4 +315,33 @@ SEQAN_DEFINE_TEST(test_align_global_alignment_score_myers_hirschberg)
     SEQAN_ASSERT_EQ(res, -8);
 }
 
+SEQAN_DEFINE_TEST(test_align_global_alignment_gotoh_score_types)
```
rrahn: Could you a) please add a general test case for the correct infinity value for floats and doubles?
See here: https://github.com/seqan/seqan/blob/5e8ca158982943fbc3398a192ece9e849c904d1f/tests/align/test_alignment_dp_cell.h#L75-L89

And could you move this test to https://github.com/seqan/seqan/blob/develop/tests/align/test_align_global_alignment.h
Basically at the end and call the test: `test_align_global_alignment_fix_floating_point_scores`

joergi-w:
> Where did you find this?

Found in file dp_cell.h in [lines 126-133](https://github.com/seqan/seqan/blob/584ae4fbff8312cfe31b3e0aea651edfb1b580ef/include/seqan/align/dp_cell.h#L126-L133). Maybe it should be changed, too.

rrahn:
Ah ok I found the line.
Actually, the default value should also use lowest and actually the type of `T`.
If you could fix that.

tests/align/test_align_global_alignment_specialized.h
joergi-w: sure!

tests/align/test_alignment_dp_cell.h
```diff
@@ -77,15 +77,19 @@ void testDPCellDefaultInfinity(TGapCosts const &)
...
+    float  result_flt_const = DPCellDefaultInfinity<DPCell_<float, TGapCosts> const>::VALUE;
+    double result_dbl       = DPCellDefaultInfinity<DPCell_<double, TGapCosts>>::VALUE;
+    double result_dbl_const = DPCellDefaultInfinity<DPCell_<double, TGapCosts> const>::VALUE;
+
+    SEQAN_ASSERT_EQ(result_int,       std::numeric_limits<int>::lowest() / 2);
+    SEQAN_ASSERT_EQ(result_int_const, std::numeric_limits<int>::lowest() / 2);
+    SEQAN_ASSERT_EQ(result_flt,       std::numeric_limits<float>::lowest() / 2.0f);
+    SEQAN_ASSERT_EQ(result_flt_const, std::numeric_limits<float>::lowest() / 2.0f);
+    SEQAN_ASSERT_EQ(result_dbl,       std::numeric_limits<double>::lowest() / 2.0);
+    SEQAN_ASSERT_EQ(result_dbl_const, std::numeric_limits<double>::lowest() / 2.0);
```
rrahn: Mhmm, shouldn't we use IN_DELTA version. Not sure, that on all platforms this will give the exact same result.

joergi-w: Yes, I agree!

Issue #2346, opened 2018-09-25
<seqan/bed_io.h> issues with parsing BED12 file


go9kata:
### parsing BED12 line, without extra data in it ('\n' vs '\t\n' line delimiter)

- [x] when reading a BED12 line, the parser fails if the line finishes with '\n' character after the blockBegins comma separated list. If in the test script one changes the following line: https://github.com/seqan/seqan/blob/584ae4fbff8312cfe31b3e0aea651edfb1b580ef/tests/bed_io/test_bed_io.cpp#L167

to:

```String<char> test = "I\t123\t456\tNAME\t3\t-\t33\t66\t255,0,0\t3\t10,11,12\t1,2,3\n";```

execution fails with UnexpectedEnd, where the following line is working fine:

```String<char> test = "I\t123\t456\tNAME\t3\t-\t33\t66\t255,0,0\t3\t10,11,12\t1,2,3\t\n";```

There is no need to introduce an empty 13th field in the BED12 file in order to progress normally.

### compatibility with the UCSC BED12 files, downloaded via the TableBrowser
- [x] if itemRgb value is missing, it is often substituted with '0'. The parser only accepts 3 integers delimited by a comma, e.g.: '0,0,0'. Maybe a detection of '0' might default to '0,0,0', instead of throwing an error. 

To test:
```String<char> test = "I\t123\t456\tNAME\t3\t-\t33\t66\t0\t3\t10,11,12\t1,2,3\t\n";```
throws an exception for BadLexicalCast.

- [x] the blockSizes and blockBegins lists might finish with a comma, e.g.:
 a file can contain a list like: 1,2,3,
the parser only works with: 1,2,3 (without trailing comma)

To test:
```String<char> test = "I\t123\t456\tNAME\t3\t-\t33\t66\t0,0,0\t3\t10,11,12,\t1,2,3,\t\n";```
throws an exception for BadLexicalCast.

smehringer:
Ho @go9kata,

thanks for reporting this and for the detailed description!

Looks like some incompatibility issue because our BED parsing is not up to date.
Our resources are mostly given to our new SeqAn3 library release so I don't know when we will have time to fix this. 
@joergi-w can you maybe have a look?

smehringer:
@joergi-w ping

smehringer:
Thanks @joergi-w  for fixing this.
@go9kata Check out the current develop branch and see if everything works fine now for your data.

go9kata:
[...]

PR #2355, opened 2018-10-24
[FIX-#2346] correct BED12 file parsing, compatible with UCSC

joergi-w:
This PR adds compatibility for BED12 files:
- allow additional comma at the end of _blockSizes_ and _blockStart_
- allow RGB field to contain a single `0`, which is interpreted as 0,0,0
- fix newline problems: Two lines were skipped, if there was no additional text after the 12th column.

### DIFF
--- a/include/seqan/bed_io/read_bed.h
+++ b/include/seqan/bed_io/read_bed.h
@@ -189,41 +189,56 @@ _readBedRecordNoData(BedRecord<Bed12> & record,
                      TForwardIter & iter,
                      CharString & buffer)
 {
-    IsNewline isNewline;
+    typedef AssertFunctor<NotFunctor<IsNewline>, ParseError, Bed> AssertNoNewline;
+
     // Read first three fields.
     _readBedRecordNoData(static_cast<BedRecord<Bed6 > &>(record), iter, buffer);
 
     // Read THICK BEGIN
     clear(buffer);
-    readUntil(buffer, iter, OrFunctor<IsTab, AssertFunctor<NotFunctor<IsNewline>, ParseError, Bed> >());
+    readUntil(buffer, iter, OrFunctor<IsTab, AssertNoNewline>());
     record.thickBegin = lexicalCast<int32_t>(buffer);
     skipOne(iter);
 
     // Read THICK END
     clear(buffer);
-    readUntil(buffer, iter, OrFunctor<IsTab, AssertFunctor<NotFunctor<IsNewline>, ParseError, Bed> >());
+    readUntil(buffer, iter, OrFunctor<IsTab, AssertNoNewline>());
     record.thickEnd = lexicalCast<int32_t>(buffer);
     skipOne(iter);
 
     // Read ITEM RGB
     clear(buffer);
-    readUntil(buffer, iter, OrFunctor<EqualsChar<','>, AssertFunctor<NotFunctor<IsNewline>, ParseError, Bed> >());
+    readUntil(buffer, iter, OrFunctor<IsTab, OrFunctor<EqualsChar<','>, AssertNoNewline> >());
     record.itemRgb.red = lexicalCast<int32_t>(buffer);
-    skipOne(iter);
+    if (value(iter) == ',')
+    {
+        skipOne(iter);
 
-    clear(buffer);
-    readUntil(buffer, iter, OrFunctor<EqualsChar<','>, AssertFunctor<NotFunctor<IsNewline>, ParseError, Bed> >());
-    record.itemRgb.green = lexicalCast<int32_t>(buffer);
-    skipOne(iter);
+        clear(buffer);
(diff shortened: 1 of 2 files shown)


include/seqan/bed_io/read_bed.h
```diff
@@ -189,74 +189,93 @@ _readBedRecordNoData(BedRecord<Bed12> & record,
...
+    else if (value(iter) == '\t' && record.itemRgb.red == 0)
+    {
+        skipOne(iter);
+        record.itemRgb.green = 0;
+        record.itemRgb.blue = 0;
+    }
+    else
+    {
+        throw ParseError("While parsing field 9 (itemRGB) in BED12 file: "
+                         "value must be either 0 or a list of three integer values r,g,b.");
```
smehringer: Empty values are not allowed in optional fields?

tests/bed_io/test_bed_io.cpp
```diff
@@ -219,6 +219,26 @@ SEQAN_DEFINE_TEST(test_bed_read_bed12_record)
     SEQAN_ASSERT_EQ(record.data, "data again!");
 }
 
+SEQAN_DEFINE_TEST(test_bed_read_bed12_ucsc_compatibility)
+{
+    // Prepare in-memory data.
+    String<char> test = "I\t123\t456\tNAME\t3\t-\t33\t66\t255,0,0\t3\t10,11,12\t1,2,3\n" // no text after 12th column
+                        "II\t999\t1000\tNAME2\t2e5\t.\t44\t55\t0\t3\t3,4,5\t4,5,6\tdata again!\n" // RGB = 0
+                        "III\t1001\t2000\tNAME3\t2e4\t.\t22\t23\t0,0,0\t3\t12,13,14,\t7,8,9,"; // additional comma
```
smehringer: So the very last line is allowed to have NO newline character?

tests/bed_io/test_bed_io.cpp
```diff
@@ -219,6 +219,26 @@ SEQAN_DEFINE_TEST(test_bed_read_bed12_record)
...
+    // Iterator to use
+    DirectionIterator<String<char>, Input>::Type iter = directionIterator(test, Input());
+
+    // The record to load into.
+    seqan::BedRecord<seqan::Bed12> record;
+    CharString buffer;
+
+    // Perform tests.
+    readRecord(record, buffer, iter, seqan::Bed());
+    readRecord(record, buffer, iter, seqan::Bed());
```
smehringer: Please assert the expected values at least for RGB in the second record (all should be 0).

include/seqan/bed_io/read_bed.h
joergi-w: For consistency, yes. All other fields must not be empty as well (otherwise lexicalCast's would fail) and up to now the field was required to contain a value. However, I understand your point. Maybe _optional_ has to be understood as extension to BED6 format, and therefore it is an all-or-nothing optionality?

include/seqan/bed_io/read_bed.h
```diff
@@ -189,74 +189,93 @@ _readBedRecordNoData(BedRecord<Bed12> & record,
...
-    if (isNewline(value(iter)))
-    {
-        skipLine(iter);
-        return;
-    }
-    skipOne(iter);
+    if (!atEnd(iter) && value(iter) == ',')
+        skipOne(iter);
+
+    if (!atEnd(iter) && value(iter) == '\t')
```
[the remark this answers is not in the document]
joergi-w: I infer from the implementation that behind the 12th column there can be text. If there is text, it is separated with another tab character, which has to be skipped here.

include/seqan/bed_io/read_bed.h
```diff
@@ -189,74 +189,93 @@ _readBedRecordNoData(BedRecord<Bed12> & record,
...
+        skipOne(iter);
+
+    // DO NOT skip line here, because the rest of the line is written into record.data!
```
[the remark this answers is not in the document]
joergi-w: Yes, but it will be skipped at a later time. The `readRecord` function contains
```cpp
readLine(record.data, iter);
```
(line 293), which stores the text behind the 12th column and skips the newline.
If we skipped the newline here, the whole next line would go to `record.data`, which is the erroneous behaviour I want to fix with this PR.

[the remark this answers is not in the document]
smehringer: OK please add this information in the comment
```suggestion
    // DO NOT skip line here, because the rest of the line (empty or not) is written
    // into record.data within the readRecord function below L:293
```

tests/bed_io/test_bed_io.cpp
joergi-w: I agree!

include/seqan/bed_io/read_bed.h
smehringer: Ok, I see. We will leave it as such.

PR #2361, opened 2018-11-20
[FIX-#2360] Initialization of iterators over empty graphs

### LINKED ISSUE #2360: Constructing an EdgeIterator on a Graph with no vertices causes segfault
I originally came across this problem when I found that calling `connectedComponents()` on a vertexless Graph caused a segfault.  I've narrowed it down to calling the constructor for an `EdgeIterator` object -- please see the attached MRE.  This breaks a "standard" loop for visiting every edge in a Graph:

```
TGraph g;
...
for (Iterator<TGraph, EdgeIterator>::Type it(g); !atEnd(it); goNext(it)) { ... }
```

(the correct behaviour on a vertexless Graph would be that the loop body executes zero times).  Having even a single vertex in the graph makes the problem go away.

If someone can suggest what "should" happen here, I could try making a fix myself, but I'm wary of making changes in this part of the code as there are many different interacting Graph and Graph iterator types.

[repro_empty_graph_iterator_crash.zip](https://github.com/seqan/seqan/files/2590389/repro_empty_graph_iterator_crash.zip)

joergi-w:
Solves a problem (#2360) with Graph Iterators, when initialized with an empty graph. 

The constructor checks, whether the vertex string is not empty before accessing the first vertex. Furthermore, the iterator of an empty graph must be always in the condition _atEnd_. 

fixes #2360

### DIFF
--- a/include/seqan/graph_types/graph_iterator_outedge.h
+++ b/include/seqan/graph_types/graph_iterator_outedge.h
@@ -92,8 +92,10 @@ public:
     Iter(TGraph_ const& _graph, TVertexDescriptor_ const v) :
         data_host(&_graph),
         data_source(v),
-        data_edge(getValue(_graph.data_vertex,v))
+        data_edge()
     {
+        if (!empty(_graph.data_vertex))
+            data_edge = getValue(_graph.data_vertex, v);
     }
 
     Iter(Iter const& _iter) :
@@ -155,8 +157,10 @@ public:
     Iter(TGraph const& _graph, TVertexDescriptor const v) :
         data_host(&_graph),
         data_source(v),
-        data_edge(getValue(_graph.data_vertex,v))
+        data_edge()
     {
+        if (!empty(_graph.data_vertex))
+            data_edge = getValue(_graph.data_vertex, v);
     }
 
     Iter(Iter const& _iter) :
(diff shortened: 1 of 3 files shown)


smehringer:
@marehr With your consent, I will merge this.

include/seqan/graph_types/graph_iterator_outedge.h
```diff
@@ -92,8 +92,10 @@ class Iter<Graph<Directed<TCargo, TGraphSpec> >, GraphIterator<InternalOutEdgeIt
     Iter(TGraph_ const& _graph, TVertexDescriptor_ const v) :
         data_host(&_graph),
         data_source(v),
-        data_edge(getValue(_graph.data_vertex,v))
+        data_edge()
     {
+        if (!empty(_graph.data_vertex))
+            data_edge = getValue(_graph.data_vertex, v);
```
rrahn: what happens if v does not exist? I would at least add an assertion for this case. Maybe thsis happens already inside getValue but I am not sure.

smehringer: I don't know what exaclt you mean. But `v` is an index (integer) so it will exist when called. The question if `_graph.data_vertex` has at least v+1 elements should not be asserted IMO. The developer has to take care of correctly using this.

joergi-w: I agree. The new commit tests for an empty graph already in EdgeIterator and the constructors of OutEdgeIterator are equipped with assertions now.

rrahn:
well, well, well. Obviously non of the tests check the ctd but for what ever reason, this breaks this PR.
I will merge it and I make a seprate PR to fix the ctd testing.

PR #2363, opened 2018-11-28
[FIX] Fixes option condition so that yara works in KNIME again.

rrahn:
With the latest version of yara, the  mapper cannot be executed within KNIME anymore.
The reason is a subtle difference in the usage of default values between the argument parser and the GKN. 
The GKN sets the default values stored in the CTD as input parameters for the tool execution. But the argument parser differentiates whether an option was set or defaulted. So while it works for the command line in KNIME the check whether both options strata-count and strata-rate were set causes an abort, even though they were intialized with their default values.

Despite this issue, I think there is a general flaw in the usage of the default values here.
Both strata-rate and strata-count are initialized with default values, 0 in particular. But for the strata-count the default value is never read, because it only tests if an option was set from the user. Second, the semantic of using the default values or setting the default values explicitly by the user differs. The first works the second causes an abort. 
So I simply remove the default values from both options and everthing should work as expected.

### DIFF
--- a/apps/yara/CMakeLists.txt
+++ b/apps/yara/CMakeLists.txt
@@ -46,7 +46,7 @@ endif (NOT BZIP2_FOUND)
 # App-Level Configuration
 # ----------------------------------------------------------------------------
 
-set (SEQAN_APP_VERSION "1.0.2")
+set (SEQAN_APP_VERSION "1.0.3")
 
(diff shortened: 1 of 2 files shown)


rrahn:
@esiragusa @temehi if you don't object until Friday I will merge as is.

Issue #2281, opened 2018-02-11
seqan2 fails to build on Linux/ia64


jduerstock:
Because ia64 uses "nop 0" rather than "nop", seqan fails to build on Linux/ia64.

From include/seqan/parallel/parallel_lock.h:
```
#else  // everything else.
    asm volatile ("nop" ::: "memory");  // default operation - does nothing => Might lead to passive spinning.
#endif
```

Is there a reason YieldProcessor() doesn't just use "{};" or "return;" rather than resorting to asm() calls when there's nothing to do?

h-2:
We had actually had successful builds on ia64 before :thinking: 

@rrahn What's up with the asm nop, why do we need it?

rrahn:
Hi, yes there is. We use a spin lock with hybrid spinning strategy (a combination of active and passive spinning). To prevent the compiler to optimise away the active spinning phase we use a instruction that does nothing but forces a tiny delay with a fixed number of idle cycles of the processor.
@jduerstock What you are referring to is actually a hard fallback, which might lead to unexpected longer waits than necessary for active spinning.
But this is only for architectures, we are not optimised for anyway. 
What I am wondering about is, why the `SSE2` macro is not detected (see below)? 
It should use the more performant `_mm_pause()` instruction here.
This should be available on IA64, shouldn't it? 
Can you please give us more detail about the architecture, the compiler and the compiler flags you used to compile SeqAn and which version of SeqAn you are referring too?
Did you use your own build infrastructure or our cmake build system?

```cpp
#elif defined(__SSE2__)  // AMD and Intel
    _mm_pause();
#else  // everything else.
    asm volatile ("nop" ::: "memory");  // default operation - does nothing => Might lead to passive spinning.
#endif
```

jduerstock:
No, IA64 does not support SSE2.

This is building under Debian sid, which uses CMake and gcc 7.3.0.  The full build log is here:

https://buildd.debian.org/status/fetch.php?pkg=seqan2&arch=ia64&ver=2.4.0%2Bdfsg-5&stamp=1518354962&raw=0

rrahn:
🤔 well, not sure why we did not see this before.
Can you make PR for IA64 with adding another macro check for IA64 before the SSE2 check adding the `nop 0` instruction.

jduerstock:
IA64 was just recently reinstated as an unofficial architecture for debian-ports with in the past couple of months, which is probably why you hadn't heard about it before now.

[...]

PR #2370, opened 2019-01-09
[Fix] Thread wait for IA64

mr-c:
Fixes https://github.com/seqan/seqan/issues/2281

### DIFF
--- a/include/seqan/parallel/parallel_lock.h
+++ b/include/seqan/parallel/parallel_lock.h
@@ -233,6 +233,8 @@ yieldProcessor()
     __asm__ __volatile__ ("or 27,27,27" ::: "memory");
 #elif defined(__SSE2__)  // AMD and Intel
     _mm_pause();
+#elif defined(__ia64__)  // IA64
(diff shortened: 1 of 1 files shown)


mr-c:
Thanks! Is this going to be an issue for Seqan3 as well? I took a quick look, but I couldn't find the equivalent code..

rrahn:
No, not at the moment. SeqAn3 is from scratch new code working with c++17 standard and upwards. So we are adding parrallel support in the near future, but will use as many features as possible from the STL. So adding an own locking mechanism will only happen, if the performance of the locks by the STL are not sufficient. Also given that there are already very good concorrent queue implementations I don't see why we need to reimplement this on our own. But that needs to be reevaluated step by step.

mr-c:
That sounds great! I'm sure that will be more portable and less maintenance work for you all.

Let me know when you're ready for the first test builds on Debian; we can upload to `experimental` and then later start with the release candidates as their own `seqan3` based packages.

PR #2373, opened 2019-02-06
[FIX] avoid ReScore for matches in T-Coffee lib files

joergi-w:
Fixes a bug in SeqAn::T-Coffee that recalculates and overwrites the scores of all matches provided in a library file. From now on, the scores of the library file will be added straight to the match graph.

### DIFF
--- a/include/seqan/graph_msa/graph_align_tcoffee_msa.h
+++ b/include/seqan/graph_msa/graph_align_tcoffee_msa.h
@@ -426,21 +426,6 @@ void globalMsaAlignment(Graph<Alignment<TStringSet, TCargo, TSpec> > & gAlign,
         }
     }
 
-    // Include a T-Coffee library
-    if (!empty(msaOpt.libfiles))
-    {
-        typedef typename Iterator<String<std::string> const, Standard>::Type TIter;
-        TIter begIt = begin(msaOpt.libfiles, Standard());
-        TIter begItEnd = end(msaOpt.libfiles, Standard());
-        for (; begIt != begItEnd; goNext(begIt))
-        {
-            std::ifstream strm_lib;
-            strm_lib.open((*begIt).c_str(), std::ios_base::in | std::ios_base::binary);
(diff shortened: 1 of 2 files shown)


joergi-w:
The windows CI fails (`No CMAKE_CXX_COMPILER could be found.`), but otherwise all tests pass.

include/seqan/graph_msa/graph_align_tcoffee_msa.h
```diff
@@ -471,17 +456,33 @@ void globalMsaAlignment(Graph<Alignment<TStringSet, TCargo, TSpec> > & gAlign,
...
 #endif
 
-    // Use these segment matches for the initial alignment graph
-    TGraph g(seqSet);
-    if (!msaOpt.rescore)
-        buildAlignmentGraph(matches, scores, g, FractionalScore());
-    else
-        buildAlignmentGraph(matches, scores, g, msaOpt.sc, ReScore());
+    // Use these segment matches for the initial alignment graph.
+    buildAlignmentGraph(matches, scores, g, FractionalScore());
```
smehringer: quick questions:
If `msaOpt.rescore` is always true, we never called `buildAlignmentGraph(matches, scores, g, FractionalScore());` but now we always do. Is that always correct? Does the rescoring has any effect if  `buildAlignmentGraph(matches, scores, g, FractionalScore());` is called afterwards anyway?

joergi-w: Good questions! 

(1) Yes, that is always correct. The function `buildAlignmentGraph(..., ReScore())` does exactly two things: `scoreMatches` and then `buildAlignmentGraph(..., FractionalScore())` (see its implementation in graph_align_tcoffee_base.h, L.262). The re-scoring I evaluate before the lib-files are read, the actual graph building is processed afterwards. 

(2) Yes, because it modifies the score values that are later used as edge weights in the graph. It ensures that e.g. there are no negative values.

PR #2374, opened 2019-02-12
[FIX] The precrement operator of SeqAn iterators doesn't statisfy the (Weakly)Incrementable concept

marehr:
This was fixed by removing const from return type of operator++.

### DIFF
--- a/include/seqan/basic/iterator_base.h
+++ b/include/seqan/basic/iterator_base.h
@@ -262,7 +262,7 @@ operator*(Iter<TContainer, TSpec> const & me)
 // ----------------------------------------------------------------------------
 
 template <typename TContainer, typename TSpec>
-inline Iter<TContainer, TSpec> const &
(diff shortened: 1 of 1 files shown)


marehr:
Not sure which is the base branch. Should I PR against master or develop?

smehringer:
I think always against develop

h-2:
Can you change this for pre-decrement, too?

marehr:
> Can you change this for pre-decrement, too?

done

marehr:
> Phew... Why do we return references at all? shouldn't preincrement return a copy? Is this function used anywhere?

If you do `++p` on an iterator, it is required that you give a reference back (see https://en.cppreference.com/w/cpp/named_req/InputIterator)

> 
> Either way, you would need to make the same change for predecrement no? As well as `+=`..

done 

> 
> Nice, `operator-` and `operator-=` are commented out laughing

yeah lucky me :grin:

PR #2377, opened 2019-03-27
Merge master to develop

rrahn:


joshuak94:
@rrahn Travis went through!

h-2:
What about the failed jenkins tests?

rrahn:
rebuild

rrahn:
rebuild

rrahn:
rebuild

rrahn:
Ok osx builds are green. Windows builds won't work, because our IT has restricted the access from external IP addresses. Which means we can't access the windows VMs from the Jenkins server 😢 
I will temporarily disable them until we have access again.

PR #2379, opened 2019-03-27
[FIX] Make `begin` SFINAE friendly for concept checks

marehr:


h-2:
I don't fully understand the problem. Why would you want to call begin on an iterator?

smehringer:
@h-2 I think the concept check on an iterator will try to evaluate begin() on an iterator and as I understand @marehr's comment, the check causes a hard compiler error right now.

h-2:
Sounds weird to me, but I'll merge it if it fixes something for someone :man_shrugging:

marehr:
> @h-2 I think the concept check on an iterator will try to evaluate begin() on an iterator and as I understand @marehr's comment, the check causes a hard compiler error right now.

Exactly, @smehringer understands me :)

PR #2383, opened 2019-06-14
[FIX] Add missing wrapper interfaces for accelerated alignment with trace

rrahn:
Currently there are no wrapper interfaces for the accelerated alignment with trace. This fixes it now.

### DIFF
--- a/include/seqan/align_parallel/parallel_align_interface.h
+++ b/include/seqan/align_parallel/parallel_align_interface.h
@@ -60,8 +60,8 @@ struct ParallelAlignmentExecutor
               typename ...TArgs>
     auto operator()(Sequential const & /*execPolicy*/,
                     TKernel && kernel,
-                    TSetH const & setH,
-                    TSetV const & setV,
+                    TSetH && setH,
+                    TSetV && setV,
                     TArgs && ...args)
     {
         SEQAN_ASSERT_EQ(length(setH), length(setV));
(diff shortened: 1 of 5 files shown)


include/seqan/align_parallel/parallel_align_interface.h
```diff
@@ -275,6 +275,38 @@ doWaveAlignment(ExecutionPolicy<WavefrontAlignment<TWaveSpec>, TVectorizationPol
  * Wrapper functions for calling globalAlignmentScore and localAlignmentScore with an ExecutionPolicy.
  * Note the parallel interfaces are documented as part of the standard documentation in seqan/align module.
  */
+template <typename TParallelPolicy, typename TVectorizationPolicy,
+       typename ...TArgs,
+       std::enable_if_t<std::is_same<TParallelPolicy, Serial>::value ||
```
smehringer:
```suggestion
       SEQAN_FUNC_ENABLE_IF<std::is_same<TParallelPolicy, Serial>::value ||
```
?

include/seqan/align_parallel/parallel_align_interface.h
```diff
@@ -275,6 +275,38 @@ doWaveAlignment(ExecutionPolicy<WavefrontAlignment<TWaveSpec>, TVectorizationPol
  * Wrapper functions for calling globalAlignmentScore and localAlignmentScore with an ExecutionPolicy.
  * Note the parallel interfaces are documented as part of the standard documentation in seqan/align module.
  */
+template <typename TParallelPolicy, typename TVectorizationPolicy,
+       typename ...TArgs,
```
smehringer: indentation :nail_care:

include/seqan/align_parallel/parallel_align_interface.h
rrahn: No, this macro only works as a function return type and placing auto there is not possible. So in order to use auto in the function return type I need to move the sfinae expression into the template argument.

PR #2386, opened 2019-07-10
[FIX] constrain complement() functions

h-2:
Does someone want to look at this or can I merge it after CI passes?

### DIFF
--- a/include/seqan/modifier/modifier_shortcuts.h
+++ b/include/seqan/modifier/modifier_shortcuts.h
@@ -200,13 +200,13 @@ typedef ReverseComplementString<Rna5String>::Type    Rna5StringReverseComplement
  */
 
 template <typename TSequence >
-inline void complement(TSequence & sequence)
+inline SEQAN_FUNC_ENABLE_IF(IsSequence<TSequence>, void) complement(TSequence & sequence)
 {
     convert(sequence, FunctorComplement<typename Value<TSequence>::Type>());
 }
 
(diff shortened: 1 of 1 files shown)


include/seqan/modifier/modifier_shortcuts.h
```diff
@@ -200,13 +200,13 @@ typedef ReverseComplementString<Rna5String>::Type    Rna5StringReverseComplement
  */
 
 template <typename TSequence >
-inline void complement(TSequence & sequence)
+inline SEQAN_FUNC_ENABLE_IF(IsSequence<TSequence>, void) complement(TSequence & sequence)
```
smehringer: Don't you want to fix the stringset overload below also?

h-2: The StringSet overload is already constrained to StringSets so it's not a problem!

smehringer: You sure that stringsets are physically restricted to contain sequences? :sweat_smile:

cpockrandt: But what about `reverseComplement(TSequence & sequence, Tag<TParallelTag> parallelTag)`? :stuck_out_tongue_closed_eyes:

h-2: What are you talking about? This is just about preventing the overload from being used for e.g. `int`.

PR #2394, opened 2019-09-24
If output is not defined segmentation fault is happens

monsanto-pinheiro:
The reason is because readFileNames[0] is not defined yet.


smehringer:
Hi @monsanto-pinheiro,

Thanks for spotting this!
Can you investigate the Travis error log? It seems that the bs_tools app test is failing.

rrahn:
the error is due to some bug in the manual build which is not related.

PR #2407, opened 2019-12-04
[INFRA] Add github workflow to deploy dox

rrahn:
This PR will deploy the documentation ~~every hour and deploy the generated html pages on our dedicated server. The correct dir to sync to must be set after the PR get's merged in order to test that the correct path is set.~~
Ok the last point is already resolved by only triggering the pipeline on push events on the develop or master branch

### DIFF
--- a/.github/workflows/deploy_dox.yml
+++ b/.github/workflows/deploy_dox.yml
@@ -1,0 +1,46 @@
+name: Workflow to deploy dox for SeqAn
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
+
+jobs:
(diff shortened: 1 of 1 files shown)


smehringer:
I am confused. Do you deploy every hour or on push events? I think the latter suffices.

rrahn:
> I am confused. Do you deploy every hour or on push events? I think the latter suffices.

The comment is outdated.

rrahn:
not related since nothing of this touches any of build steps. There is something broken in the manual test. I believe due to an outdated package or something. I'll make a fix for that soon

PR #2418, opened 2020-01-07
Small fixes for Lambda

h-2:
Two fixes:
  * the first uses explicit convert instead of static_cast. Relevant because SeqAn2 concepts only require that convert is available and not that a gapped alphabet is actually cast-able to it's non-gapped version. When combining with SeqAn3 gapped alphabets, this breaks.
* Two unused buffers prevented comparability in certain situations.


h-2:
@rrahn If you can prioritize this that would be :heart: It's just 4lines and we need it for Lambda stuff.... thanks!

PR #2420, opened 2020-01-22
[FIX] don't rely on deprecated match.qLength

h-2:
In Blast I/O the record and the match have a `.qLength` member. The one on the match has been deprecated, but internal code still (incorrectly) used it instead of the record's member.

This PR fixes this behaviour by using the record's member if it is != 0 (has been set to something by the user).


h-2:
Soooooorrryy. I wanted to merge a PR in Lambda, I clicked in the wrong windows :see_no_evil:

PR #2421, opened 2020-01-28
[FIX] Fixes #2417: Difference between SIMD and serial local alignment

### LINKED ISSUE #2417: Difference between SIMD and serial local alignment
The local alignment function in SIMD mode includes amino acids that mismatch but do not decrease the score. For example:

```
Query  1    EGLSAQADGSLRQEQVCDALRCSGPHARYTD  31 
             G S      + Q QVCDALRCSGPHARYTD
Sbjct  464  RGTSGV*PHEIEQ*QVCDALRCSGPHARYTD
```

Here, E and R score 0 in the Blossum62 matrix. Therefore, by including them the score stays the same and it is included in the alignment. This does not happen in serial mode, only in the SIMD version. It should probably be avoided to make sure that for example in case of a nucleotide mismatch score of 0 not the whole query gets aligned anyways.

rrahn:
In local alignment the comparison against 0 was fixed when
computing all traces. The rational behind this is that
whenever the score at a cell is not greater than 0 the trace
must be set to NONE to indicate the begin of the
alignment. Otherwise, a scoring matrix that evaluates to
score 0 for two residues might falsely add the substitution
to the alignment.

fixes #2417

### DIFF
--- a/include/seqan/align/dp_formula.h
+++ b/include/seqan/align/dp_formula.h
@@ -104,6 +104,25 @@ typedef Tag<RecursionDirectionZero_> RecursionDirectionZero;
 template <typename TCellTuple>
 using ExtractedScoreValueType_ = std::decay_t<decltype(_scoreOfCell(std::get<0>(std::declval<TCellTuple>())))>;
 
+// A transformation trait to select the correct _maxScore overload for checking against 0 in local alignment.
+template <typename TTraceConfig>
+struct TraceConfigForLocalAlignment_
+{
+private:
+    // If not CompleteTrace then use same type.
+    template <typename TType>
+    static auto selectType(TType) -> TType;
+
+    // In last check of local alignment always invoke SingleTrace _maxScore overload.
+    template <typename TGapsPlacement>
+    static auto selectType(TracebackOn<TracebackConfig_<CompleteTrace, TGapsPlacement>>)
+        -> TracebackOn<TracebackConfig_<SingleTrace, TGapsPlacement>>;
+
+public:
+    // The modified type
+    using Type = decltype(selectType(TTraceConfig{}));
+};
+
 // ============================================================================
 // Functions
 // ============================================================================
@@ -248,12 +267,13 @@ _computeScore(DPCell_<TScoreValue, TGapCosts> & current,
 
     if (IsLocalAlignment_<TAlgorithm>::VALUE)
     {
+        using TMaxScorePolicy = typename TraceConfigForLocalAlignment_<TTracebackConfig>::Type;
         return _maxScore(_scoreOfCell(current),
                          TraceBitMap_<TScoreValue>::NONE,
                          _scoreOfCell(current),
                          TraceBitMap_<TScoreValue>::NONE,
                          TraceBitMap_<TScoreValue>::DIAGONAL,
-                         TTracebackConfig{});
+                         TMaxScorePolicy{});
     }
     else
     {
--- a/include/seqan/align/dp_formula_affine.h
+++ b/include/seqan/align/dp_formula_affine.h
@@ -112,12 +112,13 @@ _computeScore(DPCell_<TScoreValue, AffineGaps> & current,
                     TTracebackConfig{});
     if (IsLocalAlignment_<TAlgorithm>::VALUE)
     {
+        using TMaxScorePolicy = typename TraceConfigForLocalAlignment_<TTracebackConfig>::Type;
         tmp = _maxScore(_scoreOfCell(current),
                         TraceBitMap_<TScoreValue>::NONE,
                         _scoreOfCell(current),
                         TraceBitMap_<TScoreValue>::NONE,
                         tmp,
-                        TTracebackConfig{});
+                        TMaxScorePolicy{});
     }
     // Cache score for previous vertical.
     _scoreOfCell(previousVertical) = _scoreOfCell(current);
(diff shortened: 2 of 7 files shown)


include/seqan/align/dp_formula_affine.h
```diff
@@ -112,12 +112,13 @@ _computeScore(DPCell_<TScoreValue, AffineGaps> & current,
                     TTracebackConfig{});
     if (IsLocalAlignment_<TAlgorithm>::VALUE)
     {
+        using TMaxScorePolicy = typename TraceConfigForLocalAlignment_<TTracebackConfig>::Type;
```
marehr: I hope you found all occurrences :)

So much duplicated code.

include/seqan/align/dp_formula.h
```diff
@@ -104,6 +104,25 @@ typedef Tag<RecursionDirectionZero_> RecursionDirectionZero;
...
+// A transformation trait to select the correct _maxScore overload for checking against 0 in local alignment.
+template <typename TTraceConfig>
+struct TraceConfigForLocalAlignment_
+{
+private:
+    // If not CompleteTrace then use same type.
+    template <typename TType>
+    static auto selectType(TType) -> TType;
+
+    // In last check of local alignment always invoke SingleTrace _maxScore overload.
```
marehr: I'm not completely understand this.

You change the type `TracebackOn<TracebackConfig_<CompleteTrace, TGapsPlacement>>` if it is a local alignment to `TracebackOn<TracebackConfig_<SingleTrace, TGapsPlacement>>`.

So `CompleteTrace` would result in garbage for LocalAlignments, but `SingleTrace` does the right thing.

I don't comprehend your comment:
> In last check of local alignment always invoke SingleTrace _maxScore overload.

because it suggests that this `SingleTrace` thing will only be invoked in the last check and otherwise `CompleteTrace` will be used. But, it will always be used `SingleTrace`.

I'm also not completely sure what `SingleTrace` and `CompleteTrace` mean. Has `SingleTrace` somehting to do with SIMD? Or has `CompleteTrace` something to do with ALL possible alignment trace paths?

tests/align/test_alignment_dp_formula.h
```diff
@@ -2610,7 +2610,7 @@ SEQAN_DEFINE_TEST(test_dp_formula_trace_local_linear_lower_band_direction)
         traceValue = _computeScore(activeCell, prevDiagonal, prevHorizontal, prevVertical, 'C', 'A',
                                                scoringScheme, RecursionDirectionLowerDiagonal(), TDPProfileCompleteTrace());
 
-        SEQAN_ASSERT_EQ(static_cast<int>(traceValue), static_cast<int>(TraceBitMap_<>::DIAGONAL));
+        SEQAN_ASSERT_EQ(static_cast<int>(traceValue), static_cast<int>(TraceBitMap_<>::NONE));
```
marehr: Interesting, so you always thought this is right, but you didn't think about this corner case.

tests/align/test_align_bugs.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+    seqan::String<seqan::AminoAcid> s1 = "QDPKEPCISIYCLYLPVAGLGNLRGCCLPWM**";
+    seqan::String<seqan::AminoAcid> s2 = "*SS*IVSRTPARTSPPGPGPIGGPH*TIQSVVATVGVYKGQGRNQRELMTRAYWEFLVQGK*LQFPIPST"
+                                         "TGVQRVSRTCRPRRAHADPVSVARVRPRTSKGITDLLLLNLVWLNATCPSKKLNADRDGRVTI*QARVSF"
+                                         "VIGINQTNRSTN*ERPCTTTHRIKKELSICQSSLCPGRVRFPVLSQIKPQAPLLVVPFRQFL*VSALQPY"
+                                         "FPRNPKTLVSRKLPEGSSM*RPPIASWHRL*SELGRYLIVFEPLTFALD*RKHSWQMLSQSFVLRRSKNF"
+                                         "TSNGTVRIAPVCPS*SLPRAPKTNKIEPRSYSIIPCTIIQAREPALNTLIFSK*TFRPPPTLSQEHRRRT"
+                                         "GGKARTSSTRLATDRPPAPKIQLRAF*PQQLKYILLELELPRLLAPDLPSNGYSLKLLKCTHSNYRASNE"
+                                         "SCIVIFRHYLPASGMGNLRACCLPWMW*PFLRLPLRNRTLIPRHP*EPR*ANTVPTKVDRADT*MIRRRC"
+                                         "*TVRSAKLSRVTKANAPEDAAGFWSDKCT";
```
marehr: Isn't there a much more minimal example? I don't understand what should be right and what is wrong with this example.

include/seqan/align/dp_formula_affine.h
rrahn: yes I know. But I did not want to start refactoring this now.

include/seqan/align/dp_formula.h
rrahn: Yeah, let's talk about it in the office.

tests/align/test_align_bugs.cpp
rrahn: This was a reported bug and the offending code which I used to test and fix.

tests/align/test_alignment_dp_formula.h
rrahn: Yes, I thought that would be the thing you want. But you don't

marehr:
@smehringer, @rrahn Can this be merged?

PR #2428, opened 2020-03-31
Add TSV file support and a demo

weese:
Hey! I needed to modify huge amounts of TSV files and as SeqAn is the one and only library to do this efficiently I wanted to contribute some code 😊

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,11 @@ dox/html
 manual/build
 sandbox
 !sandbox/CMakeLists.txt
+CMakeFiles
+CTestTestfile.cmake
+cmake_install.cmake
+Makefile
+.vscode
 *.pyc
 *~
 # Vim temp
--- a/include/seqan/tsv_io/read_tsv.h
+++ b/include/seqan/tsv_io/read_tsv.h
@@ -98,7 +98,7 @@ readRecord(TsvRecord & record,
     do
     {
         readLine(buffer, iter);
-    } while (empty(buffer) && !atEnd(iter));
+    } while (empty(buffer) && !atEnd(iter));    // skip empty lines
 
     strSplit(record, buffer, IsTab(), false);
 }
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -3,6 +3,11 @@ dox/html
 manual/build
 sandbox
 !sandbox/CMakeLists.txt
+CMakeFiles
+CTestTestfile.cmake
+cmake_install.cmake
+Makefile
+.vscode
```
rrahn, line 10: That looks like a fixup of in-source builds. Can you please remove this but keep the `.vscode` and make it a separate commit.

include/seqan/tsv_io/read_tsv.h
```diff
@@ -0,0 +1,108 @@
...
+// ============================================================================
+
+// ----------------------------------------------------------------------------
+// Tag Tsv
+// ----------------------------------------------------------------------------
+
+/*!
+ * @tag FileFormats#Tsv
+ * @headerfile <seqan/tsv_io.h>
+ * @brief Variant callinf format file.
```
rrahn, line 51:
```suggestion
 * @brief Variant calling format file.
```
Why is this bound to variant calling? 
As far as I understand it merely reads in tab-separated text files. 
Accordingly I would put this all into a text_io module, from which tsv is one supported format. Csv could be another. Similar to seq_io.

include/seqan/tsv_io/tsv_record.h
```diff
@@ -0,0 +1,62 @@
...
+
+// ----------------------------------------------------------------------------
+// Class TsvRecord
+// ----------------------------------------------------------------------------
+//
+
+/*!
+ * @class TsvRecord
+ * @implements FormattedFileHeaderConcept
+ * @headerfile <seqan/vcf_io.h>
```
rrahn, line 52: copy n paste errors

.gitignore
rrahn: well true, but `.vscode` is a common hidden directory for visual studio code which gains popularity. So I wouldn't mind but I have no strong feelings either.

PR #2433, opened 2020-06-14
[FIX] UB reverse on packed strings

cpockrandt:
If a ConcatDirect StringSet with packed strings is reversed (and OpenMP supported), reversing the strings is currently undefined behavior.
This is critical, because reversing the strings will use the parallel implementation by default.
These types of string sets are often used for string indices with a low memory footprint.

### DIFF
--- a/include/seqan/modifier/modifier_reverse.h
+++ b/include/seqan/modifier/modifier_reverse.h
@@ -591,6 +591,22 @@ reverse(StringSet<TSequence, TSpec> & stringSet, Tag<TParallelTag>)
         reverse(stringSet[seqNo], Serial());
 }
 
+// Reversing a Packed ConcatDirect StringSet in parallel is undefined behaviour
+template < typename TAlphabet, typename TAlloc, typename TSpec, typename TParallelTag>
+inline SEQAN_FUNC_ENABLE_IF(IsSameType<Tag<TParallelTag>, Parallel>, void)
+reverse(StringSet<String<TAlphabet, Packed<TAlloc> >, Owner<ConcatDirect<TSpec> > > & stringSet, Tag<TParallelTag>)
+{
+    reverse(stringSet, Serial());
+}
+
+// Reversing a Packed ConcatDirect StringSet in parallel is undefined behaviour
+template < typename TAlphabet, typename TAlloc, typename TSpec, typename TParallelTag>
+inline SEQAN_FUNC_ENABLE_IF(IsSameType<Tag<TParallelTag>, Parallel>, void)
+reverse(StringSet<String<TAlphabet, Packed<TAlloc> >, Owner<ConcatDirect<TSpec> > > const & stringSet, Tag<TParallelTag>)
+{
+    reverse(stringSet, Serial());
+}
+
 template < typename TText >
 inline SEQAN_FUNC_DISABLE_IF(Is<StlContainerConcept<TText> >)
 reverse(TText & text)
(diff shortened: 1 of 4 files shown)


marehr:
Can you rebase onto current develop?  It seems you don't allow modification by maintainer, I tried to rebase myself, but I couldn't push.

eseiler (approved):
LGTM, @marehr did you have something to add?

include/seqan/modifier/modifier_shortcuts.h
```diff
@@ -292,7 +292,8 @@ template < typename TSequence, typename TSpec, typename TParallelTag >
 inline void reverseComplement(StringSet<TSequence, TSpec> & stringSet, Tag<TParallelTag>)
 {
     int seqCount = length(stringSet);
-    SEQAN_OMP_PRAGMA(parallel for if(IsSameType<Tag<TParallelTag>, Parallel>::VALUE))
+    SEQAN_OMP_PRAGMA(parallel for if(IsSameType<Tag<TParallelTag>, Parallel>::VALUE &&
```
marehr, line 295: Can you add a short message why we exclude `isStringSetConcatDirect` and `isStringPacked`?

include/seqan/modifier/modifier_shortcuts.h
```diff
@@ -303,7 +304,8 @@ template < typename TSequence, typename TSpec, typename TParallelTag >
 inline void reverseComplement(StringSet<TSequence, TSpec> const & stringSet, Tag<TParallelTag>)
 {
     int seqCount = length(stringSet);
-    SEQAN_OMP_PRAGMA(parallel for if(IsSameType<Tag<TParallelTag>, Parallel>::VALUE))
+    SEQAN_OMP_PRAGMA(parallel for if(IsSameType<Tag<TParallelTag>, Parallel>::VALUE &&
```
marehr, line 307: here too

include/seqan/modifier/modifier_reverse.h
```diff
@@ -586,7 +586,8 @@ reverse(StringSet<TSequence, TSpec> & stringSet, Tag<TParallelTag>)
     typedef typename MakeSigned<TPos>::Type                         TSPos;
 
     TSPos seqCount = length(stringSet);
-    SEQAN_OMP_PRAGMA(parallel for if (IsSameType<Tag<TParallelTag>, Parallel>::VALUE))
+    SEQAN_OMP_PRAGMA(parallel for if (IsSameType<Tag<TParallelTag>, Parallel>::VALUE &&
```
marehr, line 589: here too

include/seqan/modifier/modifier_reverse.h
```diff
@@ -563,7 +563,7 @@ reverse(TSequence & sequence, Tag<TParallelTag> parallelTag)
 
     // disable multi-threading if sequence is too small
     // uint64_t cast is for 8bit size types for which comparison would be always true
-    if (IsSameType<Tag<TParallelTag>, Parallel>::VALUE && _reverseDoSequential(length(sequence)))
+    if (IsSameType<Tag<TParallelTag>, Parallel>::VALUE && (_reverseDoSequential(length(sequence)) || isStringPacked<TSequence>::VALUE))
```
marehr, line 566: here too

marehr (commented):
@eseiler Not really, I would have prefered, if we had globally set the `TParallelTag` to Sequential instead of changing the if branches.

I think another viable way would have been to add specific overloads:

```c++
template < typename TAlphabet, typename TAlloc, typename TSpec >
inline void reverseComplement(StringSet<String<TAlphabet, Packed<TAlloc> >, Owner<ConcatDirect<TSpec> > > const & stringSet, Tag<Parallel>)
{
  // reason why we do this
  return reverseComplement(stringSet, Serial??);
}
```

I would have a slight preference for this way. As it keeps the standard function the same.

eseiler:
I applied suggested changes and added tests

PR #2435, opened 2020-08-05
Fixes for cpp20

eseiler:
(1) We need some specialised overloads for `String`, `StringSet` and `Segment`, because the general overload for `begin` will be ambiguous in the context `std::ranges::begin` is evaluated in (https://eel.is/c++draft/range.access.begin#2.5).

(2) Including `index.h` will cause a deprecation warning even though the PizzaChili Index is not used; I removed the deprecation warning, because there is still one in the `impl::` function that is called when the index is actually created.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,90 +1,83 @@
-sudo: false
+version: ~> 1.0
+os: linux
+dist: xenial
 language: cpp
-cache:
-  directories:
-    - $HOME/ccache
-before_install: mkdir -p ${HOME}/.local/bin/
-matrix:
+
+cache: ccache
+
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - george-edison55-precise-backports
+      - llvm-toolchain-precise-3.9
+      - llvm-toolchain-trusty-7
+    packages:
+      - libboost-dev
+      - libbz2-dev
+      - libxml2-utils
+      - zlib1g-dev
+      - cmake
+      - g++-4.9
+      - g++-5
+      - g++-7
+      - clang-3.9
+      - clang-7
+
+jobs:
   include:
-    - os: linux
-      compiler: gcc-4.9
-      addons:
-        apt:
-          sources: ['ubuntu-toolchain-r-test', 'george-edison55-precise-backports']
-          packages: ['g++-4.9', 'zlib1g-dev', 'libbz2-dev', 'libboost-dev', 'python3', 'virtualenv', 'libxml2-utils']
-      install: export CXX="g++-4.9" CC="gcc-4.9"
+    - name: "GCC 4.9"
+      env:
+        - CXX=g++-4.9
+        - CC=gcc-4.9
+    - name: "GCC 5"
+      env:
+        - CXX=g++-5
+        - CC=gcc-5
+    - name: "GCC 7"
+      env:
+        - CXX=g++-7
+        - CC=gcc-7
+    - name: "Clang 3.9"
+      env:
+        - CXX=clang++-3.9
+        - CC=clang-3.9
+    - name: "Clang 7"
+      env:
+        - CXX=clang++-7
+        - CC=clang-7
 
-    - os: linux
-      compiler: gcc-5
-      addons:
-        apt:
-          sources: ['ubuntu-toolchain-r-test', 'george-edison55-precise-backports']
-          packages: ['g++-5', 'zlib1g-dev', 'libbz2-dev', 'libboost-dev', 'python3', 'virtualenv', 'libxml2-utils']
-      install: export CXX="g++-5" CC="gcc-5"
+    # - name: "Clang 3.8"
+    #   env:
+    #     - CXX=clang++-3.8
+    #     - CC=clang-3.8
+    # - name: "OSX Clang 3.6" # currently too slow on osx
+    #   os: osx
+    #   before_install:
+    #     - brew update
+    #     - brew tap homebrew/versions
+    #     - brew install llvm36
+    #   env:
+    #     - CXX=clang++-3.6
+    #     - CC=clang-3.6
+    # - name: "OSX Clang 3.7"
+    #   os: osx
+    #   before_install:
+    #     - brew update
+    #     - brew tap homebrew/versions
+    #     - brew install llvm37
+    #   env:
+    #     - CXX=clang++-3.7
+    #     - CC=clang-3.7
 
-    - os: linux
-      compiler: gcc-7
-      addons:
-        apt:
-          sources: ['ubuntu-toolchain-r-test', 'george-edison55-precise-backports']
-          packages: ['g++-7', 'zlib1g-dev', 'libbz2-dev', 'libboost-dev', 'python3', 'virtualenv', 'libxml2-utils']
-      install: export CXX="g++-7" CC="gcc-7"
+install:
+  - pyenv global 3.7.1
+  - pip3 install setuptools wheel
+  - pip3 install -r manual/requirements.txt
+  - ccache -M 10G
 
-    # - os: linux
-    #   compiler: clang-3.8
-    #   addons:
-    #     apt:
-    #       sources: ['ubuntu-toolchain-r-test', 'george-edison55-precise-backports' , 'llvm-toolchain-precise-3.8']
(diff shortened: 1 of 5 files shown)


eseiler:
> In general this looks good. I would like to remove the generic begin/end overloads entirely. Can you check that everything else would still work?
> Thank you

If I remove the generic begin/end, I have to offer an specialisation for everything that needs begin/end (obviously).
This includes the stl stuff, otherwise `seqan::begin(std::string)`, etc. doesn't work. The stl iterators then of course need to be wrapped in a seqan::Iterator.

It seems like quite a bit of work (mainly copy/paste the begin/end everywhere) and then some special stuff for STL containers; so I'm not really sure this fits in this PR.

marehr:
@eseiler You beat me, but I had a slightly different approach :)

This could also be formulated as enable_if for pre C++20, in C++20 we need the requires, because the standard implementation in gcc decided to use `begin(auto &)`, `begin(auto const &)`,  `end(auto &)` and `end(auto const &)` for the poison pill. That means a SFINAE enable_if implementation would not be more specific/specialised than those poison pills.

> I would like to remove the generic begin/end overloads entirely.

I tried that, and we have quite some code that assumes that any non-container gives the value back as it is, i.e. `begin(6) = 6`.

Furthermore, this wouldn't be a patch release anymore, because this is a breaking change.

My approach:

```patch
----------------- include/seqan/sequence/sequence_interface.h -----------------
index 524d47a2d..40cd9720a 100644
@@ -355,6 +355,22 @@ _beginDefault(T const & me,
 // Function begin()
 // --------------------------------------------------------------------------
 
+template <typename T>
+    requires (Is<ContainerConcept<T> >::VALUE)
+inline typename Iterator<T, typename DefaultGetIteratorSpec<T>::Type>::Type
+begin(T & me)
+{
+    return begin(me, typename DefaultGetIteratorSpec<T>::Type()) ;
+}
+
+template <typename T>
+    requires (Is<ContainerConcept<T> >::VALUE)
+inline typename Iterator<T const, typename DefaultGetIteratorSpec<T>::Type>::Type
+begin(T const & me)
+{
+    return begin(me, typename DefaultGetIteratorSpec<T>::Type()) ;
+}
+
 template <typename T>
 inline typename Iterator<T, typename DefaultGetIteratorSpec<T>::Type>::Type
 begin(T & me)
@@ -490,6 +506,22 @@ _endDefault(T const & me,
 // Function end()
 // --------------------------------------------------------------------------
 
+template <typename T>
+    requires (Is<ContainerConcept<T> >::VALUE)
+inline typename Iterator<T, typename DefaultGetIteratorSpec<T>::Type>::Type
+end(T & me)
+{
+    return end(me, typename DefaultGetIteratorSpec<T>::Type()) ;
+}
+
+template <typename T>
+    requires (Is<ContainerConcept<T> >::VALUE)
+inline typename Iterator<T const, typename DefaultGetIteratorSpec<T>::Type>::Type
+end(T const & me)
+{
+    return end(me, typename DefaultGetIteratorSpec<T>::Type()) ;
+}
+
 template <typename T>
 inline typename Iterator<T, typename DefaultGetIteratorSpec<T>::Type>::Type
 end(T & me)
```

marehr:
> (2) Including `index.h` will cause a deprecation warning even though the PizzaChili Index is not used; I removed the deprecation warning, because there is still one in the `impl::` function that is called when the index is actually created.

This is weird. So does this mean that PizzaChili isn't unused or does it mean that some backwards-compatibility allows the usage of PizzaChili indices? I personally would want to keep deprecation warnings and figure out why we still use them ourselves.

eseiler:
> > (2) Including `index.h` will cause a deprecation warning even though the PizzaChili Index is not used; I removed the deprecation warning, because there is still one in the `impl::` function that is called when the index is actually created.
> 
> This is weird. So does this mean that PizzaChili isn't unused or does it mean that some backwards-compatibility allows the usage of PizzaChili indices? I personally would want to keep deprecation warnings and figure out why we still use them ourselves.

you write
```
#include <seqan/index.h>

int main()
{}
```

And get a deprecation warning (error).

rrahn:
> I tried that, and we have quite some code that assumes that any non-container gives the value back as it is, i.e. begin(6) = 6.
> Furthermore, this wouldn't be a patch release anymore, because this is a breaking change.

Well, we haven't adopted any semantic versioning for SeqAn2 yet. So basically, that wouldn't matter. But that we have code, that assumes begin on something not range like is the entity is so weird and 🦀. 

To your approach. We discussed that as well. I was against it, because that means we need to add proper platform support for compilers that understand requires clauses etc. That means tinkering with the SeqAn cmake build system. That means to much effort for to less benefit. I suggest that with the SeqAn3 release, we will make sure that SeqAn2 compiles and runs with the three latest compilers on clang, gcc, and msvc. So that people using SeqAn3 and SeqAn2 in parallel can use the modern compilers. Then we can also pop out the next SeqAn2 release.

marehr:
> To your approach. We discussed that as well. I was against it, because that means we need to add proper platform support for compilers that understand requires clauses etc. That means tinkering with the SeqAn cmake build system.

Hmm, wouldn't this work without any CMAKE?

```cpp
#if defined(__cpp_lib_ranges)
begin_with_requires
#else
begin_with_sfinae
#endif
```

This does not seem to be too hard, or?

marehr:
> > > (2) Including `index.h` will cause a deprecation warning even though the PizzaChili Index is not used; I removed the deprecation warning, because there is still one in the `impl::` function that is called when the index is actually created.
> > 
> > 
> > This is weird. So does this mean that PizzaChili isn't unused or does it mean that some backwards-compatibility allows the usage of PizzaChili indices? I personally would want to keep deprecation warnings and figure out why we still use them ourselves.
> 
> you write
> 
> ```c++
> #include <seqan/index.h>
> 
> int main()
> {}
> ```
> 
> And get a deprecation warning (error).

What is the compiler output?

Hmm, that means that we somewhere use it internally. Can't we either just suppress this warning, if we use it internally?

rrahn:
> This does not seem to be too hard, or?

No, it is only hard if some compiler/ranges/concepts combination would cause any trouble with it. Sorry, but how much time was spent already with the concepts support. Yet ranges-v3 and stl ranges are not compatible. And we still have to support clang, msvc etc for Seqan2. So I'd like to keep the experimental level for SeqAn2 very low to not put to much effort into it. And these three definitions added do not harm and come with the least surprise.

marehr:
> Sorry, but how much time was spent already with the concepts support.

We excessively used weird requires expressions that did not work or were not standard conform. This is our fault, not the fault of the compiler implementations. The expression is pretty simple.

> Yet ranges-v3 and stl ranges are not compatible

This is not our problem, this is a problem of ranges library. And a user who wants to use it in tandem with std ranges needs to sort that out for themselves.

> And we still have to support clang, msvc etc for Seqan2.

This will work out-of-box, because any standard implementation should only add that macro if they really support it. ™ (I know this overly optimistic :grin:) 

> And these three definitions added do not harm and come with the least surprise.

TBH I think this would be the cleanest way to solve this, but I tried this route and there are a lot more container like / range structures in SeqAn that would need 4 explicit overloads for each type. The main problem was that SeqAn defines begin for every possible type that makes it hard to find all occurrences.

This is far more work than just adding 4 overloads for begin/end and testing them in our CI Matrix. The PR as it is right now is extremely lazy, because it just fixes (some) of our use cases for seqan3, but does not solve the underlying problem that will arise with the C++20 standard library.

eseiler:
> TBH I think this would be the cleanest way to solve this, but I tried this route and there are a lot more container like / range structures in SeqAn that would need 4 explicit overloads for each type. The main problem was that SeqAn defines begin for every possible type that makes it hard to find all occurrences.

Yes, I made the same observation. And I figure that we do not use all possible types that may be used. Also, we would have to start to offer `seqan::begin` and `seqan::end` for, e.g., `std::vector` and `std::string` ... this kinda sounds like we would actually need a CPO?

> This is far more work than just adding 4 overloads for begin/end and testing them in our CI Matrix. The PR as it is right now is extremely lazy, because it just fixes (some) of our use cases for seqan3, but does not solve the underlying problem that will arise with the C++20 standard library.

Yes, like this it just causes seqan3 tests to work with g++-10 -std=c++20 and seqan2.


All in all, I would favour the `requires` solution - at least for now.

rrahn:
Ok, then let's do it as proposed. We'll see sooner or later what else surprise we have.

PR #2438, opened 2020-11-18
[FIX] T-Coffee file ending for MUM segments

joergi-w:
This PR fixes an error, where the argument parser of T-Coffee requires the file ending `.mums`, while in the implementation the ending `.mummer` was considered. 
Solution: The implementation checks for files ending with `.mums`.


smehringer:
@joergi-w Can you check why the builds are failing?

joergi-w:
There are 4 app tests failing (mason2, mgs_roi, seqcons2, fiona) on OSX high sierra.
The message is an unexpected result of wget... these errors were already present in the previous PRs.

And for the Windows tests there seems to be a general configuration problem: Python reports that a file is not found for each unit test and during build there is an unsupported option error `--parallel`. These errors were not present in the previous PR and it seems we have to update the Jenkins configuration here.

smehringer:
Alright but that is nothing related to this PR :+1:

smehringer (approved):

PR #2445, opened 2020-12-16
[INFRA] make CI more robust

eseiler:
Commit 1: Rerun failed tests. This probably circumvents the spurious fails until we find a permanent solution
Commit 2: Add a timeout to the version check. The tests do not fail if the server is down, but the version check does not time out if the server is down.


smehringer (approved):

eseiler:
I'm going to have a look at the windows failure later. The fail also seems to be spurious.

PR #2447, opened 2020-12-16
[FIX] wrong results when searching with horspool

eseiler:
This seems to fix https://github.com/seqan/product_backlog/issues/261#issuecomment-746779628

In the test, we search (internally) for `GTGGT` in `GTG`. On windows, this sometimes returns `true`.


marehr (approved):
LOL?

PR #2448, opened 2020-12-16
[INFRA] new compilers

eseiler:
This adds CI for recent compilers. We still need to decide on the build matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,67 +26,110 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Linux GCC 4.9"
-            os: ubuntu-16.04
-            cxx: "g++-4.9"
-            cc: "gcc-4.9"
-            build_type: Release
-
-          - name: "Linux GCC 5"
-            os: ubuntu-18.04
-            cxx: "g++-5"
-            cc: "gcc-5"
-            build_type: Release
-
-          - name: "Linux GCC 7"
-            os: ubuntu-18.04
-            cxx: "g++-7"
-            cc: "gcc-7"
+          - name: "gcc8 (c++17) on Linux"
+            os: ubuntu-20.04
+            cxx: "g++-8"
+            cc: "gcc-8"
             build_type: Release
             cxx_flags: "-std=c++17"
 
-          - name: "Linux GCC 9"
-            os: ubuntu-18.04
+          - name: "gcc8 (c++17) on macOS"
+            os: macos-10.15
+            cxx: "g++-8"
+            cc: "gcc-8"
+            build_type: Release
+            cxx_flags: "-std=c++17"
+
+          - name: "gcc9 (c++17) on Linux"
+            os: ubuntu-20.04
             cxx: "g++-9"
             cc: "gcc-9"
-            build: unit
             build_type: Release
             cxx_flags: "-std=c++17"
 
-          - name: "Linux Clang 3.9"
-            os: ubuntu-18.04
-            cxx: "clang++-3.9"
-            cc: "clang-3.9"
-            build_type: Release
-            cxx_flags: "-Qunused-arguments -DSEQAN_IGNORE_MISSING_OPENMP=1"
-
-          - name: "Linux Clang 7"
-            os: ubuntu-18.04
-            cxx: "clang++-7"
-            cc: "clang-7"
-            build_type: Release
-            cxx_flags: "-Qunused-arguments -DSEQAN_IGNORE_MISSING_OPENMP=1"
-
-          - name: "macOS GCC 7"
-            os: macos-10.15
-            cxx: "g++-7"
-            cc: "gcc-7"
-            build_type: Release
-            cxx_flags: "-std=c++17"
-
-          - name: "macOS GCC 9"
+          - name: "gcc9 (c++17) on macOS"
             os: macos-10.15
             cxx: "g++-9"
             cc: "gcc-9"
             build_type: Release
             cxx_flags: "-std=c++17"
 
-          - name: "macOS Clang 7"
-            os: macos-10.15
-            cxx: "clang++-7"
-            cc: "clang-7"
+          - name: "gcc9 (c++2a) on Linux"
+            os: ubuntu-20.04
+            cxx: "g++-9"
+            cc: "gcc-9"
             build_type: Release
-            cxx_flags: "-Qunused-arguments -DSEQAN_IGNORE_MISSING_OPENMP=1"
+            cxx_flags: "-std=c++2a"
+
+          - name: "gcc9 (c++2a) on macOS"
+            os: macos-10.15
+            cxx: "g++-9"
+            cc: "gcc-9"
+            build_type: Release
+            cxx_flags: "-std=c++2a"
+
+          - name: "gcc10 (c++20) on Linux"
+            os: ubuntu-20.04
+            cxx: "g++-10"
+            cc: "gcc-10"
+            build_type: Release
+            cxx_flags: "-std=c++20"
+
+          - name: "gcc10 (c++20) on macOS"
+            os: macos-10.15
+            cxx: "g++-10"
+            cc: "gcc-10"
+            build_type: Release
+            cxx_flags: "-std=c++20"
+
+          - name: "clang9 (c++17) on Linux"
+            os: ubuntu-20.04
+            cxx: "clang++-9"
+            cc: "clang-9"
+            build_type: Release
+            cxx_flags: "-std=c++17"
+
+          - name: "clang9 (c++17) on macOS"
+            os: macos-10.15
+            cxx: "clang++-9"
+            cc: "clang-9"
+            build_type: Release
+            cxx_flags: "-std=c++17"
+
+          - name: "clang9 (c++2a) on Linux"
+            os: ubuntu-20.04
+            cxx: "clang++-9"
+            cc: "clang-9"
+            build_type: Release
+            cxx_flags: "-std=c++2a"
+
+          - name: "clang9 (c++2a) on macOS"
+            os: macos-10.15
+            cxx: "clang++-9"
+            cc: "clang-9"
+            build_type: Release
+            cxx_flags: "-std=c++2a"
+
+          - name: "clang10 (c++20) on Linux"
+            os: ubuntu-20.04
+            cxx: "clang++-10"
+            cc: "clang-10"
+            build_type: Release
+            cxx_flags: "-std=c++20"
+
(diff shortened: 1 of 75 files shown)


include/seqan/basic/debug_test_system.h
```diff
@@ -164,7 +164,18 @@
...
+// gcc 9.3 -std=c++2a has ignored attribute warnings on macOS
+#if defined(__GNUC__) && defined(__APPLE__) && (__GNUC__ == 9) && __cplusplus > 201703L
+#define unused_backup __unused 
+#undef __unused
+#define __unused 
 #include <dirent.h>     // DIR
+#undef __unused
+#define __unused unused_backup
+#undef unused_backup
+#else
```
marehr, line 176:
```suggestion
#else // #if defined(__GNUC__) && defined(__APPLE__) && (__GNUC__ == 9) && __cplusplus > 201703L
```

include/seqan/basic/debug_test_system.h
```diff
@@ -164,7 +164,18 @@
...
+#include <dirent.h>     // DIR
+#endif
```
marehr, line 178:
```suggestion
#endif // #if defined(__GNUC__) && defined(__APPLE__) && (__GNUC__ == 9) && __cplusplus > 201703L
```

include/seqan/basic/debug_test_system.h
```diff
@@ -164,7 +164,18 @@
...
 #include <sys/stat.h>   // mkdir()
+// gcc 9.3 -std=c++2a has ignored attribute warnings on macOS
+#if defined(__GNUC__) && defined(__APPLE__) && (__GNUC__ == 9) && __cplusplus > 201703L
+#define unused_backup __unused 
+#undef __unused
+#define __unused 
 #include <dirent.h>     // DIR
+#undef __unused
+#define __unused unused_backup
+#undef unused_backup
```
marehr, line 175: this does not seem to work:

https://godbolt.org/z/75aT9M

I found this answer:
https://stackoverflow.com/a/1545079

This seems to work:
https://godbolt.org/z/TMd3f9

.github/workflows/ci.yml
```diff
@@ -148,16 +166,16 @@ jobs:
           if [ "$RUNNER_OS" == "Linux" ]; then
             sudo apt-get install --yes ${{ matrix.cxx }}
           else
-            if [ "${{ matrix.cxx }}" == "clang++-7" ]; then
-              cxx_macos_name="llvm@7"
+            if [[ "${{ matrix.cxx }}" == "clang++-"* ]]; then
+              cxx_macos_name=$(echo "${{ matrix.cxx }}" | sed "s/clang++-/llvm@/g")
```
marehr, line 170:
```suggestion
              cxx_macos_name=$(echo "${{ matrix.cxx }}" | sed "s/clang++-/llvm@/g") # e.g. llvm@7
```

marehr (changes requested):
Thank you for doing this :) Just some small remarls

include/seqan/basic/debug_test_system.h
eseiler: It somehow worked, even though it shouldn't :)

eseiler:
The last open thing would be to decide on the build matrix (gcc 8,9,10, (11?), clang 8,9,10, std17, std20, ...).

marehr (approved):
Thank you!

marehr:
> The last open thing would be to decide on the build matrix (gcc 8,9,10, (11?), clang 9,10,11 std17, std20, ...).

We should test the highest supported standard for each compiler.

eseiler:
So,
* gcc 8 -std=c++17
* gcc 9 -std=c++2a
* gcc 10 - std=c++20
* clang 9 -std=c++2a
* clang 10 -std=c++20
* clang 11 -std=c++20
* MSVC /std:c++latest (/std:c++20 will only be available once it is fully implemented)
* then?

And then drop 8 and test 11 once it is out

marehr:
> That means:
> 
>     * gcc 8 -std=c++17
> 
>     * gcc 9 -std=c++2a
> 
>     * gcc 10 - std=c++20
> 
>     * clang 9 -std=c++2a
> 
>     * clang 10 -std=c++20
> 
>     * clang 11 -std=c++20
> 
>     * MSVC /std:c++latest (/std:c++20 will only be available once it is fully implemented)
> 
> 
> And then drop gcc 8 and test gcc 11 once it is out.
> Is this ok?

Thank you for spelling that out. If I see the list like that, I'm wondering if we should stick to the highest stable standard.

* gcc 8 -std=c++17
* gcc 9 -std=c++17
* gcc 10 - std=c++20
* clang 9 -std=c++17
* clang 10 -std=c++20
* clang 11 -std=c++20
* MSVC /std:c++latest (/std:c++20 will only be available once it is fully implemented)

eseiler:
Then I would use clang 9 and gcc 9with both std=c++17 and c++2a. The macro thingie and `seqan::erase` only occur in std=c++2a.

marehr:
Okay :)

eseiler:
I forgot to add the cxx_flags for the std=c++20 stuff, so some stuff will fail :D
There is no brew for llvm@10 because security/maintanence reasons; we can either tap some 3rd party repo, or just remove it (clang10 macOS).

marehr:
If there is no official package, yet, don't test it ;)

eseiler:
> If there is no official package, yet, don't test it ;)

There apparently won't be. There is llvm 7,8,9,11
I'll remove it

marehr:
> > If there is no official package, yet, don't test it ;)
> 
> There apparently won't be. There is llvm 7,8,9,11
> I'll remove it

Interesting?! Kay!

include/seqan/stream/stream_base.h
```diff
@@ -206,7 +206,7 @@ guessFormatFromStream(TStream &istream, Tag<TFormat_>)
 
     // check magic header
     unsigned i;
-    for (i = 0; i != sizeof(MagicHeader<TFormat>::VALUE) / sizeof(char); ++i)
+    for (i = 0; i != sizeof(MagicHeader<TFormat>::VALUE) / (sizeof(char)); ++i)
```
marehr, line 209: [FIX] suppress Wsizeof-pointer-div

Hä? What?

include/seqan/basic/basic_type.h
```diff
@@ -499,7 +499,7 @@ _referenceCast(typename RemovePointer<T>::Type & ptr)
 template <typename T>
 struct LENGTH
 {
-    enum { VALUE = 1 };
+    static const unsigned VALUE = 1;
```
marehr, line 502: I guess there are multiple different LENGTH metafunctions

apps/fiona/CMakeLists.txt
```diff
@@ -65,6 +65,10 @@ add_executable (fiona_illumina fiona.cpp)
 
 set_target_properties (fiona_illumina PROPERTIES COMPILE_DEFINITIONS "FIONA_ILLUMINA")
 
+# Boost 1.74.0 is not compatible with cpp20
```
marehr, line 68: Could we upgrade to a newer version later?

include/seqan/map/map_base.h
```diff
@@ -43,13 +43,13 @@ namespace seqan
 // Forwards
 // ==========================================================================
 
-#if !defined(_MSC_VER)
+#if !defined(_MSC_VER) || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703)
```
marehr, line 46: 👍🏽

marehr (commented):
Some questions

include/seqan/stream/stream_base.h
eseiler: There is a warning when you do this kind of division. It's supposed to warn if you try to get the size of an array by dividing the sizeof the array by the sizeof a pointer to the array. You can silence it with parentheses...

apps/fiona/CMakeLists.txt
eseiler: We could, but there is no compatible version yet.

include/seqan/basic/basic_type.h
eseiler: I.e. I should double-check if there are different function that still use the enums?

include/seqan/stream/stream_base.h
marehr: Interesting, do you know why this is in general bad, why this is okay here?

include/seqan/basic/basic_type.h
marehr: Wo kein Ankläger, da ist kein Angeklagter. Not blocking for me.

apps/fiona/CMakeLists.txt
marehr: Ah okay, so that might be a move in the future :)

marehr (approved):

include/seqan/stream/stream_base.h
eseiler: It's only a problem if the array is passed as function parameter (which we are not doing here).

from the patch to gcc (https://patchwork.ozlabs.org/project/gcc/patch/AM4PR0701MB2162D22EC7A12EB0797E231DE41B0@AM4PR0701MB2162.eurprd07.prod.outlook.com/):
```cpp
int
f1 (int *array)
{
  int i;
  i = sizeof array / sizeof *array;		/* { dg-warning "dividing the pointer size by the element size" } */
  i += sizeof array / sizeof array[0];		/* { dg-warning "dividing the pointer size by the element size" } */
  i += sizeof(array) / sizeof(*array);		/* { dg-warning "dividing the pointer size by the element size" } */
  i += sizeof(array) / sizeof(array[0]);	/* { dg-warning "dividing the pointer size by the element size" } */
  i += (sizeof(array)) / (sizeof(array[0]));	/* { dg-warning "dividing the pointer size by the element size" } */
  i += sizeof(array) / sizeof(int);		/* { dg-warning "dividing the pointer size by the element size" } */
  i += sizeof(array) / sizeof(char);
  i += sizeof(*array) / sizeof(char);
  i += sizeof(array[0]) / sizeof(char);
  return i;
}

int
f2 (void)
{
  int array[10];
  int i;
  i = sizeof array / sizeof *array;
  i += sizeof array / sizeof array[0];
  i += sizeof(array) / sizeof(*array);
  i += sizeof(array) / sizeof(array[0]);
  i += (sizeof(array)) / (sizeof(array[0]));
  i += sizeof(array) / sizeof(int);
  i += sizeof(array) / sizeof(char);
  i += sizeof(*array) / sizeof(char);
  i += sizeof(array[0]) / sizeof(char);
  return i;
}
```

PR #2464, opened 2021-09-08
[INFRA] gcc11

eseiler:


.github/workflows/ci.yml
```diff
@@ -271,4 +285,8 @@ jobs:
       - name: Run tests
         run: |
           cd seqan-build
-          ctest . -j2 --output-on-failure --timeout 120 || ctest . -j2 --output-on-failure --timeout 120 --rerun-failed
+          ctest_args="-j2 --output-on-failure --timeout 120"
+          if [[ "${{ matrix.cxx }}" == "g++-11" && "$RUNNER_OS" == "Linux" ]]; then
+            ctest_args="${ctest_args} -E app_test_mason2"
```
eseiler, line 290: Something changed with the random output of mason2, some simulated reads start 1 position early, but otherwise everything is fine

.github/workflows/ci_win.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         include:
           - name: "Visual Studio 16 2019"
             os: windows-2019
-            toolset: '14.28'
+            toolset: '14.29'
```
eseiler, line 27: Supposedly, you are able to just say `14.xx` to use whatever is available, but it didn't work, so I need to be explicit.

tests/sequence/test_sequence.h
```diff
@@ -1061,7 +1061,14 @@ void testSequenceBack(TString & /*Tag*/)
 
     // val is a reference in contrast to the const version of back().
     TReference val = back(string);
+#if defined(__GNUC__) && (__GNUC__ == 11) // Spurious error for GCC11
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
+        val = 'A';
```
eseiler, line 1067: I tried around for a bit, couldn't replicate the warning. This seems to be bogus.

.github/workflows/ci.yml
marehr: can you move this into cmake?

apps/fiona/fiona.cpp
```diff
@@ -4408,7 +4408,9 @@ unsigned correctReads(
 
     options.timeComputeOverlapSum = 0;
     if (options.verbosity >= 1)
+    {
         std::cerr << "Searching..." << std::endl;
+    }
 	SEQAN_PROTIMESTART(search);
```
marehr, line 4414: tab and whitespace

tests/sequence/test_sequence.h
marehr: whitespaces? 8 instead of 4?

marehr (changes requested):
see comments

marehr (approved):
Thank you!

PR #2472, opened 2021-11-04
[MISC] small fixes

eseiler:


apps/yara/CMakeLists.txt
```diff
@@ -14,11 +14,6 @@ message (STATUS "Configuring apps/yara")
 # Dependencies
 # ----------------------------------------------------------------------------
 
-if (COMPILER_MSVC)
-    message (STATUS "Yara is currently not build on Microsoft Visual because it requires OpenMP 3.0.")
-    return ()
-endif ()
-
```
SGSSGene, line 21: This is not true anymore?

eseiler: I don't know, I'll see what the CI says.

It skips MSVC because of OpenMP, but a few lines later it says there is no multi-threading if there is no OpenMP. So there is either no reason to skip MSVC or it's different from what's written here.

eseiler: Yara works fine with a small hack; otherwise there are some problems with ambiguous calls
Fiona compiles fine, but 2 out of 4 tests fail

SGSSGene (approved):

PR #2476, opened 2022-02-22
[FIX] mason2 vcf parsing

### LINKED ISSUE #2475: Error Mason2 parsing VCF file v4.2 with multiple formats in the `FORMAT` field
I was trying to use variations from the GIAB dataset to simulate the reads, however, I got the following error.

```
MASON SIMULATOR
===============

____INITIALIZING______________________________________________________________

Opening reference and variants file ... OK
Initializing fragment-to-contig distribution ... OK
Initializing simulation threads ... OK
Opening output file reads/Chr21.1M.HG002.R1.fastq ... OK
Opening output file reads/Chr21.1M.HG002.R2.fastq ... OK
Opening output file sams/Chr21.1M.HG002.mates.sam... OK

____READ SIMULATION___________________________________________________________

Distributing fragments to 1 contigs (2 haplotypes each) ... OK

Simulating Reads:
/home/osboxes/seqan/include/seqan/basic/basic_exception.h:363 FAILED!  (Uncaught exception of type seqan::BadLexicalCast: Unable to convert '1:.:31:11,7:9,5:177' into unsigned int.)

stack trace:
  0                      [0x41022c]
  1                      [0x566120]
  2                      [0x56d4e6]
  3                      [0x56d531]
  4                      [0x578a78]
  5                      [0x4eb735]
  6                      [0x55976d]
  7                      [0x558936]
  8                      [0x5584df]
  9                      [0x55809d]
 10                      [0x429b2c]
 11                      [0x42a115]
 12                      [0x429150]
 13                      [0x40e728]
 14                      [0x5e063b]
 15                      [0x402901]
```
It seems to me that Mason2 is not able to correctly parse the `FORMAT` column and the `HG002` column correctly, e.g., (`INFO` ellipsed)

```
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO    FORMAT  HG002
chr21   9602367 .       GTGTC   G       50      PASS    pla...  GT:PS:DP:ADALL:AD:GQ    0/1:.:31:11,7:9,5:177
```

If necessary I can provide the list of commands to reproduce the error. 

Mason installed via conda using `conda install -c bioconda mason -y` with version:
```
mason_simulator version: 2.0.9 [e165baf]
SeqAn version: 2.4.0
```

Thank you very much for your help.

eseiler:
Resolves #2475


apps/mason2/vcf_materialization.cpp
```diff
@@ -214,6 +214,9 @@ void VcfMaterializer::init()
                 throw MasonIOException("Could not write methylation levels FAI index.");
         }
     }
+
+    if (!empty(vcfFileName) && numSeqs(faiIndex) != length(contigNames(context(vcfFileIn))))
+        throw MasonIOException("Number of records in FAI index and VCF differ.");
```
eseiler, line 219: "records" is a bad name. Would `contigs` be better?

apps/mason2/vcf_materialization.cpp
```diff
@@ -546,6 +549,8 @@ void VcfMaterializer::_appendToVariants(Variants & variants, seqan::VcfRecord co
         seqan::CharString buffer;
         smallIndel.haplotype = 0;
         for (; !atEnd(inputIter); ++inputIter)
+        {
+            if (*inputIter == ':') break;
```
smehringer, line 553:
```suggestion
            if (*inputIter == ':') break; // ignore information after the genotype
```

apps/mason2/vcf_materialization.cpp
smehringer: Does the error persist if you remove this dependency again? (With only the fix below?)

I still think from a semantic POV that they may differ.

I think it is valid to only apply mutations to a specific chromosome while still simulating from the entire reference sequence.

eseiler: Semantically, they might differ. But I have no idea where to adapt the code for that.
The code definitely expects to have the same number of contigs in FASTA and VCF; you'll go out of bounds if they are not the same.

smehringer: ah I see, then we should add this as a quick fix :+1:

.github/workflows/ci_win.yml
```diff
@@ -96,7 +103,7 @@ jobs:
       - name: Run tests
         run: |
           cd seqan-build
-          ctest . -j2 -C "${{ matrix.build_type }}" --output-on-failure --timeout 240
+          ctest . -j2 -C "${{ matrix.build_type }}" --output-on-failure --timeout 240 -E "build_manual|test_py_lib_dox"
```
eseiler, line 106: I couldn't figure out why the CI fails; it has something to do with the python setup.
It works locally, and as the doc tests are also on macOS and Linux, I just disabled it for now.

apps/mason2/vcf_materialization.cpp
```diff
@@ -214,6 +214,9 @@ void VcfMaterializer::init()
                 throw MasonIOException("Could not write methylation levels FAI index.");
         }
     }
+
+    if (!empty(vcfFileName) && numSeqs(faiIndex) != length(contigNames(context(vcfFileIn))))
+        throw MasonIOException("Number of contigs in FAI index and VCF differ.");
```
smehringer, line 219:
```suggestion
    {
        throw MasonIOException("Number of contigs in the reference FAI index and VCF file differ. "
                                "Make sure your VCF file is based on reference file " + std::string{CString(fastaFileName)});
    }
```
Not sure if the cast works as intended

smehringer (approved):
just a tiny :nail_care: to make the error message more helpful.

apps/mason2/vcf_materialization.cpp
eseiler: I used a stringstream to not cast myself into despair :)

Looks like this now:
```
____INITIALIZING______________________________________________________________

Opening reference and variants file ...[...]/build/seqan/include/seqan/basic/basic_exception.h:363 FAILED!  (Uncaught exception of type MasonIOException: The Number of contigs in FAI index and VCF differ:
    195 contig(s) in FAI Index [...]/GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta.fai
    1 contig(s) in VCF file [...]/HG002_21.vcf
Ensure the VCF file and FASTA reference contain the same contigs. Either by supplying entries for each reference in the VCF file, or by removing all contigs from the FASTA reference that are not part of the VCF.)
```

PR #2485, opened 2022-08-31
Lets see what happens if we remove this weird define

smehringer:
It might cause a problem in my app SViper but I am not sure and I don't exactly know what happens here.


eseiler:
Seems ok to remove

eseiler (approved):

PR #2492, opened 2023-05-12
fix: rename master to main

SGSSGene:
This the preparation to rename the `master` branch to `main`


eseiler (approved):
gcc12 cpp23 is fixed on develop, but will still fail here

PR #2514, opened 2024-03-27
[FIX] mason_variator: Fix StructuralVariantRecord::endPosition() for deletions.

### LINKED ISSUE #2511: mason_variator Aborted (core dumped)
Hello! 

Now I am working on the linux environment, and I download mason with mamba (as mentioned in https://bioconda.github.io/recipes/mason/README.html#package-mason)

my mason version 2.0.9 

When I was using mason_variator, I got Aborted (core dumped). 

The data I used as reference, GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta, you can get here : ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/000/001/405/GCA_000001405.15_GRCh38/seqs_for_alignment_pipelines.ucsc_ids/GCA_000001405.15_GRCh38_no_alt_analysis_set.fna.gz

My command is:
```
mason_variator -ir GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta --sv-indel-rate 0.000005 --snp-rate 0.001 --small-indel-rate 0.0001 --max-small-indel-size 50 -ov SIM03.vcf
```
The output log is:
```
mason_variator -ir GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta --sv-indel-rate 0.000005 --snp-rate 0.001 --small-indel-rate 0.0001 --max-small-indel-size 50 -ov SIM03.vcf
MASON VARIATOR
==============

__OPTIONS_____________________________________________________________________

FASTA IN             	GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta
SV SIZE TSV IN       	
VCF OUT              	SIM03.vcf
FASTA OUT            	
BREAKPOINT TSV OUT   	
METHYLATION IN FILE  	

GENERATE VAR IDS     	YES

NUM HAPLOTYPES       	1
HAPLOTYPE SEP        	"/"

SNP RATE             	0.001
SMALL INDEL RATE     	0.0001

MIN SMALL INDEL SIZE 	1
MAX SMALL INDEL SIZE 	50

SV INDEL RATE        	5e-06
SV INVERSION RATE    	1e-07
SV TRANSLOCATION RATE	1e-07
SV DUPLICATION RATE  	1e-07

MIN SV SIZE          	50
MAX SV SIZE          	1000

SIM. METHYL. LEVELS  	0
METHYLATION LEVELS
  CG  MU             	0.6
  CG  SIGMA          	0.03
  CHG MU             	0.08
  CHG SIGMA          	0.008
  CHH MU             	0.05
  CHH SIGMA          	0.005


__PREPARATION_________________________________________________________________

Loading Reference Index GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta ... FAILED (not fatal, we can just build it)
Building Index        GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta.fai ... OK
Reference Index       GCA_000001405.15_GRCh38_no_alt_analysis_set.fasta.fai ... OK (195 seqs)

__SIMULATION__________________________________________________________________


Simulation...
  chr1
  snps:                230055
  small indels:        22912
  structural variants: 1297
/home/osboxes/seqan/apps/mason2/mason_variator.cpp:1161 Assertion failed : snpRecord.getPos() != svRecord.getPos() should be true but was 0 (Should be generated non-overlapping (snp pos = 148753369, sv pos = 148753369).)

stack trace:
  0                      [0x4170ab]
  1                      [0x43286b]
  2                      [0x431c7a]
  3                      [0x430ebe]
  4                      [0x415bea]
  5                      [0x581f7b]
  6                      [0x402901]

Aborted (core dumped)
```

But the interest thing for me, if I use below command, it worked without any problem
: 
```
mason_variator -ir CA_000001405.15_GRCh38_no_alt_analysis_set.fasta --sv-indel-rate 0.0000001 --snp-rate 0.0001 --small-indel-rate 0.000001 --max-small-indel-size 6 -ov SIM1.vcf
```

Could you please give me an advice with this problem?

smehringer:
This fixes #2511 

Let's see what the tests say as it seems to be an rather invasive fix.

So the reported error in Debug mode, was that a SNP is within the region of a deletion.

```
Assertion failed : snpRecord.getPos() != svRecord.getPos() should be true but was 0 (Should be generated non-overlapping (snp pos = 148753369, sv pos = 148753369).)
```

What happens is the following: 

1) mason_variator simulates SVs
2) mason_variator simulates SNP/Indels: It goes over every position in the contig (e.g. a chromosome) and at the same time maintains the current SV of the region to avoid overlaps. It jumps to the next SV if the current position is greater than the end position of the current SV [`pos >= svRecord.endPosition()`](https://github.com/seqan/seqan/blob/main/apps/mason2/mason_variator.cpp#L630)
3) An `svRecord` of type [StructuralVariantRecord](https://github.com/seqan/seqan/blob/main/apps/mason2/genomic_variants.h#L179) stores a size. The size is negative if it is a deletion.
4) [StructuralVariantRecord::endPosition()](https://github.com/seqan/seqan/blob/main/apps/mason2/genomic_variants.cpp#L51) returns `pos + size` for deletions.. but since `size` is negative, the end position of the deletion is **BEFORE** its start position.
5) (4) leads to a bug in the while loop in (2). The while loop will to jump to the next SV because and end pos is reached already before the start of the deletion has been reached. The overlap of a SNP with the start position of an Deletion is not checked anymore.

### DIFF
--- a/apps/mason2/genomic_variants.cpp
+++ b/apps/mason2/genomic_variants.cpp
@@ -59,7 +59,7 @@ int StructuralVariantRecord::endPosition() const
             if (size > 0)
                 return pos;
             else
-                return pos + size;
+                return pos - size; // for deletions, size is negative!
         case INVERSION:
             return pos + size;
         case TRANSLOCATION:
--- a/.github/workflows/ci_win.yml
+++ b/.github/workflows/ci_win.yml
@@ -52,9 +52,17 @@ jobs:
 
       - name: Setup conda
         uses: conda-incubator/setup-miniconda@v3
+        with:
+          activate-environment: ""
+          auto-activate-base: true
+          auto-update-conda: false
 
       - name: Install conda dependencies
-        run: conda install -c conda-forge boost-cpp bzip2 libxml2 zlib
+        run: |
+          conda install -c conda-forge boost-cpp bzip2 libxml2 zlib
+          $CONDA_BASE = conda info --base
+          "$CONDA_BASE\Library\include" | Out-File -FilePath $env:GITHUB_PATH -Append
+          "$CONDA_BASE\Library\lib" | Out-File -FilePath $env:GITHUB_PATH -Append
 
       - name: Configure tests
         run: |
--- a/apps/mason2/README
+++ b/apps/mason2/README
@@ -86,12 +86,6 @@ VCF file:
 3. Reference and Contact
 ------------------------------------------------------------------------------
 
-In case of questions and problems please contact the mailing list
+In case of questions and problems please contact us on
 
-  https://lists.fu-berlin.de/listinfo/seqan-dev
-
-or file a bug at
-
-  https://trac.seqan.de/newticket
-
-
+https://github.com/seqan/seqan
--- a/include/seqan/graph_types/graph_impl_hmm.h
+++ b/include/seqan/graph_types/graph_impl_hmm.h
@@ -740,6 +740,9 @@ getTransitionProbability(Graph<Hmm<TAlphabet, TCargo, TSpec> > const& g,
     typedef Graph<Hmm<TAlphabet, TCargo, TSpec> > const TGraph;
     typedef typename EdgeDescriptor<TGraph>::Type TEdgeDescriptor;
     TEdgeDescriptor e = findEdge(g, state1, state2);
+#ifdef __INTEL_LLVM_COMPILER
+    asm volatile("" : : "r,m"(e) : "memory"); // Kindly request IntelLLVM (2024.1) to not optimize result away.
+#endif
     if (e == 0) return 0.0;
     else return cargo(e);
 }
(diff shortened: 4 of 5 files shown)


smehringer:
@eseiler looks like an unrelated failure. Could you review?

apps/mason2/genomic_variants.cpp
```diff
@@ -59,7 +59,7 @@ int StructuralVariantRecord::endPosition() const
             if (size > 0)
                 return pos;
             else
-                return pos + size;
+                return pos + std::abs(size);
```
SGSSGene, line 62: Since we know that size is negative, we could use a cheaper operation. This would also equal what happens in the `nearBreakend` function.
```suggestion
                return pos - size;
```

SGSSGene (commented):
Very nice that you also fixed the  URL to the ticket system :-)

apps/mason2/genomic_variants.cpp
smehringer: also fine for me

eseiler:
LGTM, I'll fix the unrelated stuff

PR #2515, opened 2024-04-09
[FIX] Missing SQ header in BAM output

eseiler:
The BAM output is missing the `@SQ` headers derived from the context.

According to the SAM spec, the header *should* contain the references.

When rewriting a seqan2 BAM with `samtools view -hb`, the `@SQ` lines are silently regenerated (from the `l_name`, `name`, and `l_ref` data fields in the list of reference information in the BAM file).

### DIFF
--- a/include/seqan/bam_io/write_bam.h
+++ b/include/seqan/bam_io/write_bam.h
@@ -57,9 +57,7 @@ void write(TTarget & target,
     write(target, "BAM\1");
     clear(context.buffer);
 
-    // Create text of header.
-    for (unsigned i = 0; i < length(header); ++i)
-        write(context.buffer, header[i], context, Sam());
+    write(context.buffer, header, context, Sam());
 
     // Note that we do not write out a null-character to terminate the header.  This would be valid by the SAM standard
     // but the samtools do not expect this and write out the '\0' when converting from BAM to SAM.


include/seqan/bam_io/write_bam.h
```diff
@@ -54,12 +54,41 @@ void write(TTarget & target,
...
+    for (unsigned i = 0; i < length(contigLengths(context)); ++i)
+    {
+        if (writtenSeqInfos[i])
+            continue;
+        write(context.buffer, "@SQ\tSN:");
+        write(context.buffer, contigNames(context)[i]);
+        write(context.buffer, "\tLN:");
+        appendNumber(context.buffer, contigLengths(context)[i]);
+        writeValue(context.buffer, '\n');
+    }
```
smehringer, line 91:
```suggestion
    write(target, "BAM\1");
    clear(context.buffer);
    
    write(target, header, context, Sam());
```
Can you check if this works? 
I noticed you just copied from the SAM header. Seems very redundant.

smehringer (changes requested):

include/seqan/bam_io/write_bam.h
eseiler: Yup, also works. Though it has to be `write(context.buffer` instead of `write(target`

smehringer (approved):

Issue #2520, opened 2024-06-24
mason2 assertion failures


marcelm:
The pre-compiled `mason_simulator` binary available for download from http://packages.seqan.de/mason2/ (within the tarball http://packages.seqan.de/mason2/mason2-2.0.9-Linux-x86_64_sse4.tar.xz) is a lot slower than the one compiled from sources, and it often crashes with assertion failures.

Simulating 1 million 2x100bp reads for a Drosophila genome takes about ~60 minutes using the pre-compiled binary, but only 10 minutes when using a binary compiled from sources.

Additionally, for about half of my datasets, the pre-compiled binary crashes with messages like
```
/home/osboxes/seqan/include/seqan/sequence/segment_infix.h:80 Assertion failed : data_begin_position <= data_end_position was: 7378086 > 7378081
```
This does not happen for the binary compiled from sources.

I suspect that the binary that is available for download was compiled with assertions enabled.

I noticed this because the Conda package available from Bioconda kept crashing. It turns out that the [recipe](https://github.com/bioconda/bioconda-recipes/blob/6a8cd82f05665e076534b9e1dd54197887ccf952/recipes/mason/meta.yaml) re-packages the provided binary instead of compiling from sources.

If there is interest, I’d be happy to provide a small example for reproducing this.

rrahn:
Hi @marcelm thanks for reporting this. If it would be ok, we would appreciate a minimum working example that causes theses crashes to further investigate this.

marcelm:
Ok, I suggest to focus on the assertion failures in this issue then. Using the most recent sources, here is one way to get such a an assertion failure:
```bash
cmake -DSEQAN_BUILD_SYSTEM=APP:mason2 -DCMAKE_BUILD_TYPE=Debug -B build -S seqan
cmake --build build -j 8
curl ftp://ftp.ensembl.org/pub/release-97/fasta/drosophila_melanogaster/dna/Drosophila_melanogaster.BDGP6.22.dna.toplevel.fa.gz | gunzip > drosophila.fa
build/bin/mason_variator -ir drosophila.fa --sv-indel-rate 0.00002 --snp-rate 0.005 --small-indel-rate 0.001 --max-small-indel-size 100 -ov variations.vcf
ulimit -n 16384
build/bin/mason_simulator -ir drosophila.fa -n 1000 -iv variations.vcf --illumina-read-length 100 -o 1.fastq -or 2.fastq -oa out.bam
```

Output:
```
...
[...]

PR #2521, opened 2024-06-25
[FIX] mason

eseiler:
See #2520


rrahn (approved):
Looks good to me!
Thank you!

PR #2525, opened 2024-07-22
[FIX] Segfault in GlobalAlignment

### LINKED ISSUE #2524: Unallocated memory read which causes occasional segment faults during banded alignments
I have been getting occasional segment faults in Seqan2 code when running banded alignments in the [Shasta assembler](https://github.com/paoloshasta/shasta). I distilled this down to a read of unallocated memory that occurs in the simple attached program `seqanTest1.cpp` (renamed with a `.txt` extension to permit uploading to GitHub). This test program does not crash and computes a correct alignment, however I used two different memory debugging tools to see that it indeed addresses invalid memory. This is the cause of the occasional segment faults.

I am using the latest version of Seqan2, 2.4.0, as distributed under Ubuntu 22.04 (package `libseqan2-dev` at version `2.4.0+dfsg-14ubuntu1`).

### Using valgrind
* Compile the test program using `g++ seqanTest1.cpp -ggdb3 -O0`
* Run the executable using `valgrind`: `valgrind --log-file=valgrind.log ./a.out`
* The output `valgrind.log` is attached. Here are snippets from that:

```
Invalid read of size 4
at 0x12BD7B: void seqan::_preInitCacheDiagonal
...
Address 0x4de717c is 4 bytes before a block of size 8 alloc'd
at 0x4849013: operator new
void seqan::allocate
```

### Using AddressSanitizer
* Compile the test program using `g++ seqanTest1.cpp -ggdb3 -O0 -fsanitize=address`
* Run it redirecting all output: `./a.out 1>sanitize.log 2>&1`
* The output `sanitize.log` is attached. Here are snippets from that:

```
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200000000c at pc 0x55e15f8d8e95 bp 0x7ffe43f20f40 sp 0x7ffe43f20f30
READ of size 4 at 0x60200000000c thread T0
    #0 0x55e15f8d8e94 in void seqan::_preInitCacheDiagonal
...
0x60200000000c is located 4 bytes to the left of 8-byte region [0x602000000010,0x602000000018)
allocated by thread T0 here:
    #0 0x734282cb61e7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x55e15f8f553e in void seqan::allocate
```
### Conclusion

Both `valgrind` and `AddressSanitizer` agree that the code is allocating an 8-byte block, and then addresses memory 4 bytes before the beginning of that block.

[sanitize.log](https://github.com/user-attachments/files/16336695/sanitize.log)
[seqanTest1.cpp.txt](https://github.com/user-attachments/files/16336696/seqanTest1.cpp.txt)
[valgrind.log](https://github.com/user-attachments/files/16336697/valgrind.log)

eseiler:
Resolves #2524 

We are in this case: https://github.com/seqan/seqan/blob/main/include/seqan/align/dp_algorithm_impl.h#L691

Some time later, this line will also be called on the first Column:
https://github.com/seqan/seqan/blob/45fa675f63514f1b95f746644bf5648379d0101e/include/seqan/align/dp_algorithm_impl.h#L361

However, there is no previous column.

<details><summary>Testcode (click to expand)</summary>

```cpp
#include <iostream>

#include <seqan/align.h>

int main()
{
    seqan2::String<char> const seq1{"CDFGHC"};
    seqan2::String<char> const seq2{"CDEFGAHC"};

    seqan2::Align<seqan2::String<char>, seqan2::ArrayGaps> align{};
    seqan2::resize(seqan2::rows(align), 2);
    seqan2::assignSource(seqan2::row(align, 0), seq1);
    seqan2::assignSource(seqan2::row(align, 1), seq2);

    int const score = seqan2::globalAlignment(align,
                                              seqan2::Score<int, seqan2::Simple>(0, -1, -1),
                                              seqan2::AlignConfig<true, true, true, true>(),
                                              -2,
                                              -1);
    std::cout << "Score: " << score << std::endl;
    std::cout << align << std::endl;

    return 0;
}
```
</details>

### DIFF
--- a/include/seqan/align/dp_matrix_navigator_score_matrix.h
+++ b/include/seqan/align/dp_matrix_navigator_score_matrix.h
@@ -361,7 +361,7 @@ _goNextCell(DPMatrixNavigator_<DPMatrix_<TValue, FullDPMatrix, THost>, DPScoreMa
 template <typename TDPCell,
           typename TValue, typename TMatrixSpec, typename THost,
           typename TColumnType>
-inline void
+inline std::enable_if_t<!std::is_same_v<TColumnType, DPInitialColumn>>
 _preInitCacheDiagonal(TDPCell & cacheDiagonal,
                       DPMatrixNavigator_<DPMatrix_<TValue, TMatrixSpec, THost>, DPScoreMatrix, NavigateColumnWiseBanded> const & dpNavigator,
                       MetaColumnDescriptor<TColumnType, PartialColumnMiddle> const & /*tag*/)
@@ -372,7 +372,7 @@ _preInitCacheDiagonal(TDPCell & cacheDiagonal,
 template <typename TDPCell,
           typename TValue, typename TMatrixSpec, typename THost,
           typename TColumnType>
-inline void
+inline std::enable_if_t<!std::is_same_v<TColumnType, DPInitialColumn>>
 _preInitCacheDiagonal(TDPCell & cacheDiagonal,
                       DPMatrixNavigator_<DPMatrix_<TValue, TMatrixSpec, THost>, DPScoreMatrix, NavigateColumnWiseBanded> const & dpNavigator,
                       MetaColumnDescriptor<TColumnType, PartialColumnBottom> const & /*tag*/)


rrahn (approved):
Wow, good catch! Thanks a lot!

PR #2539, opened 2024-12-04
[FIX] arg_parse: version check URL

eseiler:


rrahn:
That's why thought we should not change this behaviour within SeqAn2 if it is not already the case :). Sharg replaces all argument parsers and has the right behaviour and we do not promote the usage of SeqAn2. So just adapting the right URL would be fine I guess.

eseiler:
> That's why thought we should not change this behaviour within SeqAn2 if it is not already the case :). 

I misread that part on matrix. Will change it the pr later.

PR #2541, opened 2024-12-09
K-mer index memory error

eaasna:


eseiler (approved):
Thanks!

I only cleaned up the second commit a bit.

PR #2543, opened 2024-12-10
[FIX] mason2: Skewed simulation of N

### LINKED ISSUE #2487: BUG: simulation might produce 'N' even this is not intended
I have no clue under what circumstances this code is being called. But it is for simulation of illumina (might also exist for other simulator types).

The Problem is that It randomly generates ACGTs, and shifts the value by 1 if it simulates the same value.
The first problem is that 'T' will be shifted to 'N' (line 225).
The second problem is, this creates skewed data, which has a higher likelihood for a letter being substituted with the next higher up letter, instead with a random one.

See the lines:
https://github.com/seqan/seqan/blob/04967e0979f14694a23087d5961b01eed446cab4/apps/mason2/simulate_illumina.cpp#L213-L225

eseiler:
Resolves #2487 

| File Name | A | C | G | T | N |
|-----------|---|---|---|---|---|
| simulator.left3.fa - before | 25,116 | 24,646 | 24,932 | 25,283 | 23 |
| simulator.left3.fa - after  | 25,027 | 24,692 | 25,023 | 25,258 |   |
| simulator.left4.fa - before | 24,960 | 24,924 | 24,911 | 25,174 | 31 |
| simulator.left4.fa - after  | 25,229 | 25,160 | 24,715 | 24,896 |   |
| simulator.left7.fa - before | 25,134 | 24,801 | 25,253 | 24,778 | 34 |
| simulator.left7.fa - after  | 24,845 | 25,099 | 25,052 | 25,004 |   |
| simulator.right3.fa - before| 25,009 | 24,899 | 25,030 | 25,042 | 20 |
| simulator.right3.fa - after | 25,117 | 24,917 | 24,866 | 25,100 |   |
| simulator.left1.fq - before | 27,136 | 27,762 | 32,868 | 25,283 | 23 |
| simulator.left1.fq - after  | 27,075 | 27,723 | 33,133 | 25,258 |   |
| simulator.left2.fq - before | 26,789 | 27,860 | 32,979 | 25,252 | 27 |
| simulator.left2.fq - after  | 26,906 | 28,082 | 33,067 | 25,092 |   |
| simulator.left5.fq - before | 27,136 | 7,962  | 32,844 | 45,087 | 43 |
| simulator.left5.fq - after  | 27,074 | 7,973  | 33,017 | 45,099 |   |
| simulator.left6.fq - before | 26,789 | 7,871  | 32,961 | 45,229 | 57 |
| simulator.left6.fq - after  | 26,793 | 7,966  | 33,198 | 45,154 |   |
| simulator.left8.fq - before | 53,457 | 52,663 | 53,383 | 51,825 |   |
| simulator.left8.fq - after  | 53,457 | 52,663 | 53,383 | 51,825 |   |
| simulator.right1.fq - before| 27,033 | 27,990 | 33,194 | 25,042 | 20 |
| simulator.right1.fq - after | 27,165 | 28,049 | 33,077 | 25,100 |   |
| simulator.right2.fq - before| 27,025 | 28,231 | 33,051 | 24,816 | 28 |
| simulator.right2.fq - after | 26,913 | 27,967 | 33,150 | 25,041 |   |
| simulator.right5.fq - before| 47,066 | 28,003 | 13,173 | 25,017 | 20 |
| simulator.right5.fq - after | 47,166 | 27,845 | 13,173 | 25,086 |   |
| simulator.right6.fq - before| 47,087 | 28,244 | 12,997 | 24,795 | 28 |
| simulator.right6.fq - after | 46,762 | 28,127 | 13,076 | 25,186 |   |

<details><summary>Old</summary>

### libstdc++

| File Name | A | C | G | T | N |
|-----------|---|---|---|---|---|
| simulator.left3.fa - before | 25,217 | 24,694 | 24,794 | 25,266 | 29 |
| simulator.left3.fa - after  | 25,075 | 24,731 | 24,932 | 25,177 | 85 |
| simulator.left4.fa - before | 24,984 | 24,904 | 24,906 | 25,164 | 42 |
| simulator.left4.fa - after  | 25,039 | 24,802 | 24,852 | 25,226 | 81 |
| simulator.left7.fa - before | 25,177 | 24,815 | 24,832 | 25,151 | 25 |
| simulator.left7.fa - after  | 25,225 | 24,800 | 24,849 | 25,041 | 85 |
| simulator.right3.fa - before| 25,115 | 24,890 | 24,858 | 25,118 | 19 |
| simulator.right3.fa - after | 25,106 | 24,854 | 24,753 | 25,203 | 84 |
| simulator.left1.fq - before | 27,245 | 27,820 | 32,882 | 25,266 | 29 |
| simulator.left1.fq - after  | 27,067 | 27,800 | 33,127 | 25,177 | 85 |
| simulator.left2.fq - before | 27,276 | 27,869 | 32,621 | 25,157 | 28 |
| simulator.left2.fq - after  | 27,123 | 27,779 | 32,971 | 25,255 | 71 |
| simulator.left5.fq - before | 27,245 | 8,149  | 32,861 | 44,936 | 51 |
| simulator.left5.fq - after  | 27,067 | 8,084  | 33,127 | 44,893 | 85 |
| simulator.left6.fq - before | 27,276 | 7,978  | 32,599 | 45,050 | 48 |
| simulator.left6.fq - after  | 27,123 | 7,815  | 32,971 | 45,219 | 71 |
| simulator.left8.fq - before | 53,345 | 52,501 | 53,113 | 51,394 |   |
| simulator.left8.fq - after  | 53,345 | 52,501 | 53,113 | 51,394 |   |
| simulator.right1.fq - before| 27,131 | 28,025 | 32,918 | 25,118 | 19 |
| simulator.right1.fq - after | 27,176 | 27,993 | 32,730 | 25,203 | 84 |
| simulator.right2.fq - before| 26,985 | 27,978 | 32,925 | 25,214 | 25 |
| simulator.right2.fq - after | 27,026 | 27,736 | 32,991 | 25,305 | 71 |
| simulator.right5.fq - before| 46,946 | 28,048 | 13,098 | 25,100 | 19 |
| simulator.right5.fq - after | 47,042 | 27,993 | 12,864 | 25,203 | 84 |
| simulator.right6.fq - before| 46,963 | 27,995 | 12,947 | 25,197 | 25 |
| simulator.right6.fq - after | 47,110 | 27,736 | 12,907 | 25,305 | 71 |

### libc++

| File Name | A | C | G | T | N |
|-----------|---|---|---|---|---|
| simulator.left3.fa - before | 25,116 | 24,646 | 24,932 | 25,283 | 23 |
| simulator.left3.fa - after  | 25,177 | 24,773 | 24,757 | 25,204 | 89 |
| simulator.left4.fa - before | 24,960 | 24,924 | 24,911 | 25,174 | 31 |
| simulator.left4.fa - after  | 25,047 | 24,907 | 24,839 | 25,121 | 86 |
| simulator.left7.fa - before | 25,134 | 24,801 | 25,253 | 24,778 | 34 |
| simulator.left7.fa - after  | 24,861 | 24,997 | 25,008 | 25,060 | 74 |
| simulator.right3.fa - before| 25,009 | 24,899 | 25,030 | 25,042 | 20 |
| simulator.right3.fa - after | 25,089 | 24,652 | 25,136 | 25,047 | 76 |
| simulator.left1.fq - before | 27,136 | 27,762 | 32,868 | 25,283 | 23 |
| simulator.left1.fq - after  | 27,139 | 27,739 | 32,889 | 25,204 | 89 |
| simulator.left2.fq - before | 26,789 | 27,860 | 32,979 | 25,252 | 27 |
| simulator.left2.fq - after  | 26,739 | 27,918 | 33,153 | 25,245 | 66 |
| simulator.left5.fq - before | 27,136 | 7,962  | 32,844 | 45,087 | 43 |
| simulator.left5.fq - after  | 27,139 | 7,854  | 32,889 | 45,089 | 89 |
| simulator.left6.fq - before | 26,789 | 7,871  | 32,961 | 45,229 | 57 |
| simulator.left6.fq - after  | 26,739 | 7,961  | 33,153 | 45,202 | 66 |
| simulator.left8.fq - before | 53,457 | 52,663 | 53,383 | 51,825 |   |
| simulator.left8.fq - after  | 53,457 | 52,663 | 53,383 | 51,825 |   |
| simulator.right1.fq - before| 27,033 | 27,990 | 33,194 | 25,042 | 20 |
| simulator.right1.fq - after | 27,123 | 27,754 | 33,080 | 25,047 | 76 |
| simulator.right2.fq - before| 27,025 | 28,231 | 33,051 | 24,816 | 28 |
| simulator.right2.fq - after | 27,117 | 28,087 | 32,997 | 24,773 | 85 |
| simulator.right5.fq - before| 47,066 | 28,003 | 13,173 | 25,017 | 20 |
| simulator.right5.fq - after | 47,359 | 27,754 | 12,844 | 25,047 | 76 |
| simulator.right6.fq - before| 47,087 | 28,244 | 12,997 | 24,795 | 28 |
| simulator.right6.fq - after | 47,253 | 28,087 | 12,861 | 24,773 | 85 |

</details>


apps/mason2/simulate_illumina.cpp
```diff
@@ -168,6 +168,9 @@ void _simulateSequence(TRead & read, TRng & rng, TFrag const & frag,
     typedef typename seqan2::Iterator<TFrag>::Type TFragIter;
     TFragIter it = begin(frag, seqan2::Standard());
 
+    std::uniform_int_distribution<unsigned short> dist{};
+    using param_t = typename std::uniform_int_distribution<unsigned short>::param_type;
```
rrahn, line 172:
```suggestion
    using param_t = typename decltype(dist)::param_type;
```

apps/mason2/simulate_illumina.cpp
```diff
@@ -183,22 +186,13 @@ void _simulateSequence(TRead & read, TRng & rng, TFrag const & frag,
             continue;
         }
 
+        // NOTE: We can only insert CGAT, but we can have a polymorphism to N.
+        dist.param(param_t{0u, static_cast<unsigned short>(cigar[i].operation == 'I' ? 3u : 4u)});
```
rrahn, line 190:
```suggestion
        dna_dist.param(dna_dist_param_t{0u, static_cast<dna_dist_int_t>(cigar[i].operation == 'I' ? 3u : 4u)});
```

apps/mason2/simulate_illumina.cpp
```diff
@@ -183,22 +186,13 @@ void _simulateSequence(TRead & read, TRng & rng, TFrag const & frag,
...
-                x = dist(rng);
-            int num = static_cast<int>(x / 0.25);
-
-            // NOTE: We can only insert CGAT, but we can have a polymorphism to N.
-
-            if (cigar[i].operation == 'I')
-                appendValue(read, seqan2::Dna5(num));
-            else
-                appendValue(read, seqan2::Dna5(num + (num == ordValue(*it))));
+            appendValue(read, seqan2::Dna5(dist(rng)));
```
rrahn, line 195: This does not guarantee that in case of a substitution the simulated symbol is different to the one in the fragment sequence, does it? So in 20% of the cases it is the same symbol. We would need to draw randomly until we get a distinct symbol.
That could potentially lead to long execution times. So instead we could always draw in the range from 0..3, and add a case distinction for the substitutions, such that all drawn ranks greater or equal than the `ordValue(*it)` are incremented by one.

apps/mason2/simulate_sanger.cpp
```diff
@@ -58,6 +58,9 @@ void _simulateSequence(TRead & read, TRng & rng, TFrag const & frag,
     typedef typename seqan2::Iterator<TFrag>::Type TFragIter;
     TFragIter it = begin(frag, seqan2::Standard());
 
+    std::uniform_int_distribution<unsigned short> dist{};
+    using param_t = typename std::uniform_int_distribution<unsigned short>::param_type;
+
```
rrahn, line 63: Since this is all copy'n'paste code (😢), the same comments from above apply to this part.

rrahn (changes requested):

apps/mason2/simulate_sanger.cpp
eseiler: I'll check whether I can put that function in some header

apps/mason2/simulate_illumina.cpp
SGSSGene: I think, this note is "wrong". This note shouldn't exists and an 'N' should never be drawn. It feels more like a documentation of a bug.
I really think that for substitution it was supposed to be something like: `x = (x + rand(1, 3)) % 4` (pseudo code).

In general this creates a huge doubt in whether mason actually represents any actual illumina sequence reader output. I can't imagine that this reflects how any of this works in the real world.

eseiler: Those are the four CIGARS simulated:
```
M = match     (consumes reference and read)
I = insertion (consumes read)
D = deletion  (consumes reference)
X = mismatch  (consumes reference and read)
```

It's a bit weird because `M` means alignment match, i.e. the sequence can be either a match or mismatch.
`=` would be a sequence match, which is how `M` is handled in the simulation.
`X` is a sequence mismatch.

M and D are already handled.

Seems like the intent is that an insertion cannot be `N` because it would end up just being an alignment match `M`.
But for a sequence mismatch, it could also be `N`.

rrahn: In general, Illumina reads can have Ns if the base call could not resolve the exact nucleotides. The question is, however, whether that follows the distribution given here, which is highly doubtable. So yes, Ns can be drawn, but what is the real distribution of this? This may also be more likely at the ends of a read or only for small fraction of reads that are close to the edge of the flow cell, etc.. 
So I have two takes on this. Either Mason simulates reads as given by sequencer, or simulates reads after fastqc quality control, where reads containing Ns are discarded (as @SGSSGene suggests). I am in favour of the last one as well as most scenarios would rather build on the assumption that reads have been run through quality control. It gives us the option to later add a dedicated pre-quality control option to Mason where we then may use a more realistic distribution of such failed base calls.
But removing the Ns from simulation now might require a bump of the minor version.

eseiler: I didn't find any other read simulator that simulates `N`.
mason1 also simulates `N`, but it has an option for turning it off (default is on).

Illumina can produce `N`, but this is more likely at the beginning or end of sequencing.

> But removing the Ns from simulation now might require a bump of the minor version.

I removed `N`s for now, but this release would be a version bump anyway?

apps/mason2/simulate_illumina.cpp
eseiler: Yes, substitutions were wrong.

> We would need to draw randomly until we get a distinct symbol.
> That could potentially lead to long execution times.

This is usually fine. It's only 1 in 4 that we need to draw two times, 1 in 16 that we need to draw three times, etc.
And drawing is quite fast.
mason genome does the same, and we do the same in seqan3.

> So instead we could always draw in the range from 0..3, and add a case distinction for the substitutions, such that all drawn ranks greater or equal than the ordValue(*it) are incremented by one.

But then you would have two problems:
* You would need a modulo to handle T->A
* The distribution would be skewed, instead of 1/3 for each substitution, the next rank would have 1/2, and the two others 1/4. Similar to what we have now.

rrahn: No it should not be skewed, since you randomly picked from the set DNA\{base of fragment}. There is no modulo required. This is what I thought of:
```cpp
rank += (rank >= ordValue(*it));
```
So whenever the random rank is less than `ordValue(*it)` we can use it. Otherwise if the random rank is greater than or equal to `ordValue(*it)` we add one to get the correct base whose rank is greater than the one of `ordValue(*it)`.
Of course if you only consider dna4 then the distribution would be parameterised over the interval `[0..2]`.

eseiler: Yes, that would work. I assumed `[0,3]`.

But then I would either need two distributions or change the parameters. Is that preferable to just redrawing?

apps/mason2/simulate_illumina.cpp
SGSSGene: > I think it is fine to do it within version bump for the regular 2.5. release. Just out of curiosity. You updating it within the SeqAn2 apps, but I think @SGSSGene started to pull Mason out into a separate application under the SeqAn organization: https://github.com/seqan/mason. Which one is now the most current one?

Such a good question. I have almost zero memory about that mason fork/separation. We only work on it on 16th of May and 17the of May 2023 and then never changed anything again? super weird xD

apps/mason2/simulate_illumina.cpp
rrahn: Not if you can guarantee that it never happens that for some odd reasons redrawing stalls for a long time. Which is arguably very very very unlikely and yet can cause a lot of headache if it happens. 
So I would slightly prefer single draw over redraw.

apps/mason2/simulate_illumina.cpp
eseiler: Simon forked, I noticed and was curious. Then we did something for a day and that was it.

The idea was to put mason into a separate repository because it gets frequent questions and bug reports.
So it would be easier to release new versions if it was separate from the seqan library.

Either this is something we should tackle after the 2.5.0 release, i.e., remove mason from seqan and point to the new repo; or we delete the fork to avoid confusion.

apps/mason2/simulate_illumina.cpp
eseiler: > Not if you can guarantee that it never happens that for some odd reasons redrawing stalls for a long time. Which is arguably very very very unlikely and yet can cause a lot of headache if it happens.

Well, it would require that something in the pseudorandom generation breaks. I suspect if that ever happens, mason stalling is the least of the problems that ensue (cryptography also relies on pseudorandom number generation, no?).

Anyway, so I just do two random distributions `dna_dist_insertion`, `dna_dist_substitution`?
Just asking because I don't want to regenerate the test files each time :)

rrahn (approved):
Looks great! Thank you!

PR #2549, opened 2025-02-26
Converts assertion to exception in case two traces cannot be merged.

rrahn:
Not fixing #2540 but working around assert that could in release mode lead to UB. This way there is a decent abortion when two traces cannot be merged.
Looking deeper into the code suggests that a larger rework is necessary and better be dealt with when implementing such algorithm inside of modern SeqAn versions.


eseiler (approved):
Thank you!

PR #2553, opened 2025-03-13
[FIX] mason2: Detect when out of disk space.

### LINKED ISSUE #2552: Segmentation Fault (SIGSEGV) in mason_simulator with Large -n or -oa Option
# Segmentation Fault (SIGSEGV) in `mason_simulator` with Large `-n` or `-oa` Option  

## Description  
Hi,  

I cloned the **Sequan** repository and compiled it from the **main** branch. While using `mason_simulator` to generate reads from the **T2T reference genome** ([hs1.fa.gz](https://hgdownload.soe.ucsc.edu/goldenPath/hs1/bigZips/hs1.fa.gz)), I encountered a **segmentation fault (SIGSEGV)** when running the following command:  

```bash
time mason_simulator --num-threads 16 -ir ../../references/hs1/hs1.fa \
  --fragment-mean-size 300 --fragment-size-std-dev 30 \
  --illumina-read-length 150 -n 850000000 \
  -o reads_l_2.fq -or reads_r_2.fq -oa alignment_2.sam
```

## Error Message

```bash
...
  chr10 (allele 2) ................................................................................................................................................................................................................................................................................................ (18372259 fragments) OK
  chr12 (allele 1) .................................................................................................................................................................................................................................................
________________________________________________________
Executed in   91.06 mins    fish           external
   usr time  741.42 mins    1.55 millis  741.42 mins
   sys time   29.78 mins    2.07 millis   29.78 mins

fish: Job 1, 'time mason_simulator --num-thre…' terminated by signal SIGSEGV (Address boundary error)
```

## Observations

The **segmentation fault disappears** if I either:

- Use a smaller value for -n.
- Remove the -oa option (output alignment file).

## Question
Is this an issue with my setup, or could it be a bug in the program?
Any guidance would be appreciated.

Thanks in advance! 😊

eseiler:
Resolves #2552

- [ ] Do this for all `underflow` and `overflow` calls in functions that return `void` and do not check the return value? 
- [ ] The compression stream changes should probably be ported to seqan3.
- [x] Check if putting the throw in a noreturn function produces better assembly

### DIFF
--- a/include/seqan/stream/iostream_bzip2_impl.h
+++ b/include/seqan/stream/iostream_bzip2_impl.h
@@ -118,15 +118,15 @@ namespace bzip2_stream{
                     )
     {
         int w = static_cast<int>(this->pptr() - this->pbase());
-        if (c != EOF) {
+        if (!Tr::eq_int_type(c, Tr::eof())) {
              *this->pptr() = c;
              ++w;
          }
          if ( bzip2_to_stream( this->pbase(), w)) {
              this->setp( this->pbase(), this->epptr());
-             return c;
+             return Tr::not_eof(c);
          } else
-             return EOF;
+             return Tr::eof();
     }
 
     template<
--- a/apps/mason2/mason_simulator.cpp
+++ b/apps/mason2/mason_simulator.cpp
@@ -991,14 +991,23 @@ public:
                 }
 
                 // Write out the temporary sequence.
-                for (int tID = 0; tID < options.numThreads; ++tID)
+                try
                 {
-                    unsigned idx = rID * haplotypeCount + hID;
-                    writeRecords(*seqFileOuts[idx], threads[tID].ids, threads[tID].seqs, threads[tID].quals);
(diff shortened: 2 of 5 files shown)


include/seqan/stream/iostream_bzip2_impl.h
```diff
@@ -118,15 +118,15 @@ namespace bzip2_stream{
...
         int w = static_cast<int>(this->pptr() - this->pbase());
-        if (c != EOF) {
+        if (!Tr::eq_int_type(c, Tr::eof())) {
              *this->pptr() = c;
              ++w;
          }
          if ( bzip2_to_stream( this->pbase(), w)) {
              this->setp( this->pbase(), this->epptr());
-             return c;
+             return Tr::not_eof(c);
```
eseiler, line 127: This is the important change.
Using the traits's functions (`eq_int_type`, `not_eof`, `eof`) is style / the technically-more-correct version.

rrahn (approved):
looks great! Thank you!

PR #2563, opened 2025-07-19
[FIX] ICE on MSVC 19.44

eseiler:
Note that, at time of writing, godbolt's MSVC is 19.43. Current and the runner's version is 19.44.

Simplest test is `test_basic_aggregate`.

Using `unsigned` instead of `int` in `template <typename T1, typename T2, template <unsigned, unsigned> typename TSpec, int BITSIZE1, int BITSIZE2>` causes another ICE for razers, param_chooser, splazers, and others.


SGSSGene:
I can confirm bug and fix!

SGSSGene (approved):

